PR #498, opened 2014-12-29
ensure that owners are synchronised between journals and suggestions

richard-jones:
This is implemented at the save() level, so any operation which causes the owner to change will cause the related journal/application to be updated.  It comes that the cost of an additional query to the index, and the possibility of an additional write (if the owner needs to be synchronised).

### DIFF
--- a/portality/models/suggestion.py
+++ b/portality/models/suggestion.py
@@ -142,8 +142,19 @@ class Suggestion(Journal):
             return
         self._set_suggestion_property("suggester", sugg)
 
-    def save(self, **kwargs):
+    def _sync_owner_to_journal(self):
+        if self.current_journal is None:
+            return
+        from portality.models import Journal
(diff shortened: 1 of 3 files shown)


portality/models/suggestion.py
```diff
@@ -142,8 +142,19 @@ def set_suggester(self, name, email):
...
+            return
+        from portality.models import Journal
+        cj = Journal.pull(self.current_journal)
+        if cj is not None and cj.owner != self.owner:
+            cj.set_owner(self.owner)
+            cj.save(sync_owner=False)
+
+    def save(self, sync_owner=True, **kwargs):
         self.prep()
+        if sync_owner:
```
emanuil-tolev: I didn't think of the endless loop that would otherwise occur when I first read the issue / PR :). Nice.

PR #499, opened 2015-01-02
modify reapp init script to allow generate for a. all doaj, b. just one ...

richard-jones:
...account, c. just one journal, with all the associated preparation and csv generation as needed

### DIFF
--- a/portality/migrate/reapplication/reapp_init.py
+++ b/portality/migrate/reapplication/reapp_init.py
@@ -2,27 +2,87 @@ from portality import models
 from datetime import datetime as dt
 from portality import reapplication
 from portality.core import app
-import os
+import os, time
 import codecs
 import csv
 from portality.clcsv import UnicodeWriter
 
+class ReappCsvException(Exception):
+    def __init__(self, message, account_id):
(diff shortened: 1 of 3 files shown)


portality/migrate/reapplication/reapp_init.py
```diff
@@ -2,97 +2,118 @@
...
     # remove all of the current reapplication records
     models.BulkReApplication.delete_all()
+    # models.BulkReApplication.refresh()
 
-def create_reapplication(journal):
-    if not isinstance(journal, models.Journal):
-        journal = models.Journal.pull(journal)
-    if not isinstance(journal, models.Journal):
-        print "Unable to resolve Journal from identifier"
+    # FIXME: should probably remove all the old csvs too
```
emanuil-tolev: It opens csv-s in overwrite mode. But maybe to avoid confusion we should remove them. However we can't remove all of them if only one journal or account is being deleted - otherwise it'd defeat the purpose of having these options on this script as helpers during firefighting after reapplications go live. So really the way it's now is probably best - only touch (and overwrite) files when they're being generated.

PR #500, opened 2015-01-05
Fix column label generation in reapp spreadsheet error reports

emanuil-tolev:
Turns out int2base(i, 26) doesn't work that well :). I tried diagnosing but gave up after lots of print statements failed to prod my brain into inventing a solution.

The problem is

```python
        while x:
            digits.append(digs[x % base])
            x /= base
```

Given ```40``` (or the 41st column in our 0-indexed setting a.k.a. the ```AO``` column) this will correctly identify the 2nd letter ```O```, because 40 % 26 = 14. The 15th letter in a 0-indexed list of the uppercase alphabet.
But 40 / 26 = 1. So it passes the while loop condition, as it should (we do need 2 letters). However now you're doing ```1 % 26``` in the 2nd pass, which results in ```1```, or the 2nd letter of the alphabet - ```B```. So you end up with ```BO```, not ```AO```.

Doing something like ```x = x / base - 1``` (instead of ```x /= base```) does not work, because then the while loop will stop. We have a (simple) contradiction - x needs to be 0 in order to ever select ```A``` as a letter, but if it ever goes to 0 the loop stops.

The whole thing looks like a recursive function with a base case of 0 == A for a reason :), the only thing missing is it wasn't recursive. I didn't feel like converting it so I translated a tried & tested code from stackoverflow into Python and added tests.

### DIFF
--- a/doajtest/unit/test_reapplication.py
+++ b/doajtest/unit/test_reapplication.py
@@ -197,7 +197,13 @@ APPLICATION_COL = [
 
 @classmethod
 def mock_issns_by_owner(cls, *args, **kwargs):
-    return ["1234-5678", "2345-6789", "3456-7890"]
+    issns = ["1234-5678", "2345-6789", "3456-7890", "4567-8901"]
+    for c_num in range(5, 80):
+        c_num = str(c_num)
+        issn = "6529-540" if len(c_num) == 1 else "6529-54"
+        issn += c_num
+        issns.append(issn)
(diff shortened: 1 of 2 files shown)


richard-jones:
huh, interesting.  As it happens, I thought this was also a tried and tested function, which I snagged from SO :) A quick prod suggested it worked fine, but goes to show you are always better off with real tests - good catch, here.

PR #502, opened 2015-01-06
Feature/448 second attempt

richard-jones:
Makes the URL fields required, and separates them from a dependency on the APC/Submission Yes/No fields.

Note that there are data model changes here - I have not provided a migrate script, as this data should not be in the production system.


emanuil-tolev:
Are test and staging going to be affected in some way by the data model changes? It might be simplest to refresh them from live first if so. But there's no point in spending time doing that if they'll still work if this gets merged.

richard-jones:
They should still work - my local test instance did.  Depending on whether
you want to refresh the data anyway, I would just roll it out and then
refresh the data again only if there are problems.

On 6 January 2015 at 18:07, Emanuil Tolev <notifications@github.com> wrote:

> Are test and staging going to be affected in some way by the data model
> changes? It might be simplest to refresh them from live first if so. But
> there's no point in spending time doing that if they'll still work if this
> gets merged.
>
> —
> <https://github.com/DOAJ/doaj/pull/502#issuecomment-68905413>.
>



-- 

Richard Jones,

Founder, Cottage Labs
w: http://cottagelabs.com

emanuil-tolev:
I read the changes a bit more deeply now. Yeah, it won't have a bad effect, just some test records will have superfluous data, so if somebody had entered APC info in there before, it will disappear now. But live has never had the new structure, so nothing is going to happen when it goes live, I see your point.

I wasn't planning on refreshing the data but we might do next week for some kind of final testing.

emanuil-tolev:
And I see the test server reapplications should be regenerated after this deploys :).

Issue #462, opened 2014-12-04
Can we add timezone details to the Upload Date?


dommitchell:
Currently this field says:

04 Dec 2014 15:23:38 

Can we add GMT or UTC? Or something to indicate it is not local time?

Steven-Eardley:
@emanuil-tolev @richard-jones I've spent more time (ha) on this and I've decided there's either a naïve and superficial solution, or a real fix that actually changes quite a few things.

The problem is simply that in the dao underlying everything the timezones aren't recorded. So timestamps in the index are relative to whichever machine they were created on. To add a timezone label  when we display the timestamp, we could do the following:

* naïvely, use something like ```tzlocal.get_localzone()``` to find the timezone on the server and append its name whenever we print out a timestamp. In our case, this will say 'GMT' for files uploaded in winter, 'BST' for files uploaded in summer, and will only affect the UI. The problem is, if the server data has crossed timezones these will be incorrect - we are only checking where the server is *now*.

As an example:
```python
>>> here = tzlocal.get_localzone()   # here is a tzinfo object
>>> here.tzname(datetime.now())    # as of 07 Jan 2015
'GMT'
>>> summer = datetime(2014,6,6)    # 06 Jun 2014 00:00:00
>>> here.tzname(summer)
'BST'
```

* properly, we only ever encode dates using ```datetime.utcnow()``` to prevent these daft problems. Wherever the server is, the timestamps are written unambiguously and if there's a requirement to show local times, the translation is trivial using available libraries. This method will require some changes to the dao and each time ```datetime.now()``` is used, and a migration if the timestamps already written to the index are to be trusted.

richard-jones:
Yes, you are quite right - the real underlying problem is that we record the times in the timezone of the server, rather than absolutely with timezone information.  I would like, overall, to fix that, by - as you suggest - always storing the utc time when we store a timestamp.

But, from a practical point of view for this project, I propose that we do your first option, since this should work provided that DOAJ runs on the same machine.  You can see how I did this for page lock timestamps here:

https://github.com/DOAJ/doaj/blob/develop/portality/models/lock.py#L52

So, propose to do the fix the simple way, but could you open an issue describing what we should do in the long run, and we'll look at how much is required to implement that later.

Steven-Eardley:
Splendid, thanks.

emanuil-tolev:
[...]

PR #508, opened 2015-01-08
Added timezone label to timestamps in file upload tables. Achieved by ad...

Steven-Eardley:
...ding a context processor to apply formatting from Flask.

new requirement: pytz, for timezone information. #462

### DIFF
--- a/portality/app.py
+++ b/portality/app.py
@@ -10,6 +10,10 @@ import os
 from flask import request, abort, render_template, redirect, send_file, url_for, flash
 from flask.ext.login import login_user, current_user
 
+from datetime import datetime
+import tzlocal
+import pytz
+
 import portality.models as models
 from portality.core import app, login_manager
 from portality import settings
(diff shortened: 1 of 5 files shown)


Steven-Eardley:
As I said in my FIXME, is there a better place to put the new context_processor?

emanuil-tolev:
You could use a jinja filter function. See line 90 in portality.core . Feel free to define it right there - if we get enough of them they'll just go in a separate module. The way you use a filter is

{{ template_variable | filter_name }}

So for the debug one that is defined you might do

```
{# Journal ID seems to be blank but it was set in the python route that renders this template! Oh noes what is jinja doing with my value #}
{{ journal.id | debug }}
{# the above will output *nothing* to the template because the debug filter returns "", but will print the var to the log#}
```

Does that sound adequate for this use case? It's not gonna get applied to all dates automatically, obviously, but then neither is a context processor function.

Steven-Eardley:
Well, I'm happy to do it either way. The context_processor is just a way of making my function available to the template. You're right though, the behaviour is quite filter-y so I'll rewrite it as one.

Steven-Eardley:
I've made it a filter, but left it where it was. It's next to the other ones written for ```publisher.py```, and it's only used there for now. I also prefer the way of registering them via the blueprint. I'm not sure why.

emanuil-tolev:
Looks good! I'd probably put it in a more general place though, like core or app or similar. Formatting dates is not something I'd like to see 2 blueprints, or any 2 parts of the app do differently from each other :). Making it more general increases the chance it will be reused in the future instead of yet another one written?

Steven-Eardley:
Putting it in app.py meant I could still register it nicely, so smiles all round.

emanuil-tolev:
Cool, let's try it on test!

(BTW the above was just an opinion, if you really feel like it's better off in the publisher view, feel free to put it there.)

Steven-Eardley:
I think we found the best solution! :rainbow: collaboration :rainbow:

emanuil-tolev:
oh god the c-word :D

PR #510, opened 2015-01-10
fix date granularity issue

richard-jones:
Introduces a DateFormat object that handles the granularity and date formatting issues

### DIFF
--- a/portality/view/oaipmh.py
+++ b/portality/view/oaipmh.py
@@ -71,8 +71,38 @@ def oaipmh(specified=None):
     return resp
 
 #####################################################################
-## Utility methods
+## Utility methods/objects
 #####################################################################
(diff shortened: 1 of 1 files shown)


portality/view/oaipmh.py
```diff
@@ -71,9 +71,39 @@ def oaipmh(specified=None):
...
+class DateFormat(object):
+    @classmethod
+    def granularity(self):
+        return "YYYY-MM-DDThh:mm:ssZ"
+
+    @classmethod
+    def default_earliest(cls):
+        return "1970-01-01T00:00:00Z"
+
+    @classmethod
```
emanuil-tolev: Could be @staticmethod . Not blocking though.

Issue #323, opened 2014-09-23
Legacy Data Archive Scripts: Journals


richard-jones:
1. Automatically withdraw all items with incomplete re-applications.  This will be run only at some date to be specified by the DOAJ team.  All journals in the DOAJ which have an active re-application (which will be linked to by the journal record admin.current_application) and which do not have a re-application date in their metadata (last_reapplication) will be marked as in_doaj=False, along with all of their associated articles.  See #307 for data model details.
2. Archive all withdrawn journals and their associated articles (and those article’s histories).  For every journal with in_doaj=False, they should be copied, along with their article metadata, into a static (offline) archive and deleted without record from the DOAJ index.  The output of this script should be zipped up and delivered to the DOAJ to retain for as long as is necessary.

Steven-Eardley:
@richard-jones Regarding part 2: 
* Are we deleting all journals with ```in_doaj = false```, or just the ones flipped in the first part?
* Similarly, to catch their articles, is it all with ```in_doaj = false``` or those found explicitly with ```journal.all_articles()```?

richard-jones:
@Steven-Eardley.  After the first stage, the second stage is like this:

* for each journal in_doaj=False get all their articles and remove them
* for each journal in_doaj=False, remove the article

Is that what you were expecting?

Steven-Eardley:
> for each journal in_doaj=False, remove the article

Yes, what I've expected (assuming you meant journal for that second 'article') - my quandary was that it could scoop up more than reapplication-specific stuff. By design, I hope.

richard-jones:
yes, sorry - I'll correct the text.

It will - by design - remove any old journals and any old articles from the system that are no longer published, irrespective of whether they went through the reapplication process or not.

richard-jones:
RJ to review during next testing cycle

dommitchell:
@richard-jones Can this one be closed?

emanuil-tolev:
This relates to the reapplication shutdown @dommitchell , which hasn't come about yet :)

richard-jones:
reapplication has been shut down so this can be closed

PR #511, opened 2015-01-12
Feature/323 journal archives

Steven-Eardley:
Use of ```Journal.all_in_doaj()``` might mean the withdraw script needs to be run more than once - the default page size is not enough to catch them all. (So I've doubled it in this case).

#323

### DIFF
--- a/portality/migrate/reapplication/delete_unreapplied.py
+++ b/portality/migrate/reapplication/delete_unreapplied.py
@@ -9,27 +9,21 @@ conn = raw.make_connection(None, 'localhost', 9200, 'doaj')
 archv_query = {
             "query" : {
                 "term" : {
-                    "admin.application_status.exact" : "rejected"
+                    "admin.in_doaj" : "false"
                     }
                 }
             }
 
-out_file = open('journals_not_reapplied.json', 'w')
+json_writer = tasks.JSONListWriter('journals_not_reapplied.json')
 
-# Dump all rejected suggestions to file
-tasks.dump(conn, 'suggestion', q=archv_query.copy(), out=out_file)
-out_file.close()
+# Dump all journals not in DOAJ to file
+tasks.dump(conn, 'journal', q=archv_query.copy(), out=json_writer)
+json_writer.close()
 
-# Scroll through all rejected selections, delete all
+# Ask how many journals will be deleted.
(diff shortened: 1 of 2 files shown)


portality/migrate/reapplication/delete_unreapplied.py
```diff
@@ -0,0 +1,41 @@
+from esprit import tasks, raw
+from portality import models
+
+'''Delete journals and corresponding articles with incomplete reapplications'''
+
+# Connection to the ES index
+conn = raw.make_connection(None, 'localhost', 9200, 'doaj')
```
emanuil-tolev: You can guess the question here :). Why raw instead of through the DOAJ dao?

emanuil-tolev:
> Use of Journal.all_in_doaj() might mean the withdraw script needs to be run more than once - the default page size is not enough to catch them all. (So I've doubled it in this case).

There are more than 10k journals in live.

Isn't the idea with paging that it will go through all of them, page by page? (So as to e.g. not load all of them in memory at once, or some other reason - but it should nonetheless iterate over all of them, even if you gave it a page size of 10.) Maybe I'm misunderstanding it here..

Steven-Eardley:
I'm sure @richard-jones and I have run into this before - the behaviour seems to be that the page size does affect the number of results. I certainly had to run it twice when I tested it with a page size of 5000; first time getting just over 5000 results, the second getting about 2000 (with a total of 7k).

I've seen this in a few more places though, so if this outcome is a bug, it definitely needs looking into.

emanuil-tolev:
Hm, right. This is just a script to be run once so a few more MB memory won't hurt, it's already got to load ~300 MB if I remember correctly how big the journal type is. Could you change the page_size to 20'000 then? Otherwise it will miss a few journals when we run it in live.

portality/migrate/reapplication/delete_unreapplied.py
Steven-Eardley: The archiving is handled through esprit, which I usually set up thusly - is there a better way to dump them all to file by query?

emanuil-tolev: Ah you're using tasks.dump . No idea if there's a better way, but it seems reasonable, and presumably you've ran it successfully + we'll run it on staging eventually.

emanuil-tolev:
:+1:

PR #513, opened 2015-01-12
when a suggestion overwrites an existing journal, preserve the create date

richard-jones:
works by loading the journal being overwritten and nabbing its create date, before overwriting it.

Also with test.


emanuil-tolev:
> Also with test.

Oh my!

PR #527, opened 2015-01-19
Blank questions in the reapp csv where appropriate and fix APC always ye...

### LINKED ISSUE #525: blank out certain questions in the csv
If possible, don't put in default values of No policy in place and No for q13, q16, q23, q26, q35, q49. This will force users to fill them in as they're required.

RJ: unfortunately, for each of these questions, omission of a field means "No", and presence of a field means "Yes". So it is impossible to tell the difference between a deliberately missing field, and a field that has not yet been completed. That means we either stick with the current situation, or we wilfully blank out these questions in the CSV, whether they were previously populated or not (and maybe they wouldn't be?). It will be much much harder to do the same thing for the UI (thought not impossible).

Decision: we will blank out the values of these questions in the CSV

emanuil-tolev:
...s 

Fixes #525 .

APC and submission charges question (and a few others) were bugged in addition to the request to blank certain values if they were negative answers (see #525 for list of q-s). APCs and submission charges and a couple others were always set to Yes. Now I've tested it, it will be set to Yes / No dependent on data ( :) ).

### DIFF
--- a/portality/reapplication.py
+++ b/portality/reapplication.py
@@ -497,10 +497,10 @@ class Suggestion2QuestionXwalk(object):
             return ", ".join(aids)
 
         def yes_no(val):
-            return "Yes" if val in [True, "True", "Yes"] else "No"
+            return "Yes" if val in [True, "True", "Yes", "true", "yes"] else "No"
 
         def yes_or_blank(val):
-            return "Yes" if val in [True, "True", "Yes"] else ''
+            return "Yes" if val in [True, "True", "Yes", "true", "yes"] else ''
 
         def license_checkbox(val):
             opts = {}
--- a/portality/reapplication.py
+++ b/portality/reapplication.py
@@ -482,6 +482,12 @@ class Suggestion2QuestionXwalk(object):
             aids = forminfo.get(main_field, [])
             if aids is None or aids == "" or aids == "None":
                 aids = []
+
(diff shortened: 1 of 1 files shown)


portality/reapplication.py
```diff
@@ -482,6 +482,12 @@ def other_list(main_field, other_field, other_value):
             aids = forminfo.get(main_field, [])
             if aids is None or aids == "" or aids == "None":
                 aids = []
+
+            # if the xwalk has returned a single-list element like ["None"]
+            # we want to strip that "None" for the purpose of the CSV
+            if choices.Choices.NONE in aids:
+                aids.remove(choices.Choices.NONE)
```
emanuil-tolev: It is acceptable to do this on a generic basis here in the function instead of per question, because it's only article_identifiers, deposit_policies (both asked to be blanked in case of negative answer in #525) and fulltext_formats that use this function. fulltext_formats will not be affected by this change - it's a list question like the others and it's acceptable to blank it in the absence of data.

Steven-Eardley:
> ```python
return "Yes" if val in [True, "True", "Yes"] else "No"```

Will this take into account lower case letters (e.g. 'true' or ''yes')?

emanuil-tolev:
No, shouldn't be any, but I will add handling! Thanks

emanuil-tolev:
OK, added lowercase versions too.

Issue #548, opened 2015-01-27
Assigning journal to Editor Group invokes Bad Gateway


dommitchell:
https://doaj.org/admin/suggestion/0281894b53cc4196b4bf6404cbe0c4d2#first_problem at 15.54pm (14.54 UTC)

![gate](https://cloud.githubusercontent.com/assets/5955396/5920452/ef42d674-a63c-11e4-806f-344060f59901.jpg)

It's happened to me 3 times now. 

I'm opening applications that are already in progress and assigning them to Editor group 'Dom's group'. If I hit back on my browser, unlock and close the journal is indeed assigned to the correct group

emanuil-tolev:
Does it hang there for 30s (or so) before showing you this 502 Bad Gateway screen?

emanuil-tolev:
Yeah, something weird is going on. https://rpm.newrelic.com/accounts/526071/applications/3171791/transactions#id=540963092 the application admin page has suddenly started taking 30s to load (and timing out).

emanuil-tolev:
Nothing helpful in the server logs, just an indication from the Newrelic daemon that the DOAJ app has timed out. Strange way for it to time out though, something seems to have called sys.exit - maybe not our own code, but the point is it seems the interpreter is shutting down. I'm sure it's only a symptom of what the problem actually is :). Will try locally now.

emanuil-tolev:
OK, so naturally I can't reproduce it with a random application from Staging (using same code as Live), otherwise life would be too easy.

emanuil-tolev:
Hahah, it only seems to happen to "Dom's group" @dommitchell :). I tried assigning that application you have in the original post to English1 and it worked. Switch back to Dom's group, booom.

emanuil-tolev:
Hm, I made "Dom's group" on staging (using the current live code) and assigned an application to it. But it worked :(. So still can't reproduce it..

emanuil-tolev:
I am even more confident it's "Dom's group" that causes this, and that it does not happen at random times. I just assigned that application to 5 different editor groups and Dom's group broke the page again. But only on live.

emanuil-tolev:
[...]

PR #549, opened 2015-01-27
Do not crash and burn when no good target is specified when sending an email

emanuil-tolev:


emanuil-tolev:
issue #548

richard-jones:
Looks ok to me, and certainly good to fix this issue initially.

That said, whatever is passing a [None] in should also be sorted out at some point :)

emanuil-tolev:
Not fault of the code @richard-jones , the account has no email set.

richard-jones:
Yup, sure - maybe best just to handle this in app_email then, as you've done.

Issue #539, opened 2015-01-24
New/updated facets required in Public Search interface


richard-jones:
- [x] change facet label 'Publication charges' to 'Article processing charges (APCs)'
- [x] change facet label 'Journal language' to 'Language of full text'

*Natural state*
- Journals or articles: Journals, Articles. Fixed ordering.
- Subject (?) - unsure of how this one behaves when we add the subject browser. Do we remove this altogether? Can we discuss?
- Article processing charges (APCs): Yes, No
- Journal license: CC BY, CC BY-NC, CC BY-NC-ND, CC BY-NC-SA, CC BY-SA, CC BY-ND, None, Not CC-like, Publisher's own
- Publisher: publisher names in full. Default sorting: descending numeric order
- Country of publisher: country names in full, not in code. Default sorting: descending numeric order
- Peer review: Editorial review, peer review, blind peer review, double blind peer review, open peer review
- Full text format: PDF, HTML, ePUB, XML. Default sorting: descending numeric order

*'Journals' selected*
- Subject (?)
- The Seal: Yes, No
- Article processing charges (APCs)
- Journal license
- Publisher
- Country of publisher
- Language of full text: languages in full, no codes.
- Peer review
- Article identifiers: None, DOI, Handle, ARK. Default sorting: descending numerical order.
- Waiver policy: Yes, No
- Archiving Policy: Yes, No
- Plagiarism policy: Yes, No
- Machine-readable license: Yes, No
- Deposit policy: Yes, No
- Year journal added to DOAJ: 2015, 2014, 2013, 2012... Default sorting: descending chronological order.

*'Articles' selected*
- Subject (?)
- Journal title
- Article processing charges (APCs)
- Journal license
- Publisher
- Language of full text
- Peer review
- Archiving policy
- Machine-readable license
- Year of publication (articles): 2015, 2014, 2013, 2012... Default sorting: descending chronological order.

- [x] The addition of these facets presents us with a usability problem. We would like to install a [More/Less] button that would Reveal/Hide further facets.

dommitchell:
Note to self: affiliation

richard-jones:
Some comments on the facets as I work through them the first time (will update as I go):

* APC: Yes/No - this is now a bit tricky, as there is no "there is an apc" field; we know there is an APC based on whether there is an apc_url, but this can't be trivially faceted on.  We would probably want to add to the "index" portion of the object an APC Yes/No field.

emanuil-tolev:
[...]

PR #596, opened 2015-03-10
Feature/facets

richard-jones:
This is a first pass on the new public search facets, implementing what I suggest in #539 .

Now that a general approach to enabling/disabling facets has been developed, depending on what other filters are selected, we can easily change things around.

In the mean time, we should merge this to develop so that it goes out to the test server, and can be reviewed.


emanuil-tolev:
> In the mean time, we should merge this to develop so that it goes out to the test server, and can be reviewed.

No test server exists just now and it's not been quite set up. @richard-jones do we need to consider running ES1.4 on yours while @markmacgillivray and I iron out the snags in the new infrastructure?

richard-jones:
I've just added some more code to this branch which handles link shortening.

bitly requires oauth credentials for shortening, so there is a back-end method which proxies for the bitly api, and which constrains the scope of the urls that can be shortened to the doaj search queries only.

In production, this is going to require an oauth token that belongs to the DOAJ's bitly account; for the time being I'm using my own one in testing, and will share that for the test server.

richard-jones:
I'm just adding new search interface enhancements to this branch, until it gets merged.  Latest changes normalise usage of Subject/Classification/Keywords across all search interfaces (with special treatment for the public interface).

richard-jones:
This next commit causes the page to scroll to the top when the next/previous buttons are used

richard-jones:
This one adds a new pager to all the facetviews

emanuil-tolev:
OK, let's see this go on http://doaj.cottagelabs.com !

Issue #559, opened 2015-01-28
Article results page


Steven-Eardley:
Article page for article query - related to PVOD, see #540

Issue #540, opened 2015-01-24
Single article landing page


richard-jones:
* includes embedded metadata
* links to OpenURL work

Steven-Eardley:
@dommitchell You can now review the article page changes via the search on doaj.cottagelabs.com

emanuil-tolev:
@Steven-Eardley Some articles seem to be causing an error, like https://doaj.cottagelabs.com/article/cc3a7008c7484cc68c319ca28358e937

This is the error
```
[2015-03-21 17:30:32 +0000] [16533] [ERROR] Error handling request
Traceback (most recent call last):
  File "/home/cloo/repl/apps/doaj/local/lib/python2.7/site-packages/gunicorn/workers/sync.py", line 130, in handle
    self.handle_request(listener, req, client, addr)
  File "/home/cloo/repl/apps/doaj/local/lib/python2.7/site-packages/gunicorn/workers/sync.py", line 176, in handle_request
    for item in respiter:
  File "/home/cloo/repl/apps/doaj/local/lib/python2.7/site-packages/newrelic-2.46.0.37/newrelic/api/web_transaction.py", line 711, in __iter__
    for item in self.generator:
  File "/home/cloo/repl/apps/doaj/local/lib/python2.7/site-packages/newrelic-2.46.0.37/newrelic/api/web_transaction.py", line 1087, in __call__
    self.start_response)
  File "/home/cloo/repl/apps/doaj/local/lib/python2.7/site-packages/flask/app.py", line 1701, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/cloo/repl/apps/doaj/local/lib/python2.7/site-packages/newrelic-2.46.0.37/newrelic/api/web_transaction.py", line 1215, in _nr_wsgi_application_wrapper_
    return wrapped(*args, **kwargs)
  File "/home/cloo/repl/apps/doaj/local/lib/python2.7/site-packages/flask/app.py", line 1689, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/home/cloo/repl/apps/doaj/local/lib/python2.7/site-packages/newrelic-2.46.0.37/newrelic/hooks/framework_flask.py", line 98, in _nr_wrapper_Flask_handle_exception_
    return wrapped(*args, **kwargs)
  File "/home/cloo/repl/apps/doaj/local/lib/python2.7/site-packages/flask/app.py", line 1687, in wsgi_app
    response = self.full_dispatch_request()
[...]

PR #601, opened 2015-03-12
First pass of article page, based on the current journal template. Issue...

Steven-Eardley:
...s #559, #540.
May require more thought into the design and required fields.

Once this is merged, @emanuil-tolev can add his metadata stuff.

Sorry for the whitespace at the end of the html file.

### DIFF
--- a/portality/templates/doaj/article.html
+++ b/portality/templates/doaj/article.html
@@ -1,0 +1,144 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+<div class="row-fluid">
+    <div class="span9">
+        <h1>{{article.bibjson().title}}</h1>
+        <p>
+        {% for identifier in article.bibjson().get_identifiers(idtype='pissn') %}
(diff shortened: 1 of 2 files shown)


emanuil-tolev:
@richard-jones @Steven-Eardley was wondering whether the new article page shouldn't follow the design of the new journal (TOC) page, which has not been written yet pending mine and @markmacgillivray's changes to the journal TOC back-end and URL routing :). Something to discuss next meeting p'raps.

emanuil-tolev:
(So for now I think @Steven-Eardley's gone for a basic design, correct me if wrong)

emanuil-tolev:
Looks good :+1: - not tested if it actually works locally since my repo is in a different state to do with the new infrastructure and I'm in a hurry to finish that, but it definitely does not break any current code and can thus be merged, with potential bugs to be fixed later when we can actually see the work .. working.

Steven-Eardley:
@emanuil-tolev correct, a basic design based on ransacking the toc template.

Issue #455, opened 2014-12-01
Specify OpenGraph metadata for Facebook


Steven-Eardley:
In issue #266 it was noted the wrong image appears when DOAJ is shared on Facebook:

![karger](https://cloud.githubusercontent.com/assets/5955396/5187163/69b4bc86-74cb-11e4-9ae1-099cdf88f190.jpg)

* This could be fixable by specifying og tags, per https://blog.kissmetrics.com/open-graph-meta-tags/
* Inferred open graph data can be searched via https://developers.facebook.com/tools/debug/og/object/
* facebook's wisdom on URL migrations https://developers.facebook.com/docs/sharing/opengraph/using-objects#update

Steven-Eardley:
@dommitchell after the pull request is merged, the share dialogue should look like that shown below (apart from the demo URL):

![screen shot 2015-03-12 at 20 39 32](https://cloud.githubusercontent.com/assets/1446355/6627881/393086b2-c8f9-11e4-9dbd-8b2c6c8f9297.png)

Let me know if you'd like something different!

dommitchell:
This is great! Very happy with this.

Steven-Eardley:
This can best be reviewed once it's live at the end of the PVOD project.

emanuil-tolev:
@dommitchell does it work from http://doaj.cottagelabs.com ?

dommitchell:
Yep!

![fb](https://cloud.githubusercontent.com/assets/5955396/6820586/e74a8eba-d2d0-11e4-8982-da5ff51d954a.jpg)

emanuil-tolev:
Alright, cool, it'll definitely work on Live then.

dommitchell:
Hi, this seems to have switched back showing the Karger logo.
![fb](https://cloud.githubusercontent.com/assets/5955396/7315918/6242b954-ea72-11e4-9c86-a14ec09684af.jpg)

Steven-Eardley:
@dommitchell I confirmed your issue using the object debugger, checked the tags are properly set (they are), and asked it to refresh the scraped data. The correct image and text appears there now for www.doaj.org, https://doaj.org, http://doaj.org. (it's a caching thing on their end)

@emanuil-tolev - it seems facebook isn't quite as smart as we hoped. Changing the canonical url to https has changed the entity resolved:

**results for now**
Fetched URL	http://www.doaj.org/
Canonical URL	https://doaj.org/ (23 likes, 24 shares)

**when last fetched**
[...]

PR #602, opened 2015-03-12
Open Graph tags added to page head, which improves the appea...

Steven-Eardley:
...rance of the dialogue when the site is shared on social media.

Added some text copied from the index page as the site tagline.

Added a new square logo to fit Facebook's share box.

issue #455


portality/templates/head.html
```diff
@@ -7,9 +7,14 @@
     <meta name="author" content="{{ app.config['ADMIN_NAME'] }}">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-
     <link href="{{ url_for('static', filename='doaj/images/favicon.ico') }}" rel="icon" type="image/x-icon" />
 
+    <!-- Open Graph tags for Facebook -->
+    <meta property="og:title" content="{{ app.config['SERVICE_NAME'] }}"/>
+    <meta property="og:url" content="{{ app.config['BASE_URL'] }}" />
```
emanuil-tolev: Should this always be the base then? Like if I load the search page (since head.html gets included in every single page load), this should still be https://doaj.org, not https://doaj.org/search ?

Steven-Eardley: yes I believe so; this specifies the 'canonical url' - the entity that collects the 'likes' and whatnot, no matter which individual page is shared.

PR #603, opened 2015-03-14
Feature/articlepage additions

Steven-Eardley:
#540

### DIFF
--- a/portality/app.py
+++ b/portality/app.py
@@ -168,6 +168,16 @@ def utc_timestamp(stamp, string_format="%Y-%m-%dT%H:%M:%SZ"):
     utcdt = datetime(tt.tm_year, tt.tm_mon, tt.tm_mday, tt.tm_hour, tt.tm_min, tt.tm_sec, tzinfo=pytz.utc)
     return utcdt.strftime(string_format)
 
+@app.template_filter('doi_url')
+def doi_url(doi):
+    """
+    Create a link from a DOI.
+    :param doi: the string DOI
+    :return: the HTML link
+    """
+    tendot = doi[doi.find('10.'):]
+    return "<a href='http://dx.doi.org/{0}'>{0}</a>".format(tendot)
+
 @app.before_request
 def standard_authentication():
     """Check remote_user on a per-request basis."""
--- a/portality/templates/doaj/article.html
+++ b/portality/templates/doaj/article.html
@@ -2,16 +2,12 @@
 
 {% block content %}
 
-<script type="text/javascript">
-
-</script>
-
 <div class="row-fluid">
     <div class="span9">
         <h1>{{article.bibjson().title}}</h1>
         <p>
         {% if article.bibjson().get_identifiers(idtype='doi') %}
-            <em>DOI: </em>{{ article.bibjson().get_identifiers(idtype='doi')[0] | doi_url | safe}}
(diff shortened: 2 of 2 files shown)


portality/app.py
```diff
@@ -168,6 +168,16 @@ def utc_timestamp(stamp, string_format="%Y-%m-%dT%H:%M:%SZ"):
...
 
+@app.template_filter('doi_url')
+def doi_url(doi):
+    """
+    Create a link from a DOI.
+    :param doi: the string DOI
+    :return: the HTML link
+    """
+    tendot = doi[doi.find('10.'):]
+    return "<a href='http://dx.doi.org/{0}'>{0}</a>".format(tendot)
```
emanuil-tolev: Why are we stripping the 10.? I assume this works as you've committed it but I've never seen people explicitly remove the part of the handle that identifies it as a DOI handle...

portality/templates/doaj/article.html
```diff
@@ -61,16 +63,21 @@
     <div class="span4">
         <div class="box with-borders" style="height: 120px">
             {% if article.bibjson().journal_title %}
+                {% if article.bibjson().get_identifiers(idtype='pissn') or article.bibjson().get_identifiers(idtype='eissn') %}
+                    <p><strong>Journal</strong>:&nbsp;
+                    <a href={{ "/toc/" + article.bibjson().get_identifiers(idtype='pissn')[0] or article.bibjson().get_identifiers(idtype='eissn')[0]}}>{{ article.bibjson().journal_title }}</a></p>
```
emanuil-tolev: Wait what :D? So this will output /toc/pissn if a pissn is present, or if it isn't, it'll do /toc/eissn? Never seen an or in a jinja expression evaluation (rather than control flow block, with {% %}).

emanuil-tolev:
Just a few minor comments. Sorry for the late review, I'll rejig my inbox settings a bit.

portality/app.py
Steven-Eardley: I was replicating the following from facetview ```facetview_results_render_callbacks.js```, which I think drops anything before (but not including) the 10., in case the data has the http://... etc stuff in front of it:

``` javascript

fv_doi_link = (function (resultobj) {
    var that = function(resultobj) {
        if (resultobj.bibjson && resultobj.bibjson.identifier) {
            var ids = resultobj.bibjson.identifier
            for (var i = 0; i < ids.length; i++) {
                if (ids[i].type === "doi") {
                    var doi = ids[i].id
                    var tendot = doi.indexOf("10.")
                    var url = "http://dx.doi.org/" + doi.substring(tendot)
                    return "<a href='" + url + "'>" + doi.substring(tendot) + "</a>"
                }
            }
        }
        return false
    };
    return that;
})();
```

portality/templates/doaj/article.html
Steven-Eardley: Well I think it worked when I tested it! Feel free to check out the branch first and have a go, or ask me to provide proof.

emanuil-tolev: I was asking since I wasn't sure it even looked right, but now have tested it and it does work! Fantastic Jinja find.

portality/app.py
emanuil-tolev: Fair enough, best if it's a similar display. If that works this'll work.

PR #614, opened 2015-03-20
fix xml ingester by not printing remote filenames

emanuil-tolev:


emanuil-tolev:
@Steven-Eardley the change itself is simple enough, but I wanted a 2nd opinion on whether anything else in the script might cause it to break in a similar manner. I reckon not, having seen all the print statements, but this is the 2nd time I'm dealing with this issue in this script so clearly I thought this a few months ago too :), and it needs a 2nd person to look over it.

emanuil-tolev:
(build fails because of ES 1.4 on codeship)

Issue #542, opened 2015-01-24
OAI DC metadata review


richard-jones:
- [x] repoint dc:identifier to journal toc
- [x] repoint dc:identifier to article page (which doesn't yet exist)
- [x] On article OAI data, put link to journal TOC page in dc:relation

emanuil-tolev:
Simple / Basic DublinCore - just check if the 15 fields are populated appropriately with the data we have

emanuil-tolev:
- [ ] Look into specifying vocabulary for where Subject is coming from (currently LCC). Somehow put this info ( http://dublincore.org/documents/dcmi-terms/#H4 ) into the scheme
- [ ] Add keywords to Subject (and tag if possible to show they're human-sourced non-formal)

emanuil-tolev:
Year guessing algorithm may break currently, needs to deal with 2 chars that are not digits (@emanuil-tolev writing it)

emanuil-tolev:
Maybe we should tell people about the changes on the FAQ site and perhaps a blog item. Perhaps even in advance (*gasp*). They're not big changes at all but if anybody's relying on anything (and OAI-PMH is a public interface) it'll change under their feet obviously.

dommitchell:
@emanuil-tolev  Can you review this page http://doaj.org/features and let me know if it needs any changes after your work here?

richard-jones:
@dommitchell - not sure if you've had a chance to review this yet, but I believe the new DC metadata is all now available, so you should be able to check it at (for example):

https://doaj.cottagelabs.com/oai?verb=ListRecords&metadataPrefix=oai_dc

@emanuil-tolev - do you have any outstanding tasks around this still to do?

dommitchell:
Again, not sure what I am looking at. Would be good to go through these two on a Hangout. Also,

> Can you review this page http://doaj.org/features and let me know if it needs any changes after your work here?

richard-jones:
If you go to that link above you should see the OAI Dublin Core output - just check it over to see if the metadata in the XML looks reasonable.

Otherwise, yes, we can talk through on Monday.

dommitchell:
Yes, looks good. Am happy to close this but assigning to @emanuil-tolev just in case this needs changes to /features.

emanuil-tolev:
> Yes, looks good. Am happy to close this but assigning to @emanuil-tolev just in case this needs changes to /features.

It did, I'll assign back for docs review after PR is merged.

[...]

PR #616, opened 2015-03-22
Changes that came out of OAI PMH review

emanuil-tolev:
#542


emanuil-tolev:
There's no point in this pull request as #617 includes all the work done here. I only opened it since then this work would go in faster (the other PR's quite big), but the other branch is complete and it's going to be easier to review 1 PR than 2. So closing this one.

PR #619, opened 2015-03-24
add first draft of subject browser

richard-jones:
I've gone for a jstree based approach, since once I made a start trying to lay out the subjects it was clear that the varying depths of the tree at different points would make a static layout prohibitively difficult.

Instead the solution is more javascript oriented, and reactive to the user.  Hopefully it is the right kind of thing.

### DIFF
--- a/portality/templates/_js_includes.html
+++ b/portality/templates/_js_includes.html
@@ -16,15 +16,18 @@
 <!-- get select2 -->
 <script type="text/javascript" src="/static/portality/vendor/select2-release-3.2/select2.min.js"></script>
 
-{% if search_page %}
+{% if search_page or subject_page %}
 <!-- get facetview -->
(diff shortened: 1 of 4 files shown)


portality/templates/_js_includes.html
```diff
@@ -16,15 +16,18 @@
 <!-- get select2 -->
 <script type="text/javascript" src="/static/portality/vendor/select2-release-3.2/select2.min.js"></script>
 
-{% if search_page %}
+{% if search_page or subject_page %}
```
emanuil-tolev: We should go for some kind of declarative approach to JS dependencies in the future, all these if-s are getting a bit iffy, now nested I see :). (but not now in this PR)

emanuil-tolev:
ok, perfect, I tried it, looks very nice

PR #621, opened 2015-03-25
updated doaj view blueprint so that toc calls Articles directly

markmacgillivray:
No need for ToC index now. But still uses the JournalVolumeToc model. If ToC is not needed anywhere else then that model should be simplified down to a shell model as there will be no index to send data to.

### DIFF
--- a/portality/models/article.py
+++ b/portality/models/article.py
@@ -437,8 +437,12 @@ class ArticleBibJSON(GenericBibJSON):
 class ArticleQuery(object):
     base_query = {
         "query" : {
-            "bool" : {
-                "must" : []
+            "filtered": {
+                "filter": {
+                    "bool" : {
+                        "must" : []
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


emanuil-tolev:
Just need to check for performance but don't have a suitably big dataset here and travelling so @richard-jones might be able to help review that aspect, he's probably got the whole index locally in ES1.4 .. which I should really get too. It does work, as in, displays tocs.

markmacgillivray:
I changed the article queries to be filters instead of queries, as there
was no need for the query. If performance is a problem then instead of
building the table on the server side the articles could be thrown to the
UI and built in JS on the template. I did not do that yet because I was
trying to keep it as close to the old version as possible. As far as I can
see this process does not issue any more queries than the old process does,
and as they are now filters on quite specifically defined fields and
values, they should have very little impact. So I would suspect any
performance issues to be in the python code that builds the journaltoc
objects rather than in querying the index.

This is something I was going to check by having it accepted to the develop
branch and then by looking on the infrastructure server when I issue some
requests, but if you require performance to be checked before accepting the
pull request, then I'll just wait and assume performance is fine if
accepted.

emanuil-tolev:
> This is something I was going to check by having it accepted to the develop
> branch and then by looking on the infrastructure server when I issue some
> requests, but if you require performance to be checked before accepting the
> pull request, then I'll just wait and assume performance is fine if
> accepted.

All I was gonna do is load some tocs and check it's not ridiculous, like half a minute :). Also if it were ridiculous, git checkout develop - is it still that bad? I'll probably do that at the next station where I can just open doaj.org/toc/journal_id and localhost:5004/toc/journal_id and compare!

emanuil-tolev:
Hmm.

I get
http://doaj.org/toc/1307-4733 - 400 ms
http://localhost:5004/toc/1307-4733 - 2750 ms

markmacgillivray:
Oh right, I did that already when building the feature. I took a journal
with 250 or so articles in it, over seven years and about 40 issues. It was
pretty much instant, and responded faster than from doaj.org for the same
TOC.

emanuil-tolev:
ah cool, it's just my connection to remote ES then

markmacgillivray:
I used acta polytechnica hungarica

emanuil-tolev:
Loaded 1307-4733 and all its articles locally, no problems. Merging.

PR #625, opened 2015-03-27
Feature/uitweaks

richard-jones:
A whole grab-bag of changes to the UI, along with some supporting work in loading and working with LCC and also working towards a generic upgrade process that we can build for each release.


emanuil-tolev:
Build failing? Not sure why, ES 1.4 is on Codeship. Also the develop branch succeeded 2h ago, so it seems to be this branch that's failing specifically :).

richard-jones:
The build is allegedly failing, but I ran all the tests locally before submitting, and I've just re-run the supposed broken one, and all is working fine

richard-jones:
Could be an ES race-condition on the set-up - the tests now load the LCC classifications at start up, because they are required by a number of the models.  If they are not refreshed by the time they need to be used, that could be an issue.

emanuil-tolev:
Does lcc setup in the tests need to sleep a sec then? It'd only affect tests which use the LCC setup func.

richard-jones:
Yes, but it already sleeps.  The problem is that the sleep time is fixed, but the time ES takes to refresh is variable.  Sometimes the sleep time will be insufficient.

I'm just adding a feature which confirms the test index is ready to go before proceeding.

richard-jones:
Something is not right with the codeship deployment.  The number of tests failing (22) is the number that were failing for me /before/ I fixed the index deployment with the LCC data.

emanuil-tolev:
Yeah, I guess so, though develop continues to succeed. I've increased sleep after ES startup to make sure it's fully up. There must be a reason in the branch though, we'll find out..

richard-jones:
Ok, maybe a reason in the branch, but if the tests run locally but not on codeship that would seem to indicate a problem with codeship.

emanuil-tolev:
Yeah, weird, since it also blanks the index right? But I'm seeing some 400s by ES (Bad Request) as LCC is being loaded, which are changes introduced here. I'll try this myself on my machine and fix it, then merge.

emanuil-tolev:
Tests fail for me locally @richard-jones . In the process of fixing.

emanuil-tolev:
also traceback

```
ERROR: test_01 (test_article_match.TestArticleMatch)
----------------------------------------------------------------------
  File "/home/emanuil/cl/doaj/src/doaj/doajtest/unit/test_article_match.py", line 19, in test_01
    a.save()
  File "/home/emanuil/cl/doaj/src/doaj/portality/models/article.py", line 285, in save
    self._generate_index()
  File "/home/emanuil/cl/doaj/src/doaj/portality/models/article.py", line 253, in _generate_index
    classification_paths = lcc.longest(classification_paths)
AttributeError: 'NoneType' object has no attribute 'longest'
```

emanuil-tolev:
The problem is portality.lcc:131

```lcc = LCC.pull('lcc')```

What if "lcc" is not defined in elasticsearch, like in a fresh new index? (Like in codeship, or apparently mine.)

I've got a branch with a fix. @richard-jones I have made a PR against *this* branch. If you like the changes in #630 , hit Merge on it, wait for the build in this PR to turn green, and we can merge this one too!

PR #630, opened 2015-03-30
load up the LCC classification if it is not loaded on first use

emanuil-tolev:


richard-jones:
I'm confused as to why this is any different to the existing solution, though it obviously does work.

emanuil-tolev:
Yeah, I hadn't seen your solution in helpers.py. No idea why this one works better.

You shouldn't need to wait after creating an item in the index btw (like your solution in the test setup was doing) - you have to wait for it to be available for *searching* but you can .pull (i.e. GET /index/type/id) immediately. See http://www.elastic.co/guide/en/elasticsearch/reference/1.4/docs-get.html#realtime .

richard-jones:
Ah, in which case my guess is that the save() doesn't work immediately
after index creation - probably the sleep needs to be between create_index
and LCC.save().  Your code then works because there is a sleep in the test
init method, before your loadLCC call is run.

On 30 March 2015 at 09:20, Emanuil Tolev <notifications@github.com> wrote:

> Yeah, I hadn't seen your solution in helpers.py. No idea why this one
> works better.
>
> You shouldn't need to wait after creating an item in the index btw (like
> your solution in the test setup was doing) - you have to wait for it to be
> available for *searching* but you can .pull (i.e. GET /index/type/id)
> immediately. See
> http://www.elastic.co/guide/en/elasticsearch/reference/1.4/docs-get.html#realtime
> .
>
> —
> <https://github.com/DOAJ/doaj/pull/630#issuecomment-87588613>.
>



-- 

Richard Jones,

Founder, Cottage Labs
w: http://cottagelabs.com

PR #633, opened 2015-03-30
Feature/openurl

Steven-Eardley:
This is the bulk of the openurl work; some tweaks may follow but they'll benefit from being merged to develop first.


portality/templates/openurl/400.html
```diff
@@ -0,0 +1,15 @@
+{% extends "/base.html" %}
```
emanuil-tolev: Hmm. 400? That's Bad Request (so wrong format of supplied info) rather than Not Found (using supplied info) I'd think.

portality/view/openurl.py
```diff
@@ -0,0 +1,92 @@
...
+from portality.models import OpenURLRequest
+from urllib import unquote
+
+blueprint = Blueprint('openurl', __name__)
+
+@blueprint.route("/openurl", methods=["GET", "POST"])
+def openurl():
+
+    # Drop the first part of the url to get the raw query
+    url_query = request.url.split(request.base_url).pop()
```
emanuil-tolev: problem with request.query_string?

portality/view/openurl.py
```diff
@@ -0,0 +1,92 @@
...
+        abort(400)
+
+def parse_query(url_query_string):
+    """
+    Create the model which holds the query
+    :param url_query_string: The query part of an incoming OpenURL request
+    :return: an object representing the query
+    """
+    # Check if this is new or old syntax, translate if necessary
+    match_1_0 = re.compile("url_ver=Z39.88-2004")
```
emanuil-tolev: Is that supposed to be a raw string match? If so I'd recommend just doing ```"url_ver=Z39.88-2004" in url_query_string``` on the line below rather than compile a RE. Performance is one reason, but more importantly things like the fact that the dot will match any character. So this should be ```"url_ver=Z39\.88-2004"``` as a RE.

Also if compiling RE-s, best to put them just under module imports up at the top. That way they'll be compiled on import and available for use throughout the lifetime of the web app thread.

portality/view/openurl.py
```diff
@@ -0,0 +1,92 @@
...
+    if not match_1_0.search(url_query_string):
+        print "Legacy OpenURL 0.1 request: " + unquote(request.url)
```
emanuil-tolev: app.logger.info or .warn or .error or even .debug may be best. from portality.core import app if not imported to get access to the logger. It'll give the message a timestamp too.

Also URL might contain unicode, I think? Probably good to take care of that here. If it's already a decoded string (so not a unicode obj) which it might well be (need to see Flask docs), then we can print safely.

portality/view/openurl.py
```diff
@@ -0,0 +1,92 @@
...
+        return old_to_new()
+
+    print "OpenURL 1.0 request: " + unquote(request.url)
```
emanuil-tolev: same unicode note here

portality/view/openurl.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Pack the list of parameters into a dictionary, while un-escaping the string.
+    dict_params = {rem_ns(key): value for (key, value) in request.values.iteritems()}
+
+    # Create an object to represent this OpenURL request.
+    try:
+        query_object = OpenURLRequest(**dict_params)
+    except:
+        query_object = None
+        print "Failed to create OpenURLRequest object"
```
emanuil-tolev: logger

portality/view/openurl.py
```diff
@@ -0,0 +1,92 @@
...
+    # In OpenURL 0.1, jtitle is just title. This function substitutes them.
+    sub_title = lambda x: re.sub('^title', 'jtitle', x)
+
+    # Add referrent tags to each parameter, and change title tag using above function
+    rewritten_params = {"rft." + sub_title(key): value for (key, value) in request.values.iteritems()}
+
+    # Add the rewritten parameters to the meta params
+    params.update(rewritten_params)
+
+    redirect_url = url_for('.openurl', **params)
```
emanuil-tolev: really nitpicky so feel free to ignore, but no need for the redirect_url var, we can just ```return [its contents]```

portality/view/openurl.py
```diff
@@ -0,0 +1,92 @@
...
+        if results['hits']['hits'][0]['_type'] == 'journal':
+            return url_for("doaj.toc", identifier=results['hits']['hits'][0]['_id'])
+        elif results['hits']['hits'][0]['_type'] == 'article':
+            return url_for("doaj.article_page", identifier=results['hits']['hits'][0]['_id'])
+    else:
+        # No results found for query
+        return None
+
+@blueprint.errorhandler(400)
+def bad_request(e):
```
emanuil-tolev: ok, if that is the agreed thing to do (bad request rather than not found for supplied openurl params) then let's do that, but I'd do a (possibly openurl customised) 404 page otherwise.

emanuil-tolev:
Very very nice implementation of a pretty big feature! Minor things noted (unavoidable with so much added code really) and on we'll go!

Steven-Eardley:
Brilliant, thanks for looking through that in detail. I’ll look into your points shortly!

Steve

portality/templates/openurl/400.html
Steven-Eardley: Reconsidered this. Changed.

portality/view/openurl.py
Steven-Eardley: Useful! thanks.

portality/view/openurl.py
Steven-Eardley: Much simpler, ta

portality/view/openurl.py
Steven-Eardley: changed throughout

portality/view/openurl.py
Steven-Eardley: the var was left over from printing to debug. removed.

portality/view/openurl.py
Steven-Eardley: I've reconsidered - 404 does make more sense - I just hadn't checked 400's full meaning. Changed.

Steven-Eardley:
@emanuil-tolev ready for you to have another look :)

PR #638, opened 2015-04-08
Feature/uitweaks4

richard-jones:
Various minor tweaks, plus an update to the upgrade script that will hopefully help us find out what went wrong during the test upgrade.


emanuil-tolev:
:+1:

PR #642, opened 2015-04-09
changes to routing to allow continuations to be displayed on tocs, and a...

richard-jones:
...ctive navigation in the page between them


richard-jones:
also cheekily add whitespace stripping in xml importer

Issue #180, opened 2014-05-12
add doajArticle OAI format


richard-jones:
In the previous version of the system there was a doajArticle format, which has been lost in the migration to the new system.  This is essentially a doaj-native XML format (probably the same as the import format), and some users would like it back.

emanuil-tolev:
Make sure to use xml escaping stuff that @richard-jones did for OAI. Use settext() instead of .text=

richard-jones:
There's a bit more to this, too, that I'd forgotten to put down here.

We will also need to update the metadata formats response, since now we support more than oai_dc but only for the article.

If you want to assign this to me, @emanuil-tolev, when you have done the XML export, I can update the list of metadata formats - would be quicker than me explaining what that means, and how to do it :)

emanuil-tolev:
@richard-jones what is the XML namespace that this sits in?

like when you define it in the settings

```
        "metadataPrefix": "oai_doaj",
        "schema": "https://doaj.org/static/doaj/doajArticles.xsd",
        "metadataNamespace": ""
```

OAI_DC has http://www.openarchives.org/OAI/2.0/oai_dc/ but we don't really have a page that describes it. I guess we've got http://doaj.org/features and this is where a friendlier explanation than the XSD itself would eventually go, but it's not a very unique URL. For now I'm using http://doaj.org/features/oai_doaj/1.0 (and will make a veeery simple page at that URL which just points out where the schema is, and refers to doaj.org/features where the schema is described for *importing* stuff). Any other ideas here?

emanuil-tolev:
For intermediate discussion here, this is what /oai?verb=ListMetadataFormats responds with now:

```xml
<?xml version="1.0"?>
<OAI-PMH xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.openarchives.org/OAI/2.0/" xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/ http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd">
<responseDate>2015-03-22T12:06:27Z</responseDate>
<request verb="ListMetadataFormats">http://localhost:5004/oai</request>
<ListMetadataFormats>
  <metadataFormat>
    <metadataPrefix>oai_dc</metadataPrefix>
    <schema>http://www.openarchives.org/OAI/2.0/oai_dc.xsd</schema>
[...]

PR #645, opened 2015-04-10
add supporting documentation for changes to OAI endpoint

emanuil-tolev:


emanuil-tolev:
I should probably add a word or two to describe: this updates the /features page with all changes from #180 and #542 . In particular it also introduces revision history for the 2 OAI endpoints so we can record changes we make in the future. I already make use of the new revision tables to explain the changes to subject classification and other minor changes.

I've based the info in the revision history on actual code committed, not decisions reached in the issue (ideally they'll be the same ofc.)

PR #653, opened 2015-04-13
minor fixes to revision history on OAI docs page

emanuil-tolev:


emanuil-tolev:
@richard-jones @Steven-Eardley very minor updates here. This is to close #542 when merged/reviewed by Dom.

Issue #652, opened 2015-04-13
OpenURL error page


dommitchell:
Saw that this page is produced when OpenURL fails

![openurl_error](https://cloud.githubusercontent.com/assets/5955396/7113260/3decefd2-e1d2-11e4-9c03-cb03f7d4a091.jpg)

e.g. https://doaj.cottagelabs.com/openurl?rft_val_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Ajournal&rft.spage=41&rft.issue=2&rft.volume=4&rft.date=2010&url_ctx_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Actx&url_ver=Z39.88-2004&rft.issn=19909128&rft.genre=article

Is this message entirely correct? "Check the parameters or try reducing search constraints." doesn't apply to OpenURL does it?

Steven-Eardley:
I suppose it pertains more to our implementation that OpenURL's philosophy - from our perspective it's a search because we run it through an ElasticSearch query, so if you've over-specified the metadata it might help if you remove some. E.g. if you are finding the resource with ISSN and alternative title, but the alternative title isn't in our index. 

From an OpenURL perspective, a URL is meant to reach a resource and there's no specific manner in which it must happen. It's not a search; it's fetching a resource. There'll be a better way of phrasing the message, for sure.

However, the user's best action for a chance of finding something is still to discard parameters from the URL - I'm not sure we can categorically say it's not there due to the uncertainty of a ranked search.

dommitchell:
@Steven-Eardley Yes, I see. I htink we can just delete "Check the parameters or try reducing search constraints." because the page then goes on to say do a search which I think is the correct process.

Sound good?

Steven-Eardley:
Yup, I'll take that out as I'm writing my documentation.

Steven-Eardley:
My changes are visible on the test server; here's a bogus url to see the page.
http://testdoaj.cottagelabs.com/openurl?rft.issn=invalid_issn

dommitchell:
Looking good!

PR #657, opened 2015-04-15
Feature/openurl toc fixes

Steven-Eardley:
Added date parsing, and a query to check if a TOC exists before routing there. Basic documentation; and text changes for issue #652.

### DIFF
--- a/portality/models/openurl.py
+++ b/portality/models/openurl.py
@@ -69,12 +69,6 @@ class OpenURLRequest(object):
         # Set i to use either our mapping for journals or articles
         i = SUPPORTED_GENRES.index(getattr(self, 'genre').lower())
 
-        # FIXME: you can query a journal by volume and issue (which makes sense) but
-        # these values are not used to query the index, so you can get a positive match
-        # for a volume and issue that aren't held by DOAJ.  Since these input values
-        # are then used to point the user to the right endpoint in the journal, this can result
-        # in unexpected errors/behaviour down the line
-
         # Add the attributes to the query
         for (k, v) in set_attributes:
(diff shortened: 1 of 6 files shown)


portality/models/openurl.py
```diff
@@ -124,11 +119,32 @@ def get_result_url(self):
...
-                # for sure at this point whether the user will hit a valid ToC
-                # construct the toc url using the ident, plus volume and issue if they are available
-                jtoc_url = url_for("doaj.toc", identifier=ident, volume=self.volume, issue=self.issue if self.volume else None)
-
+                # If there request has a volume parameter, query for presence of an article with that volume
+                if self.volume:
+                    volume_query = deepcopy(TERMS_SEARCH)
+                    volume_query["size"] = 0
+
+                    issn_term = { "term" : { "index.issn.exact" : ident} }
```
richard-jones: A couple of things could go wrong here that we should look at:

1. ident might not be an issn.  For some unknown reason we might have wound up with a journal id, in which case this match may fail when it should have succeeded.

2. Irrespective of (1), journal continuations may have up to 2 issns, and articles may list one or both of them, so a match for one issn won't cover all possible cases.  We need to ask the bibjson for the specific continuation we are interested in for all of its issns (bibjson.issns()), and use these here as part of a "terms" query (not a "term" query).

richard-jones:
Nearly there, just one outstanding issue around the issns used to query the articles

Steven-Eardley:
@richard-jones this is working fine now, although there's more that could be done - see line note.

richard-jones:
Looks good, merged

PR #658, opened 2015-04-15
Feature/uitweaks6

richard-jones:
Minor changes to toc and search display for date added to doaj

Also adds offline mode for developers (loads javascript from local static directory - doesn't work perfectly, but good enough for train journeys most likely), and doaj version query arg to all css and javascript requests to overcome browser caching.


portality/settings.py
```diff
@@ -1,5 +1,9 @@
 import os
 
+DOAJ_VERSION = "2.10.1_SNAPSHOT-1"
```
emanuil-tolev: This should eventually just be the current commit - ideally it will be the last commit / timestamp when a static resource last changed, but that's too complex. My point is that then devs don't have to know / remember about bumping the version when they edit a static file, every commit will bust the cache automatically. Doesn't prevent merging this now ofc,.

emanuil-tolev:
quick rebase against develop? that readonly branch I merged may have touched the same code as this

I'll do after infrastr shenanigans if you can't do it now, then it'll be mergeable again.

richard-jones:
rebased - will merge when the build finishes

PR #675, opened 2015-04-20
Feature/pvod review

richard-jones:
A whole range of changes after the PVOD review last week:

* change migrate script to include suggestions
* admin area facetview reviews
* fix link between user view and journal view
* publisher facet fix
* site stats generation bug
* updated read-only mode script
* unicode fix on language display on toc

### DIFF
--- a/portality/models/journal.py
+++ b/portality/models/journal.py
@@ -838,6 +838,20 @@ class JournalBibJSON(GenericBibJSON):
         if self.language is not None:
             langs = self.language
         langs = [datasets.name_for_lang(l) for l in langs]
+
+        # now we need to ensure that these are all correctly unicoded
+        def to_utf8_unicode(val):
+            if isinstance(val, unicode):
+                return val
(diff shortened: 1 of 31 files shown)


emanuil-tolev:
Additionally

> @emanuil-tolev can you cherry pick this commit for the production branch: ae2fe8c

portality/models/journal.py
```diff
@@ -838,6 +838,20 @@ def language_name(self):
...
+
+        # now we need to ensure that these are all correctly unicoded
+        def to_utf8_unicode(val):
+            if isinstance(val, unicode):
+                return val
+            elif isinstance(val, basestring):
+                try:
+                    return val.decode("utf8", "strict")
+                except UnicodeDecodeError:
+                    raise ValueError(u"Could not decode string")
```
emanuil-tolev: This could error out quite easily, but we'll catch it when it gets reported I guess.

Issue #178, opened 2014-05-12
Embedded metadata + enhance search results


richard-jones:
see also #12 

Is there anyway to improve the display of search results now that the effects of the sitemap are being seen in Google? For example, a search on an article returns: http://bit.ly/1iQ2SKx where the 3rd result down shows the volume number, 6, and DOAJ.

Ideally this would show the name of the journal and say 'Volume 6' (and DOAJ if possible).

Is this related to our future development request for structured metadata perhaps?

Thoughts?

richard-jones:
* Embedded metadata for the ToC pages (use meta tags)
* Urls should contain the journal title (but not have any effect on the routing otherwise)
* Queries for the ISSN should redirect to the canonical url
* Update search interface to point to correct url

emanuil-tolev:
We should use subject and keywords in the meta keywords tag. + a few generic ones like DOAJ

emanuil-tolev:
> Embedded metadata for the ToC pages (use meta tags)

Done.

> Urls should contain the journal title (but not have any effect on the routing otherwise)

Not done. This has proven to be quite sticky to achieve, mostly because we use the URL to also put in other optional args like volume and issue. Basically the only straightforward situation right now that could merit journal title in the URL would be if you also specify the volume and issue, which is a bit pointless.

However, I do have an alternative approach in mind. Putting a separator, like three dashes --- after the ISSN or internal ID in the URL, and then putting in a slug of the journal's (or article's) title.

http://doaj.org/toc/1307-4733---Novitas_ROYAL - obvious
http://doaj.org/toc/1307-4733---Novitas_ROYAL/6 - vol. 6 of the same journal
http://doaj.org/toc/1307-4733---Novitas_ROYAL/6/8 - vol. 6, iss. 8 of the same journal

This requires a bit more trickery and is outside the scope of the original issue for getting basic SEO in so Google doesn't display gibberish, but I think has some value as one of the 3 main things search engines look for. As it's both worth introducing and outside of scope I'll try to find some time this or next week to do it but no guarantee. It can be easily hotfixed in, in as much as the work is non-trivial (need to slugify titles in python and JS and modify the /toc route), but once "done" is very self-contained and can be rolled out on its own at any time.

> Queries for the ISSN should redirect to the canonical url

> Update search interface to point to correct url

emanuil-tolev:
@dommitchell you can observe this e.g. here: https://doaj.org/toc/1980-6108

[...]

PR #678, opened 2015-04-20
ToC SEO

emanuil-tolev:
Introduce title and description meta tags for journal *and* article pages (bonus!). Keywords tag is no longer used and in fact penalized by Bing so not adding that.

Reroute all ToC pages to use ISSNs. The canonical URL is the one with the ISSN now, but an internal ES hex id will be accepted if no ISSN can be provided. Continuations should continue to work. All search results, sitemap, etc. have adopted the new approach. Finally, E-ISSNs are preferred over P-ISSNs where available.

### DIFF
--- a/portality/formcontext/formcontext.py
+++ b/portality/formcontext/formcontext.py
@@ -541,7 +541,7 @@ class ManEdApplicationReview(ApplicationContext):
             j.save()
 
             # record the url the journal is available at in the admin are and alert the user
-            jurl = url_for("doaj.toc", identifier=j.id)
+            jurl = url_for("doaj.toc", identifier=j.toc_id)
             if self.source.current_journal is not None:
                 self.add_alert('<a href="{url}" target="_blank">Existing journal updated</a>.'.format(url=jurl))
                 #self.source.remove_current_journal()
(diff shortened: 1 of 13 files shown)


emanuil-tolev:
Issue #178 .

emanuil-tolev:
fyi @richard-jones @Steven-Eardley , last (expected) PR for PVOD :)

richard-jones:
When I try to visit a ToC for a journal with both print and electronic issns I get a redirect loop!

emanuil-tolev:
Hm, I thought I took care of that. I'll double check my test data and code, thanks.

emanuil-tolev:
Yep, I thought I had tested it but in fact had never left my test journal with both ISSNs defined.

I've got a journal like this:
pissn: 1307-4733
eissn: 6954-854X
id: e52d9e100f5147edbf65b756f1e9bab6

/toc/1307-4733 redirects to /toc/6954-854X
/toc/e52d9e100f5147edbf65b756f1e9bab6 redirects to /toc/6954-854X
/toc/6954-854X displays the ToC

all the links on the ToC itself (like issues, volumes), and in the search results are to /toc/6954-854X

emanuil-tolev:
updated

richard-jones:
ok, that seems to have fixed it.

PR #691, opened 2015-04-23
correct usage of application statuses in the formcontexts

richard-jones:


emanuil-tolev:
@richard-jones rolled out

Issue #696, opened 2015-04-26
Continuations not working on live


Steven-Eardley:
TOCs for journal continuations are being redirected to the most recent version - i.e. the previous title link has no effect, and OpenURL queries return the wrong TOC. Additionally, it redirects twice; with ```http``` then ```https```.

Might be worth going over changes made in PR #678 again, @emanuil-tolev? Could be something more recent, though.

* https://doaj.org/openurl?url_ver=Z39.88-2004&url_ctx_fmt=info:ofi/fmt:kev:mtx:ctx&rft_val_fmt=info:ofi/fmt:kev:mtx:journal&rft.genre=journal&rft.issn=0366-1644
* https://doaj.org/toc/0366-1644

emanuil-tolev:
sort of unrelated note: why doesn't this cause a whitespace compile error? https://github.com/DOAJ/doaj/pull/678/files#diff-15fcdea4c39a69a3a91220826502d799R165

(just wondering if it WILL cause an error... should fix that together with whatever the fix is for this issue)

emanuil-tolev:
> https://doaj.org/toc/0366-1644

OK so for this: that ISSN is in the index.issn part of the record, which is how the find_by_issn method works. This is the record it's finding: https://doajgate.cottagelabs.com/doaj/journal/30a6fbcbc2a241ffbb376da7be41b29c

Afterwards it simply redirects to the current E-ISSN of the journal. It assumes the ISSN in the request is assigned to the *current* journal record (because it found that with the find_by_issn method...), and redirects you to the E-ISSN of the current journal record (which is what it's supposed to do).

So, not sure why that 0366-1644 ISSN is recorded in the current record. Should it be? Do we want to find records by their past issns when calling find_by_issn? It might be easiest for me to talk this over with @richard-jones , perhaps at the Fri meeting..

> Additionally, it redirects twice; with http then https.

Not sure why the first redirect is happening, it should just direct you to the HTTPS page. The 2nd redirect is just because we enforce HTTPS everywhere (except the doajArticles XSD schema, apparently), that's the right behaviour there.

Steven-Eardley:
> https://doaj.org/toc/0366-1644
OK so for this: that ISSN is in the index.issn part of the record, which is how the find_by_issn method works. This is the record it's finding: https://doajgate.cottagelabs.com/doaj/journal/30a6fbcbc2a241ffbb376da7be41b29c

> Afterwards it simply redirects to the current E-ISSN of the journal.

[...]

Issue #694, opened 2015-04-23
OpenURL does not respect in_doaj


Steven-Eardley:
OpenURL requests are successfully finding the item in the journal index on search, redirecting to the TOC page, which doesn't render (because in_doaj is false) so the normal 404 page is shown. 

e.g. https://doaj.org/doaj?func=openurl&genre=journal&issn=15593053

Fix with a filter for ```in_doaj = True```

richard-jones:
@Steven-Eardley - is this now working in live?

Steven-Eardley:
Yes, the OpenURL part works as intended. However, I might've uncovered another issue with the TOC!

PR #697, opened 2015-04-26
Hotfix: openurl in_doaj

Steven-Eardley:
**MERGE OFFLINE**
This prevents OpenURL for asking for TOCs that won't render because their target has ```in_doaj``` set to ```false```. For Issue #694.

N.B. this PR should work properly once #696 has been resolved.

### DIFF
--- a/portality/models/openurl.py
+++ b/portality/models/openurl.py
@@ -29,8 +29,9 @@ OPENURL_TO_ES = {
     'isbn' : ('index.issn.exact', 'index.issn.exact'),
 }
 
-TERMS_SEARCH = { "query" : {"bool" : { "must" : [] } } }
+# Terms search template. Ensure all queries from OpenURL return publicly visible results with in_doaj : true
 IN_DOAJ_TERM = { "term" : { "admin.in_doaj" : True } }
+TERMS_SEARCH = { "query" : {"bool" : { "must" : [ IN_DOAJ_TERM ] } } }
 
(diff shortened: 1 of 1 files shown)


emanuil-tolev:
> N.B. this PR should work properly once #696 has been resolved.

ok .. so, should this be merged / deployed then? Should we wait till 696 is resolved?

portality/models/openurl.py
```diff
@@ -29,7 +29,9 @@
     'isbn' : ('index.issn.exact', 'index.issn.exact'),
 }
 
-TERMS_SEARCH = { "query" : {"bool" : { "must" : [] } } }
+# Terms search template. Ensure all queries from OpenURL return publicly visible results with in_doaj : true
+IN_DOAJ_TERM = { "term" : { "admin.in_doaj" : True } }
+TERMS_SEARCH = { "query" : {"bool" : { "must" : [ IN_DOAJ_TERM ] } } }
```
emanuil-tolev: @richard-jones I've noticed you do filtered queries, should we do this or is this OK?

Steven-Eardley:
We don't need to wait - this will make an improvement ahead of a fix for #696

portality/models/openurl.py
emanuil-tolev: ok, right, we can change this later, it'll work like this so merging

emanuil-tolev:
@Steven-Eardley rolled out!

Issue #698, opened 2015-04-27
RSS feed content links are wrong now - should probably point to ToC


richard-jones:

emanuil-tolev:
Fixed I believe, on live. @richard-jones feel free to close if this looks OK to you: https://doaj.org/feed

richard-jones:
looks good

PR #700, opened 2015-04-28
fix rss feed URLs

emanuil-tolev:
Fix for #698 and ditch that unused feed file which I modified first by mistake :).


Steven-Eardley:
looks good to me - is this a hotfix for immediate deployment?

emanuil-tolev:
Yep, it is, and yep :). Just pull sysadmin repo first to get latest fabric script

PR #701, opened 2015-04-28
Feature/issns for alt title

Steven-Eardley:
* searching for a jtitle will fall-back to trying alternative_title, to handle some title translations.
* show correct vol/issue using stitle, since issns_for_title now also accepts alternative_title so volume queries work.

### DIFF
--- a/portality/models/journal.py
+++ b/portality/models/journal.py
@@ -274,20 +274,28 @@ class Journal(DomainObject):
 
         incoming_title = title.strip().lower()
 
-        # first check the main title on the journal
+        # first check the main title / alt_title on the journal
         current_title = self.bibjson().title
+        current_alt_title = self.bibjson().alternative_title
         if current_title is not None:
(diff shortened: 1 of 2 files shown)


emanuil-tolev:
Not sure if it would have been useful for this open url work but there is a method called q2obj in the dao. Instead of all the checks against results['hits'] key. May not be applicable though since I see _type checks too.

emanuil-tolev:
Leaving open aince best released as hotfix now that that project is done. But +1, feel free to release as hotfix if you get to it before me.

richard-jones:
ok, so this is good to merge?  But just needs to be re-issued as a hotfix?  Who is in charge of this one now?

emanuil-tolev:
> ok, so this is good to merge? But just needs to be re-issued as a hotfix? Who is in charge of this one now?

I'll just convert it to a hotfix locally and roll out I think.

emanuil-tolev:
Rolled out.

PR #704, opened 2015-04-30
only display publication charges if the journal has reapplied

richard-jones:


emanuil-tolev:
Oops, hotfix. No problem though, will just ```git flow hotfix finish``` it locally and this will back-merge it properly into develop!

PR #707, opened 2015-05-04
closes #703

### LINKED ISSUE #703: Bug in Toc generation: page numbers incorrect
Unfortunately the ToCs are showing the incorrect page range. The first page is repeated and the last page is missing. This is not a problem in search results or on the article homepages. Only the ToCs. For example:

'After you launched the new design of your website we noticed that the page range for all published articles become 1-1, would you please check and correct? https://doaj.org/toc/2314-5234'

https://doaj.org/toc/2013-0953

https://doaj.org/toc/2014-4806

screeny shot 27-4-2015 12 26 28

markmacgillivray:


richard-jones:
@emanuil-tolev - can you merge this to production and deploy when you get a moment?

Issue #512, opened 2015-01-12
DOAJ Seal


richard-jones:
Rules: to be displayed for those journals who have had the 'DOAJ Seal' checkbox on their journal page checked (e.g. https://doaj.org/admin/journal/2ae83e2d8df84ada9a7702235d582d4f) 
To be displayed in both:
  -  the 'short view' (search results) and
  -  the 'long view' (journal issue pages), top right hand corner under the CC logo

Short view: 
![doaj_seal_logo_mini](https://cloud.githubusercontent.com/assets/5955396/3014983/d3715da4-df5d-11e3-8194-b2ab5069569d.png)

Long view: 
![doaj_seal_text_logo_mini](https://cloud.githubusercontent.com/assets/5955396/3014984/d9da1640-df5d-11e3-88c1-994056c86d77.png)

richard-jones:
DOAJ Seal:
* Changes to core data model to support new info
* Updates to relevant administrator web forms to allow Seal
to be applied by checkbox
* Updates to search interface to facet by Seal, and present
Seal alongside other Journal data
* Updates to Table of Contents to present Seal
* Essential data migration and roll-out

Steven-Eardley:
@dommitchell, I'd like you to make sure you're happy with how the various badges are placed on the page. So far I have this, after following the instructions above:

![screen shot 2015-05-25 at 14 20 29](https://cloud.githubusercontent.com/assets/1446355/7808780/a1e8fd32-038d-11e5-929f-60162b1935c6.png)

![screen shot 2015-05-25 at 14 20 48](https://cloud.githubusercontent.com/assets/1446355/7808788/a94d5a5a-038d-11e5-991a-c9b0a6ccf7c9.png)

RJ has suggested the following changes:
* in the facetview results use the 'short view' seal and place it under the journal type icon on the left of the text
* on the TOC page, place the seal on the same line as the tick, side-by-side.

I'd just like to know if you have any different suggestions before I go ahead.

dommitchell:
@Steven-Eardley I think I had imagined it as you have it pretty much. I'd be interested to know @richard-jones reasons for the alternative display? Are there things that I am not taking ito consideration?

Can you show me one in facet view, as you have the layout now, and that has the APC charge as well?

I think the Seal should go above the license image and under the Tick in both views.

richard-jones:
[...]

PR #765, opened 2015-06-03
Feature/doaj seal

Steven-Eardley:
Issue #512. This finally passes the tests so I'm fairly confident the underlying code is ready. There are a couple of questions still open about the layout and images, but can be adjusted later if required.


portality/formcontext/forms.py
```diff
@@ -370,6 +370,11 @@ class Workflow(Form):
                     ' completed your review.'
     )
 
+class Seal(Form):
+    """ Field to set the DOAJ Seal """
+
+    doaj_seal = BooleanField('Set DOAJ Seal', [validators.Optional()], description='description for doaj seal field.', false_values=(BooleanField.false_values + (False,)))
```
emanuil-tolev: "description for doaj seal field." ? I don't recall what "description" is used for, wasn't it the little gray text below fields though?

emanuil-tolev:
Excellent I think. I've not tried it myself since tbh I'm unsure of the criteria, and hence the conditions required for a manual test, but as far as reviewing goes it's tip-top - tests, forms, the xwalks, front-end (toc). Just the one minor q on the form field description, just because it looks like placeholder text, after that will merge!

portality/formcontext/forms.py
Steven-Eardley: Ah yes, you've caught one of the placeholders!

Steven-Eardley:
Thanks, will reassign to you when it's ready to merge - I've had more info on the layout changes in the meantime which I'll include in my next commit.

Steven-Eardley:
That's another set of layout changes added

portality/templates/formcontext/maned_journal_review.html
```diff
@@ -95,8 +95,9 @@ <h3 class="heading-aligned-with-fields">Old journal information</h3>
 
         </div>
 
-        {# include the base journal form which contains all the biblio data #}
-        {% include 'formcontext/_base_journal_form.html' %}
+        {# include the base form which contains all the biblio data #}
+        {# we use the base application here because it contains extra information regarding the DOAJ seal #}
```
emanuil-tolev: OK I double checked and at the moment the base journal form does not contain anything the base application form doesn't have as well. That could change however.

Also feels weird to include the base app form in the ManEd *journal* review form.

To tackle both of these "smells", can we refactor the seal to be included in a file separate from the base application file? Just thinking about maintainability here, the journal and application files are driven by quite different needs and changes to either may occur without consulting the other. Making the ManEd journal form **not** based on the base journal form feels quite wrong.

Currently we have ```formcontext/_basic_biblio_fields.html```. How about a ```formcontext/_seal.html``` which is then included in ```formcontext/_base_application_form.html``` as well as ```formcontext/maned_journal_review.html```? Just using an include, the same way basic_biblio_fields is included in both files already.

In other words, instead of using the file which already has what we want, we refactor the seal content out of that file into a separate include, and then very explicitly include the seal content where we want it. The explicitness (yes, it's a word) and the composition together make it clear what is (semantically) supposed to be on the screen and we keep our "basic journal form content is shared by all journals forms" hierarchy in place.

emanuil-tolev:
Hiya - looks good, one non-trivial q about template inheritance (but with a suggestion :)). Since this is urgent to be deployed within the next 2 days feel free to ping me at the sign of trouble - e.g. the includes don't quite work the intended way - it's happened several times to me and took a couple of days to get my head around). I think what I've suggested should work straight off the bat in just a few minutes, but don't hesitate to ping me if it doesn't...

Steven-Eardley:
I agree with your changes. I'll get on it now.

Steven-Eardley:
Done

portality/migrate/seal/seal.json
```diff
@@ -0,0 +1,12 @@
+{
```
emanuil-tolev: Actually, sorry for q after merge but how does this add the new seal field? I understand this just specifies which types to apply the changes to, but where are the changes actually defined? Or is it a purely new field created upon saving an object (in which case why use a migration)?

emanuil-tolev: I just ctrl+F-ed for "migrate" in the PR's code, so apologies if I've missed it. Looking for "seal" is not a very easy endeavour in this set of code for obv reasons :).

emanuil-tolev: Need to .save every object so use a migration to regen .index

PR #779, opened 2015-06-25
modify article upload logic to allow match on only one issn provided …

richard-jones:
…there are no inconsistencies on the matched journals.  Also adds full suite of tests for all eventualities.

@emanuil-tolev - I've done this as a PR against develop, but perhaps it would be better against production, if we can roll features like these out regularly?

### DIFF
--- a/doajtest/fixtures/article.py
+++ b/doajtest/fixtures/article.py
@@ -1,0 +1,33 @@
+import os
+from lxml import etree
+from StringIO import StringIO
+
+RESOURCES = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "unit", "resources")
+ARTICLES = os.path.join(RESOURCES, "article_uploads.xml")
+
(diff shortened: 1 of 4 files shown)


emanuil-tolev:
Yes, I would treat the occasional unrelated feature as a hotfix to be released immediately - merging into production and develop. In that case you can use git flow hotfix start YYYY-MM-DD_issuenum and then just push it up. When making the PR include ```# Hotfix, do not merge in github``` at the top and make it against production (that way the diff will be what's really important - how it merges into production). The reviewer then +1's and merges the hotfix with git flow hotfix finish (this is on the wiki). If there is a problem back-merging into develop (e.g. your new feature conflicts with an ongoing project) then the PR is bumped back to you.

This PR looks good to me. What I will do now is make a new hotfix, merge your feature branch into it, and release it (so retrofit your branch into the process above, easy atm because production and develop are the same). In the future we can follow the above if it sounds simple.

PR #780, opened 2015-06-26
change atom feed crosswalks and add feed test

richard-jones:
@emanuil-tolev - sorry by the time I saw your other email, I'd already started this as a feature. From this point on I'll do them as hotfixes, promise!


emanuil-tolev:
Very nice test coverage :+1:

Issue #754, opened 2015-05-29
API design (CRUD, Bulk)


emanuil-tolev:
also see https://github.com/JiscPER/jper/blob/develop/docs/API.md

emanuil-tolev:
Proposed url design:

- /api
  - /v1
    - GET /search
    - GET /article/{id}
    - POST /articles
    - etc.

There will be an api_directory function in portality.app which will list all API versions and doc and base URLs.

There will be a new portality.views.api_v1 Flask view with all API routes inside. These will be the only public-facing parts of the API and will just deal with converting incoming parameters into objects and converting the responses into suitable form for HTTP (stringified JSON all the time pretty much) as well as converting any errors raised by the API back-end into suitable error codes and error details in JSON format. This is the presentation layer.

The business logic for the API will be in a package: portality.api.v1 . This will in turn contain .common, .discovery, .crud and .bulk modules. Each specific module will contain a class that represents its API and its actions, e.g. discovery.DiscoveryApi, which might in turn have a .search(type, query) classmethod. The classes will be made available at the v1 package level (so from portality.api.v1 import DiscoveryApi would work).

So to recap:
portality.views.api_v1.search is a simple function, a route which takes care of converting HTTP data to Python objects and back. It would call portality.api.v1.DiscoveryApi.search. That method will do the actual look-ups by using the relevant models. Then portality.views.api_v1.search will take the results and make them into JSON to be sent back over the wire.

In order to create a new API version in the future, portality.views.api_v1 will be copied to portality.views.api_v2 and the portality.api.v1 folder will be copied to portality.api.v2 . The new version will be written without any regard for backwards compatibility in the presentation and business logic - the models will be pretty much the only thing always shared by API versions. This allows us to deprecate versions and work on new ones in (relative) peace as the old ones keep working, but it also allows backwards-compatible fixes to current versions if needed.

PR #783, opened 2015-06-26
Feature/754 initial api structure

emanuil-tolev:
You need to turn on the api by overriding FEATURES to ['api'] in your local dev.cfg otherwise you won't get anything from any of the routes because of feature toggling. Let me know if you have any trouble. cc @richard-jones @Steven-Eardley @Nimphal 

Everything is now contained in the proper folders and the whole structure is laid out in actual python rather than my head. Fuller description in #754  .

### DIFF
--- a/portality/api/v1/__init__.py
+++ b/portality/api/v1/__init__.py
@@ -1,0 +1,3 @@
+from portality.api.v1.common import jsonify_models
+
+from portality.api.v1.discovery import DiscoveryApi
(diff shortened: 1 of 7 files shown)


richard-jones:
I've made 2 minor tweaks: adding Flask-Swagger to setup.py (generic version - you may wish to freeze it, but I'll leave that to you to worry about later), and making the root /api route completely optional based on FEATURES.

I'll just wait for the build to finish, then I'll merge this.

emanuil-tolev:
ok, I froze Swagger - don't want the app to fail to start in the future because the API docs library cannot be loaded :) generally for stable apps just freeze everything (or point out specific git commit if repo).

PR #787, opened 2015-07-02
Feature/735 discovery api

richard-jones:
This provides the actual implementation of the public discovery API, but it doesn't really do very much about the documentation.  I'm not sure how that fits in with the swagger stuff - I expect the docs will probably want to contain some examples, a bit of detail about the allowed lucene queries, and possibly references to the data that is produced, but that seems like a lot to go into a single method comment.

### DIFF
--- a/portality/api/v1/discovery.py
+++ b/portality/api/v1/discovery.py
@@ -1,14 +1,166 @@
 from portality.api.v1.common import Api
-
 from portality import models
+from portality.core import app
+from datetime import datetime
+import esprit
+import re, json
+
+class DiscoveryException(Exception):
+    pass
+
+class SearchResult(object):
+    def __init__(self, raw=None):
+        self.data = raw if raw is not None else {}
+
+def substitute(query, substitutions):
+    if len(substitutions.keys()) == 0:
+        return query
+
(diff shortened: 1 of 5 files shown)


portality/api/v1/discovery.py
```diff
@@ -1,14 +1,175 @@
...
+                    },
+                    "query" : {
+                        "query_string" : {
+                            "query" : self.qs
+                        }
+                    }
+                }
+            },
+            "_source": {
+                "include": ["last_updated", "created_date", "id", "bibjson"],
```
emanuil-tolev: hah, very nice

portality/view/api_v1.py
```diff
@@ -1,11 +1,22 @@
...
 
-from portality.api.v1 import DiscoveryApi, jsonify_models
+from portality.api.v1 import DiscoveryApi, DiscoveryException, jsonify_models
 from portality.core import app
 
+import json
+
 blueprint = Blueprint('api_v1', __name__)
 
+def _bad_request(message=None, exception=None):
```
emanuil-tolev: :+1:

emanuil-tolev:
Very cool! Can't find any problems. Let's merge and see it in action then!

PR #788, opened 2015-07-05
Feature/journal csv cherrypicked

Steven-Eardley:
Cherry-picked the updated CSV commits.


emanuil-tolev:
Something weird going on with the .gitignore, not sure why it's now marked as "binary" in this PR. It'd be preferable to be able to diff changes to it rather than have it be treated like a picture. I'll see if I can ferret out which commit changed it and why git is treating it as binary.

Steven-Eardley:
Aye, I saw that - it really is odd. It'll happily preview the file on github, though. Who decides it's binary? On my client side?

emanuil-tolev:
@Steven-Eardley btw there is some feedback in the CSV issue - looks like small changes that you might want to add while we figure this binary thing out. From a brief Google search it looks like UTF-16 characters or extra Linux FS permissions can cause this, so I'd be reluctant to merge it atm until we resolve it, otherwise we might be stuck with an undiffable .gitignore.

emanuil-tolev:
@Steven-Eardley unfortunately I believe .gitignore is already binary in develop from a previous commit - maybe "Adding Mac .gitignore template" by you around 30 March 2015?

I think the problem may be

```
# Icon must end with two \r
Icon^@^@
```

This is what I see in vim, I assume you see something other than ^@ there.

I did the following to revert the file to a text one (just as a test):
1. git checkout develop && git pull
2. git checkout -b tmp
3. delete the ```Icon^@^@``` line
4. git commit -a -m 'test'
5. edit the file again, add a test line at the end
6. git diff works again

So I believe the problem is that Icon^@^@ line. For this PR, you could either just say what you changed (develop already has the problem so you don't *have* to deal with it in this PR), or remove that Icon^@^@ line if it's not immediately necessary on your machine, enabling the web/cmd diff to work again and then I can see what changed myself.

emanuil-tolev:
@Steven-Eardley also, sorry to be a pain, but would you mind doing this as a hotfix against production? That way we can release it this week without releasing API work in progress, but at the same time also integrate it into the develop branch so it doesn't conflict with API work either.

Steven-Eardley:
@emanuil-tolev this commit:[https://github.com/DOAJ/doaj/commit/727f85034b76735c76209f9530370dd8d25bd528] is now showing the line changes rather than a binary file, so hopefully github will understand the .gitignore better once that commit is in. I'll now also reshuffle and create a new PR as a hotfix.

Issue #769, opened 2015-06-10
Enable alt attributes in sponsor images


dommitchell:
Even though I can see the alt attributes in the html code, they are not showing up in SEO browsers:

http://seo-browser.com/index.php?address=https%3A%2F%2Fdoaj.org%2F&button0.x=49&button0.y=11 where only the image names appear.

Can we fix this?

Steven-Eardley:
@dommitchell I've pushed this to live now - ready for review. Thanks for the pointers!

dommitchell:
Checked in an SEO browser. This is great!

PR #791, opened 2015-07-06
Alt image attributes for sponsors

Steven-Eardley:
issue #769


emanuil-tolev:
Obviously :+1: :). For some reason the hotfix branch was not deleted - git flow usually does this when you do git flow hotfix finish, so might be worth double checking that it's all merged ok in production and develop before releasing. Or after!

Issue #742, opened 2015-05-29
API - Documentation framework


emanuil-tolev:

emanuil-tolev:
Docs will ideally be written by whoever is not coding that element. A framework for docs should be put in place and instructions put here to make it easy for anybody coding a part of the API to just add their section in with minimal effort, and still gain any styling, demo functionality (if there is any) and consistency with the other sections.

emanuil-tolev:
If @Nimphal is available (TBA), I just need to do a bit of brainstorming with her about the API docs look, feel, location, then she can take over making a nice foundation page to which we easily add bits.

emanuil-tolev:
> If @Nimphal is available (TBA)

she is working on this now, confirmed

emanuil-tolev:
@Nimphal the next steps after #783 is merged are:

- integrate swagger-ui's dist folder into the DOAJ somehow. We can pair on this if you want, but roughly the whole dist folder should probably go under /portality/static/portality/vendor/swagger-ui . Then the index.html from the dist folder needs to be converted to a DOAJ page if possible - move index.html from the dist folder to a template - portality/templates/api/v1/docs.html might be good (does not exist now). Take care of any CSS and JS includes - the DOAJ's base.html **should** have a way of inserting these already. Then change the portality/view/api_v1.docs view function to render the new template.
- customise the docs like e.g. the colours should probably be DOAJ's and some of the swagger branding will go from the top and bottom. Essentially integrate it into the DOAJ's visual look and feel. I expect this to be quite smooth but if you run into incompatibilities (e.g. DOAJ overrides their styles or vice versa) let me know - there is an option to host this as a totally separate page and just apply DOAJ branding sparsely to the top. We can integrate it "properly" later, for now it's important to just get it working.

FYI I've written a proper swagger config for the search route (not one from a POST example) and the live demos work!! So you can GET /api/v1/search/{search_type}/{search_query} directly from the swagger ui, and it works, which is quite exciting.

emanuil-tolev:
@Nimphal , @richard-jones makes a good point here https://github.com/DOAJ/doaj/pull/787#issue-92634389 . There is usually some space in API docs to have some blurb of text and examples. Overall comments as per call today.

richard-jones:
@emanuil-tolev can this one be closed?

PR #796, opened 2015-07-08
Feature/742 api doc framework

Nimphal:
For #742 

I have put lots of lorem ipsum around as this is just the foundations. Swagger allows a lot of free text in various places as well as response models. I think it covers all our needs neatly.

### DIFF
--- a/portality/templates/doaj/api_docs.html
+++ b/portality/templates/doaj/api_docs.html
@@ -1,0 +1,95 @@
+{% extends "base.html" %}
+
+{% block extra_stylesheets %}
+<!-- get Swagger UI css -->
+    <link href='/static/portality/vendor/swagger-ui/src/main/html/css/typography.css' media='screen' rel='stylesheet' type='text/css'/>
+    <link href='/static/portality/vendor/swagger-ui/src/main/html/css/reset.css' media='screen' rel='stylesheet' type='text/css'/>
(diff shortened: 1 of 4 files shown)


portality/templates/doaj/api_docs.html
```diff
@@ -0,0 +1,95 @@
...
+    <link href='/static/portality/vendor/swagger-ui/src/main/html/css/typography.css' media='screen' rel='stylesheet' type='text/css'/>
+    <link href='/static/portality/vendor/swagger-ui/src/main/html/css/reset.css' media='screen' rel='stylesheet' type='text/css'/>
+    <link href='/static/portality/vendor/swagger-ui/src/main/html/css/screen.css' media='screen' rel='stylesheet' type='text/css'/>
+    <link href='/static/portality/vendor/swagger-ui/src/main/html/css/reset.css' media='print' rel='stylesheet' type='text/css'/>
+    <link href='/static/portality/vendor/swagger-ui/src/main/html/css/print.css' media='print' rel='stylesheet' type='text/css'/>
+    <link href='/static/doaj/css/doaj_api.css' media='screen' rel='stylesheet' type='text/css'/>
+
+{% endblock %}
+
+{% block extra_head %}
```
emanuil-tolev: It'd be nice to include these in the ```extra_js_bottom``` block instead and see if they still work? Literally changing the name of the block from extra_head to extra_js_bottom should work for that. It will put the scripts at the bottom of the page instead of in the head, allowing browsers to start displaying the page and then render the content on. It makes the biggest difference on slow or mobile (synonymous?) connections.

Issue #790, opened 2015-07-06
Update Journal CSV


Steven-Eardley:
Add the extra fields from the reapplication to the CSV created regularly, covering all journals by ISSN.

dommitchell:
Feedback:

- blank line between each entry needs to go
- two missing columns: The Tick, The Seal
- country of publisher and currency needs to be written in full, not the codes

Steven-Eardley:
I'm not seeing any blank lines (libreoffice) - does anyone else see them?

dommitchell:
I am using OpenOffice. I don't see blank lines on the existing CSV that I downloaded from the live site.
![csv](https://cloud.githubusercontent.com/assets/5955396/8521996/d8ccdb8a-23e9-11e5-9ceb-5188bcf9219b.jpg)

Steven-Eardley:
We think the new-lines are a product of me generating the example on my mac - Richard confirmed they were there in the file I emailed, but not present when he generated them on his machine.

So I'll make the other changes then I'll send another sample, this time generated from a system on the same platform as the live site.

Steven-Eardley:
@dommitchell this is now live, can you check there are no blank lines issues with the CSV it's generated?

dommitchell:
Looking good!

dommitchell:
Could you update the column heading for column AW which currently says 'Open Access'? Can it be changed to: 'Does the journal allow readers to 'read, download, copy, distribute, print, search, or link to the full texts' of its articles?'

dommitchell:
@Steven-Eardley  Can you take care of the former comment above and please could you add in the subject classification column, as it is currently missing.

dommitchell:
Hey @Steven-Eardley can I get an update on this one?

dommitchell:
> Does the journal allow readers to 'read, download, copy, distribute, print, search, or link to the full texts' of its articles?

change to:

'Does this journal allow unrestricted reuse in compliance with BAOI?'

emanuil-tolev:
> BAOI

It's BOAI btw :) Budapest Open Access Initiative

dommitchell:
Ha! Cheers!

Steven-Eardley:
@dommitchell shall I include subjects with the schema, as in ```schema_subject```, or without, as in ```classification```? And shall I add another additional column for the codes?

```json
"schema_subject": [
    "LCC:Technology",
    "LCC:Chemical technology",
    "LCC:Biotechnology"
],
[...]

PR #797, opened 2015-07-08
Hotfix/2015 07 08 1145 journal csv

Steven-Eardley:
issue #790


Steven-Eardley:
hmm, the .gitignore is still appearing as binary.

emanuil-tolev:
Weird, last commit shows it ok? Let me check locally too

emanuil-tolev:
Yeah it works locally. I can see text diffs from before the 30 Mar and now I see a text diff of 2 removed lines again. I think it's ok now, probably next PR to change it will work in github.

portality/formcontext/xwalk.py
```diff
@@ -917,7 +917,8 @@ def obj2form(cls, obj):
 
         forminfo['subject'] = []
         for s in bibjson.subjects():
-            forminfo['subject'].append(s['code'])
+            if "code" in s:
```
emanuil-tolev: sure, though I'd expect every subject to have a code

emanuil-tolev:
:+1: . Are you happy to release this? I could do too, or just be around, if you think it'll be troublesome. I guess the thing to watch out for is will the new CSV actually be good on live. deploy/crontab-production-apps tells me that the CSV runs every 30 min, so if you deploy around :50 you'd be able to check in about 10 minutes by going to /csv on the live site.

portality/formcontext/xwalk.py
Steven-Eardley: when I was demonstrating the CSV generation to @richard-jones he got an error on his local index and added this line to fix.

Steven-Eardley:
I'll release this, thanks!

Issue #734, opened 2015-05-29
Update user accounts module


emanuil-tolev:
- new user role "api" - without it you're pretty much banned from the API
- [x] add api_key to user acc
- [x] UI for retrieve/reset api key
- [x] data migration to add api role to all users and generate api keys
- [x] modify existing account creation code to add api role automatically (and maybe add a key automatically - users can make a key themselves, but can't add the role, and we don't want that much manual management)

Steven-Eardley:
@emanuil-tolev close if happy with this :)

emanuil-tolev:
Yep let's close - if problems arise we can reopen. But looks good and it's got tests and a UI and all, so doubt it.

PR #798, opened 2015-07-09
User / authentication 

Steven-Eardley:
for API issue #734:

* added new roles 'user' (top level) and 'api' to allow api access
* added 'api_key' to Account model
* updated account creation to give api role and set api_key by default
* added new decorator for checking requests have active api_key
* updated readme to include new fields in data model
* new tests for api account
* refactored decorators out of portality.core to avoid circular import (including ssl_required etc)
* tidied up some imports
* updated an outdated function call in core.py

### DIFF
--- a/portality/models/account.py
+++ b/portality/models/account.py
@@ -162,7 +162,7 @@ class Account(DomainObject, UserMixin):
     @classmethod
     def pull_by_api_key(cls, key):
         """Find a user by their API key - only succeed if they currently have API access."""
-        res = cls.query(q='api_key:"' + key + '"')
+        res = cls.query(q='api_key.exact:"' + key + '"')
         if res.get('hits', {}).get('total', 0) == 1:
             usr = cls(**res['hits']['hits'][0]['_source'])
             if usr.has_role('api'):
--- a/portality/models/account.py
+++ b/portality/models/account.py
@@ -1,9 +1,10 @@
-from portality.dao import DomainObject as DomainObject
-from flask.ext.login import UserMixin
 import uuid
+from flask.ext.login import UserMixin
 from datetime import datetime, timedelta
+from werkzeug.security import generate_password_hash, check_password_hash
+
+from portality.dao import DomainObject as DomainObject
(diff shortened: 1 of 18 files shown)


portality/models/account.py
```diff
@@ -133,3 +148,24 @@ def set_role(self, role):
...
+        else:
+            return None
+
+    def generate_api_key(self):
+        self.data['api_key'] = uuid.uuid4().hex
+
+    @classmethod
+    def pull_by_api_key(cls, key):
+        """Find a user by their API key - only succeed if they currently have API access."""
+        res = cls.query(q='api_key:"' + key + '"')
```
emanuil-tolev: .exact? Not that ES would try to analyse a 32 char continuous string, but still better to use the non-analysed field? (Since we thought that of usernames and then the space-at-the-end thing happened)

portality/settings.py
```diff
@@ -111,12 +114,17 @@
...
-    "associate_editor" : [
+    "associate_editor": [
         "edit_journal",
         "edit_suggestion",
-        "editor_area"
+        "editor_area",
+        "api"
+    ],
+    "user": [
+        #"api"              # added on account creation, commented out here so it can be revoked
```
emanuil-tolev: So that means that a publisher will always have the API role then? Lines 107-108 above

Steven-Eardley:
@emanuil-tolev I've made the changes you asked for, plus included the migration script to add role 'api' to all users, plus generating their first api_key.

one slight anomaly is that admins will be given an 'api' role, but this can never be revoked, since auth returns true for all tasks, so the extra flag is superfluous (but the migration runs on all accounts without checking current role)

emanuil-tolev:
Most excellent

emanuil-tolev:
And I suppose we should run the migration on test. You wanna try?

ssh cloo@178.62.92.200
cd /home/cloo/repl/test/doaj/src/doaj
. ../../bin/activate
DOAJENV=test I activate my spell card!!1! (migration run)

Issue #752, opened 2015-05-29
Discovery API - Documentation


emanuil-tolev:
- [ ] docs for #735 , the public search over journals and articles
- [ ] docs for #751 , the authenticated search for applications

emanuil-tolev:
Biggest question here for me now is - should the lump of text related to search be:

A. below the live demos
B. if it remains in the main description above the live demos, should it be collapsible? Thinking that eventually we'll have CRUD routes in there as well and the search notes don't apply to those.

emanuil-tolev:
Another big note to make is that I've decided to leave the API docs public - including for the private routes. I was about to hack flask-swagger to let us have one public docs page with 2 routes and one private docs page for publishers with all the rest of the routes, but actually this is not usual. I would recommend leaving the docs open and fielding the occasional question "how do I use this route", "how do I get an API key" - I think it's important to expose what the DOAJ *can* do and be aware of what people might want to do with it.

If we keep the docs for the authenticated APIs private we'll never know what people might want to do with it. It may well be that DOAJ decides to create a publisher account for a particularly enthusiastic researcher, OA advocate, or developer from a non-publisher organisation like a repository manager and let them submit data to DOAJ. Additionally there may be considerable interest in the applications that DOAJ fields from the public.

@dommitchell I'd love to hear your thoughts on this (and everybody else of course). It is possible to keep the authenticated API pages private as well, though a bit more difficult. The usual case is that the documentation for an entire API is published, and then you may or may not have a key and may or may not have rights to do certain things (but can still read related docs).

dommitchell:
@emanuil-tolev happy with this approach but let me get this clear: DOAJ would still have to assign the API role to someone for them to get a key to use the API right?

dommitchell:
> DOAJ would still have to assign the API role to someone for them to get a key to use the API right?

API role is assigned automatically

PR #799, opened 2015-07-09
Public search API docs

Nimphal:
For #752

### DIFF
--- a/portality/view/api_v1.py
+++ b/portality/view/api_v1.py
@@ -21,7 +21,7 @@ def _bad_request(message=None, exception=None):
 def api_spec():
     swag = swagger(app)
     swag['info']['title'] = "DOAJ API documentation"
-    swag['info']['description'] = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. In a augue nulla. In vehicula est eget erat iaculis, et lacinia enim faucibus. Aliquam diam dolor, euismod ultrices viverra vitae, congue sit amet massa. Duis vestibulum, quam ut ultrices sagittis, nunc diam cursus odio, sed efficitur eros erat a sapien. Lorem ipsum dolor sit amet, consectetur adipiscing elit. In a augue nulla. In vehicula est eget erat iaculis, et lacinia enim faucibus. Aliquam diam dolor, euismod ultrices viverra vitae, congue sit amet massa. Duis vestibulum, quam ut ultrices sagittis, nunc diam cursus odio, sed efficitur eros erat a sapien. I am fairly sure there is no limit to this, but it is limited to the type of stuff we can put in here because it's all one big string, so no paragrahing or the like, I think. "
+    swag['info']['description'] = "This page documents the first version of the DOAJ API."
 
     return make_response((jsonify(swag), 200, {'Access-Control-Allow-Origin': '*'}))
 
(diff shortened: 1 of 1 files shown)


portality/view/api_v1.py
```diff
@@ -40,61 +40,322 @@ def list_operations():
...
+        in: path
         required: true
-        type: "string"
+        type: string
       -
-        name: "search_query"
-        description: "this is definitely a very useful place to put some description of each parameter"
-        in: "path"
+        name: page
+        description: Which page of the results you wish to see. This field is optional.
```
emanuil-tolev: Is it necessary to say this field is optional? There is a required:false below it. If it is necessary (e.g. for visibility) then the pageSize one below, starting at line 64, does *not* have the sentence, but also has required:false declared.

portality/view/api_v1.py
```diff
@@ -40,61 +40,322 @@ def list_operations():
...
+        in: path
+        required: false
+        type: integer
+      -
+        name: pageSize
+        description: How many results per page you wish to see, the default is 100.
```
emanuil-tolev: Can we get the default to be a constant defined in this module? So the docs and the actual code use the same value?

portality/view/api_v1.py
```diff
@@ -40,61 +40,322 @@ def list_operations():
...
+                    format: dateTime
+            query:
+              type: string
+            total:
+              type: integer
+            page:
+              type: integer
+      400:
+        description: Bad Request
+    """
```
emanuil-tolev: A bit long-ish but good I think. I'm sure we can move off the docstrings to someplace else if really necessary.

portality/view/api_v1.py
```diff
@@ -40,61 +40,322 @@ def list_operations():
...
+    # get the values for the 2 other bits of search info: the page number and the page size
+    page = request.values.get("page", 1)
+    psize = request.values.get("pageSize", 10)
```
emanuil-tolev: hm, @Nimphal pageSize seems to be 10? So definitely go for definiing DEFAULT_PAGE_SIZE in the api common.py file or at least top of this module.

portality/view/api_v1.py
```diff
@@ -40,61 +40,322 @@ def list_operations():
...
+        in: path
         required: true
-        type: "string"
+        type: string
       -
-        name: "page"
-        description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit. In a augue nulla. In vehicula est eget erat iaculis, et lacinia enim faucibus. Aliquam diam dolor, euismod ultrices viverra vitae, congue sit amet massa. Duis vestibulum, quam ut ultrices sagittis, nunc diam cursus odio, sed efficitur eros erat a sapien."
-        in: "body"
+        name: page
+        description: Which page of the results you wish to see. This field is optional.
```
emanuil-tolev: same comments as for the journal route at same place

portality/view/api_v1.py
```diff
@@ -40,61 +40,322 @@ def list_operations():
...
-        type: "int"
+        type: integer
       -
-        name: "pageSize"
-        description: "Wooo, we can write small essays in here as well. Again limited as it is a string, but pretty neat"
-        in: "body"
-        required: false,
-        type: "int"
+        name: pageSize
+        description: How many results per page you wish to see, the default is 100.
```
emanuil-tolev: same comments as for the journal route at same place, maybe have a DEFAULT_ARTICLE_PAGE_SIZE and DEFAULT_JOURNAL_PAGE_SIZE if they're different.

portality/view/api_v1.py
Nimphal: It's not necessary, it was just description filler, can remove if needed.

portality/view/api_v1.py
```diff
@@ -40,61 +40,322 @@ def list_operations():
...
+              type: integer
+            page:
+              type: integer
       400:
         description: "Bad Request"
     """
+
     # get the values for the 2 other bits of search info: the page number and the page size
     page = request.values.get("page", 1)
     psize = request.values.get("pageSize", 10)
```
emanuil-tolev: and yeah, page size is 10 for articles too, in the code

emanuil-tolev:
OK to recap for others looking at this you've split the search endpoint into 2 for searching articles and journals, and removed the dispatcher endpoint, so that we can write separate documentation for each, I think owing to some documentation framework limitations. This is fine by me, if that's the right abstraction then ok.

portality/view/api_v1.py
Nimphal: Tested this, doesn't work with the current framework.

emanuil-tolev:
OK, yeah, the swagger extension evaluates the docstrings before Python gets to expanding them (?!) so putting .format on them doesn't work :/. Ignore those comments @Nimphal

portality/view/api_v1.py
emanuil-tolev: Yeah, if the UI makes it clear in some other way that this field is optional / not required then no need to have a string that says the same.

PR #805, opened 2015-07-12
[hotfix] make continuations work with ToC pages

### LINKED ISSUE #696: Continuations not working on live

emanuil-tolev:
# HOTFIX, do not merge here!

Fix #696 . Do not rely on information in the journal, rely on the bibjson. Continuations code currently replaces the bibjson when it finds a continuation, not the journal object (bad conceptually - that's what led to this long-standing bug - but the only thing that can be done if continuations are not "proper" journal objects, which will happen later this year).

### DIFF
--- a/portality/models/journal.py
+++ b/portality/models/journal.py
@@ -1085,6 +1085,14 @@ class JournalBibJSON(GenericBibJSON):
     @publication_time.setter
     def publication_time(self, weeks): self.bibjson["publication_time"] = weeks
 
+    # to help with ToC - we prefer to refer to a journal by E-ISSN, or
+    # if not, then P-ISSN
+    def get_preferred_issn(self):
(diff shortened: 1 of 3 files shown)


Steven-Eardley:
:+1: tested and works fine with TOC and openurl

PR #806, opened 2015-07-13
Feature/751 application discovery

richard-jones:
Contains the following features:

* sort ability on all api endpoints
* jsonp responses to all requests (including errors)
* authenticated application search

### DIFF
--- a/portality/api/v1/common.py
+++ b/portality/api/v1/common.py
@@ -1,5 +1,6 @@
 import json
 from portality.core import app
+from flask import request
 
 class Api(object):
     pass
@@ -9,4 +10,10 @@ class ModelJsonEncoder(json.JSONEncoder):
         return o.data
 
 def jsonify_models(models):
-    return app.response_class(json.dumps(models, cls=ModelJsonEncoder), 200, {'Access-Control-Allow-Origin': '*'}, mimetype='application/json')
+    callback = request.args.get('callback', False)
+    data = json.dumps(models, cls=ModelJsonEncoder)
+    if callback:
+        content = str(callback) + '(' + str(data) + ')'
+        return app.response_class(content, 200, {'Access-Control-Allow-Origin': '*'}, mimetype='application/javascript')
+    else:
+        return app.response_class(data, 200, {'Access-Control-Allow-Origin': '*'}, mimetype='application/json')
--- a/portality/api/v1/common.py
+++ b/portality/api/v1/common.py
@@ -1,4 +1,4 @@
-import json
+import json, uuid
 from portality.core import app
 from flask import request
 
@@ -22,6 +22,7 @@ def respond(data, status):
         return app.response_class(data, status, {'Access-Control-Allow-Origin': '*'}, mimetype='application/json')
 
 def bad_request(message=None, exception=None):
+    # Note - no magic number here, as it will already be in the exception message
     if exception is not None:
         message = exception.message
     app.logger.info("Sending 400 Bad Request from client: {x}".format(x=message))
@@ -29,11 +30,13 @@ def bad_request(message=None, exception=None):
     return respond(data, 400)
 
 def not_found(message=None):
-    app.logger.info("Sending 404 Not Found from client: {x}".format(x=message))
-    data = json.dumps({"status" : "not_found", "error" : message})
+    magic = uuid.uuid1()
+    app.logger.info("Sending 404 Not Found from client: {x} (ref: {y})".format(x=message, y=magic))
+    data = json.dumps({"status" : "not_found", "error" : message + " (ref: {y})".format(y=magic)})
     return respond(data, 404)
 
 def forbidden(message=None):
-    app.logger.info("Sending 401 Forbidden from client: {x}".format(x=message))
-    data = json.dumps({"status" : "forbidden", "error" : message})
+    magic = uuid.uuid1()
+    app.logger.info("Sending 401 Forbidden from client: {x} (ref: {y})".format(x=message, y=magic))
+    data = json.dumps({"status" : "forbidden", "error" : message + " (ref: {y})".format(y=magic)})
     return respond(data, 401)
--- a/portality/api/v1/common.py
+++ b/portality/api/v1/common.py
@@ -10,10 +10,30 @@ class ModelJsonEncoder(json.JSONEncoder):
         return o.data
 
 def jsonify_models(models):
-    callback = request.args.get('callback', False)
     data = json.dumps(models, cls=ModelJsonEncoder)
+    return respond(data, 200)
+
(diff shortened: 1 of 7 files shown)


portality/api/v1/discovery.py
```diff
@@ -65,15 +65,14 @@ def allowed(query, wildcards=False, fuzzy=False):
 
     return True
 
-
 class DiscoveryApi(Api):
 
     @classmethod
-    def _make_query(cls, q, page, page_size, subs):
+    def _sanitise(cls, q, page, page_size, sort, search_subs, sort_subs):
         if not allowed(q):
```
emanuil-tolev: can you make q from this point on a deepcopy of the q you get passed in? The DAO was doing this same thing and it ended up modifying the q object that's passed in, resulting in very subtle bugs. Well, not being able to delete articles is not so subtle I guess.

One way of doing that: I've modified the DAO to name the parameter passed in ```theq``` (star trek anyone?) and then the first line of the method is to make q = deepcopy(theq).

portality/view/api_v1.py
```diff
@@ -40,8 +34,74 @@ def list_operations():
...
+        in: "body"
+        required: false,
+        type: "int"
+      -
+        name: "sort"
+        in: "body"
+        required: false
+        type: "string"
+      -
+        name: "api_key"
```
emanuil-tolev: Ah, I think this is usually used as HTTP basic auth instead of passing the key into the params, so you don't get your API key plastered over your requests, but just get it put in a HTTP header. My fault for not spotting on @Steven-Eardley's PR, I'll discuss in that issue. It's not an issue with *this* PR.

emanuil-tolev: On the other hand it's probably going to be easier in the real world to let people tack it onto the URL than force them to subscribe to a particular best practice.

portality/api/v1/discovery.py
```diff
@@ -140,16 +184,35 @@ def search_journals(cls, q, page, page_size):
...
+        search_subs = app.config.get("DISCOVERY_APPLICATION_SEARCH_SUBS", {})
+        sort_subs = app.config.get("DISCOVERY_APPLICATION_SORT_SUBS", {})
+        query, page, page_size = cls._make_application_query(account, q, page, page_size, sort, search_subs, sort_subs)
+
+        # execute the query against the articles
+        res = models.Suggestion.query(q=query.query(), consistent_order=False)
+
+        # check to see if there was a search error
+        if res.get("error") is not None:
+            app.logger.error("Error executing discovery query search: {x}".format(x=res.get("error")))
```
emanuil-tolev: Wise to include a magic id for debugging support reqs and api feedback? Like this: https://github.com/DOAJ/doaj/blob/develop/portality/view/account.py#L178 ?

portality/api/v1/common.py
```diff
@@ -9,4 +10,30 @@ def default(self, o):
...
+    if callback:
+        content = str(callback) + '(' + str(data) + ')'
+        return app.response_class(content, status, {'Access-Control-Allow-Origin': '*'}, mimetype='application/javascript')
+    else:
+        return app.response_class(data, status, {'Access-Control-Allow-Origin': '*'}, mimetype='application/json')
+
+def bad_request(message=None, exception=None):
+    if exception is not None:
+        message = exception.message
+    app.logger.info("Sending 400 Bad Request from client: {x}".format(x=message))
```
emanuil-tolev: Consider a magic like https://github.com/DOAJ/doaj/blob/develop/portality/view/account.py#L178 here. Doesn't cost anything to make a UUID. Best to give the magic to the client as well though! So when they say "I got error with magic X", we find it in a flash.

portality/api/v1/common.py
```diff
@@ -9,4 +10,30 @@ def default(self, o):
...
+    data = json.dumps({"status" : "error", "error" : message})
+    return respond(data, 400)
+
+def not_found(message=None):
+    app.logger.info("Sending 404 Not Found from client: {x}".format(x=message))
```
emanuil-tolev: same as above

portality/api/v1/common.py
```diff
@@ -9,4 +10,30 @@ def default(self, o):
...
+    data = json.dumps({"status" : "not_found", "error" : message})
+    return respond(data, 404)
+
+def forbidden(message=None):
+    app.logger.info("Sending 401 Forbidden from client: {x}".format(x=message))
```
emanuil-tolev: same as above

emanuil-tolev:
Most excellent! The only substantial feedback I had was around logging. I'm happy to add the magic to the logging if you've not got time after all this code up there @richard-jones .

portality/api/v1/discovery.py
richard-jones: I don't think that's necessary here - "q" in these functions is the query string, not a query object, so it doesn't suffer from the same by-reference issues.  Additionally, on the subsequent lines, there is a call to 

    q = query_substitute(q, search_subs)

and this returns a totally new value for the query string, too, rather than modifying the existing one in place.

Further on still, the query itself is assembled as an object:

    query = SearchQuery(q, fro, page_size, sortby, sortdir)

and this only yields an actual query object as a dict when query.query() is called.  Even if the output of that function does get modified later on by the DAO, this won't affect the upstream object, which can generate the original query at any time by calling query.query() again (not that we need to).

In summary, yes, we should be cautious of modifying inputs in the way that the dao does (and it shouldn't do that, really), but this code insulates itself from those kinds of issues, and is well-behaved within its own scope.

portality/view/api_v1.py
richard-jones: Ah, sorry, I wasn't sure what the overall plan was, but putting ?api_key=blah certainly works, so I thought I'd better put it in the docs.

Need me to change anything, or will @Nimphal sort that out when the docs are finalised?

portality/api/v1/discovery.py
richard-jones: Ok, sounds like a good idea, will add

portality/api/v1/common.py
richard-jones: Yup, will come with the next commit

richard-jones:
Added the magic numbering

portality/api/v1/discovery.py
emanuil-tolev: Ah yes it is a string, sorry. :+1:

portality/view/api_v1.py
emanuil-tolev: No, nothing to do, just an idea for me. Besides it's right as it is. We could decide to change it later but not to do with your feature.

emanuil-tolev:
Marvelous!

PR #807, opened 2015-07-13
Feature/782 bulk delete

richard-jones:
Provides bulk delete buttons in the administrative site search page for both journals and articles.


emanuil-tolev:
Amazing, works great. Merging into develop so we can deploy with phase 1 of API which is coming very shortly.

PR #816, opened 2015-07-17
prepare final version of discovery API docs sans applications

emanuil-tolev:
Overall prep of discovery API docs for final release. Only missing thing is applications schema and sort options. Journals and articles routes should be complete and you should not need to refer to github in order to perform a search - all info should be there.

Would be great if you have the time to have a play with the 2 public endpoints @richard-jones @Steven-Eardley @Nimphal . Paging, sorting, searching, shorthand fields were my stops on the happy api bus.

Also overall how does it look / feel? There is now a screenful of text before it gets to the examples - however you can link to a specific live route demo by just opening it (the URL will change and if you share that, it'll scroll past the text). My biggest question is should the text go somehow below the live demos, or is it OK as it is.

When you check out this branch go to http://localhost:5004/api/v1/docs and make sure you have a little bit of data so you get something back (11+ journals and 11+ articles will do).

### DIFF
--- a/portality/app.py
+++ b/portality/app.py
@@ -214,9 +214,9 @@ if 'api' in app.config['FEATURES']:
             {
                 'api_versions': [
                     {
-                        'version': '0.0.1',
-                        'base_url': url_for('api_v1.list_operations', _external=True),
-                        'note': 'First version of the DOAJ API (IN DEVELOPMENT DO NOT USE!)',
+                        'version': '1.0.0',
+                        'base_url': url_for('api_v1.api_spec', _external=True),
+                        'note': 'First version of the DOAJ API',
                         'docs_url': url_for('api_v1.docs', _external=True)
                     }
                 ]
--- a/portality/static/doaj/css/doaj_api.css
+++ b/portality/static/doaj/css/doaj_api.css
@@ -1,3 +1,7 @@
+em {
+    font-style: italic;
+}
+
 body.swagger-section {
     padding-top: 1em;
     font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
@@ -21,6 +25,23 @@ footer {
     font-weight: normal;
 }
 
-.swagger-section .swagger-ui-wrap .markdown ol {
+
+.swagger-section .swagger-ui-wrap ol {
     list-style-type: decimal;
+}
+
+#extra-docs > ul {
+    list-style: disc inside;
+}
+
+#extra-docs > ul > li {
+    margin-bottom: 0.3em;
+}
+
+ul.options {
+    font-size: 125%;
+}
+
+#api_info {
+    line-height: 150%;
 }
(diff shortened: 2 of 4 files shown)


richard-jones:
I tried to review this but I get a blank page at /api/v1/docs

I poked about underneath, and it seems that /api/v1/spec is throwing a 500

This is the case on my local machine and on the test server.

richard-jones:
Wait, the 500 is my fault (locally) - it actually gives me 404s

richard-jones:
It's because the swagger-ui directory is empty

Steven-Eardley:
I was getting the same thing during the meeting this morning.

emanuil-tolev:
@richard-jones @Steven-Eardley git submodule init && git submodule update

richard-jones:
ah, right.  You probably need to do that on the test server too, then

richard-jones:
Ok, some general thoughts in no particular order:

* The very first thing that you read on the docs page is some detailed stuff about json dot notation, which doesn't seem like the best way in.  Perhaps move these details to the bottom, and hyperlink them out of the collapsible swagger sections instead?

* I'm not sure that first section should be called "Friendly field names", as this might be confused with the substitutions that are documented elsewhere

* Need to add regex to list of disabled features

* Update the links to the latest version of the lucene query string docs

* In the swagger section, you've included the documentation on the substitute field names, but this sort of feels like it needs to be read alongside the "Friendly field names" section.  I'd move that out, and just reference the text wherever it actually ends up.  I like that there are examples here, though, we should keep them

* Same comments on the "sort" parameter

Not sure what you want to do next - want to play around more, or just merge it and work on refinements later?

emanuil-tolev:
Nah, I'd rather do the refinements - asap in morning. Thanks a lot for the comments. The test server specifically will be fixed when we merge this as it fixes a YAML syntax error (that's what Nevelina was referring to today when she said "refresh page locally if you write docs").

Good thing is it sounds like there aren't any big foundation problems, and I agree with all the comments - luckily they're quite easy to change.

emanuil-tolev:
@richard-jones I've updated the PR and acted on all the points above, hopefully it's quite a bit nicer now!

I did leave one thing as is:

> Same comments on the "sort" parameter

I see the point of consistency with the query parameter. However, this section (for both articles and journals) is really small - the text currently in the sort parameter description box is all there is to it. There is also nothing to explain that is shared between articles and journals, unlike the query string explanation of shorthand fields. I feel that taking the sort out to the bottom section of extra docs will unnecessarily disconnect it from the endpoint documentation, considering it's just a few lines and there's no common ground. I'm happy to put it under the short field names section if there's a reason it should be taken out of the swagger bit that I've missed though.

emanuil-tolev:
Also rebased against develop so DOI queries will work, now includes all latest fixes.

Nimphal:
@emanuil-tolev @richard-jones I've pushed a commit with the documentation for the application search API.

emanuil-tolev:
I'm going to merge this so we can all see working docs on the test server and comment. I'm happy to make any further amends to the docs that come out of that. Thanks @Nimphal for adding the final piece of docs.

PR #818, opened 2015-07-20
Feature/735 escape slash

richard-jones:
Escapes all / characters in incoming search strings if they are not already escaped

### DIFF
--- a/doajtest/unit/test_api_discovery.py
+++ b/doajtest/unit/test_api_discovery.py
@@ -116,6 +116,13 @@ class TestArticleMatch(DoajTestCase):
         with self.assertRaises(DiscoveryException):
             res = DiscoveryApi.search_journals("Test", 1, 10, "some.missing.field:asc")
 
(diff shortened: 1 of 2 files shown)


emanuil-tolev:
So, happy to merge as-is, but wondering if it's easy to write a test as well? Like save an article with a doi, then search for it and, well, get the result back. But can leave out if it's more than very straightforward.

richard-jones:
Yes, wise, especially as in the process of looking at a test I found a bug

richard-jones:
Ok, tests in - fun with layers of escape characters.

emanuil-tolev:
Wow, that's pretty subtle with the regex overlapping matches.

Does the ```http:``` test you've added also mean that people can't put in ```:``` colon into searches then? I'll add that to the docs if so.

richard-jones:
Yes, that's right, the ":" has special meaning so we can't trivially escape it automatically, so users will just have to be aware of what they're doing.  If they read the lucene docs it should be obvious what would happen if you put in a URL, but it could well be worth giving an example, since URL search might be something people will want to do.

PR #820, opened 2015-07-23
add CRUD API foundation together with journal route notes

emanuil-tolev:
This contains the CRUD API foundation ref #754 

Some notes have been added as well for @Nimphal and @bg280 to start on.

The overall structure is one of composition: we have a package for the whole CRUD API, and inside that package we have an ApplicationsCrudApi, an ArticleCrudApi and a JournalCrudApi. The ```CrudApi``` which is eventually used by the view and exposed at the top-level (portality.api.v1) simply inherits from those 3 APIs, gaining all their methods in the process.

The whole thing enables:

1. adding new types to CRUD very easily
2. keeping code separate for applications, articles and journals - if we had one CrudApi class it'd be hard to edit and git would also complain a lot when multiple people worked on the one file. And it'd be like 1500 lines long by the end.
3. encapsulation - CrudApi is one object at the end of the day, the view routes are none the wiser to its implementation details. So long as you keep the classmethod names you can write the underlying code in assembly if you want.

@richard-jones @Steven-Eardley FYI and for comments of course!

### DIFF
--- a/doajtest/unit/test_api_account.py
+++ b/doajtest/unit/test_api_account.py
@@ -5,9 +5,23 @@ from doajtest.helpers import DoajTestCase
 from portality import models
 from portality.core import app
 from portality.app import load_account_for_login_manager
-from portality.decorators import api_key_required
+from portality.decorators import api_key_required, api_key_optional
 
 class TestClient(DoajTestCase):
+    @classmethod
+    def setUpClass(cls):
+        @app.route('/hello')
+        @api_key_required
+        def hello_world():
+            return Response("hello, world!")
+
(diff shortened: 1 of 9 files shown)


emanuil-tolev:
Added a test for the api_key_optional decorator as well. Had a spot of trouble with the test Flask app (can't set up any more routes after you run one test, because once it serves one request in one test it refuses to register new routes and throws a fit when you try). So now set up app before *all* tests and then ... test.

Steven-Eardley:
This looks good to me, although I haven't tested the decorator myself, I'm satisfied if it passes your tests!

PR #821, opened 2015-07-24
modify CRUD api internals to now expose 3 CRUD API-s with a common parent

emanuil-tolev:
No tests to write/change yet. This changes the CRUD tech design so it fits with what we discussed in the meeting - now you can import an ApplicationsCrudApi, etc. from portality.api.v1 .


Steven-Eardley:
Makes sense to me

PR #822, opened 2015-07-30
add liberal use of escapeHtml to custom render functions to prevent e…

richard-jones:
…mbedded html killing the interface

DON'T MERGE THIS IN GITHUB

(I think - it's a hotfix, anyway)


emanuil-tolev:
yep :+1:

PR #825, opened 2015-08-03
Some changes for API-specific error handlers.

Steven-Eardley:
* Improved error handlers for the API project, along with tests.
* updated api_key decorator to use new exception method.
* added route to api_v1 which returns json 404 rather than HTML
* Added new setting VALID_FEATURES which lists all available features
* test boostrap will activate all features in VALID_FEATURES, so tests don't fail


richard-jones:
@emanuil-tolev - we had a look at this (Steve and I) and decided that probably you should have a check over it to ensure it fits with your API worldview, so I haven't merged it.

emanuil-tolev:
Yep, seems reasonable. Thanks a lot for doing all that @Steven-Eardley !

PR #841, opened 2015-08-13
New data object and Application Create API

richard-jones:
This provides the following:
* A new data object based on the previous iteration, but with more features
* Revised model objects and tests for journal/vanilla data object
* The Create portion of the Application CRUD API
* Some additions to the api common library which may be shared over all object types
* a README which outlines the API interactions planned for the Applications

It isn't the final Application CRUD but because of the new data object, and the overall pattern demonstrated around Application Create I think it is useful to merge this now.  It also has common code that other CRUD endpoints will want to make use of.

### DIFF
--- a/portality/api/v1/data_objects/application.py
+++ b/portality/api/v1/data_objects/application.py
@@ -1,0 +1,323 @@
+from portality.lib import dataobj
+from portality import models
+
+class IncomingApplication(dataobj.DataObj):
+    def __init__(self, raw=None):
+        struct = {
+            "fields": {
+                "id": {"coerce": "unicode"},                # Note that we'll leave these in for ease of use by the
+                "created_date": {"coerce": "utcdatetime"},  # caller, but we'll need to ignore them on the conversion
+                "last_updated": {"coerce": "utcdatetime"}   # to the real object
+            },
+            "objects": ["bibjson", "suggestion", "admin"],
+            "required" : ["bibjson", "suggestion", "admin"],
+
+            "structs": {
+                "bibjson": {
+                    "fields": {
+                        "title": {"coerce": "unicode"},
+                        "alternative_title": {"coerce": "unicode"},
+                        "country": {"coerce": "country_code"},       # FIXME: need to make sure this is the correct form
+                        "publisher": {"coerce": "unicode"},
+                        "provider": {"coerce": "unicode"},
+                        "institution": {"coerce": "unicode"},
+                        "apc_url": {"coerce": "url"},
+                        "submission_charges_url": {"coerce": "url"},
+                        "allows_fulltext_indexing": {"coerce": "bool"},
+                        "publication_time": {"coerce": "integer"},
+                    },
+                    "lists": {
+                        "identifier": {"contains": "object"},
+                        "keywords": {"coerce": "unicode", "contains": "field"},
+                        "language": {"coerce": "isolang_2letter", "contains": "field"}, # FIXME: need to make sure this is the correct form
+                        "link": {"contains": "object"},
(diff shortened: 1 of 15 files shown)


portality/api/v1/data_objects/application.py
```diff
@@ -0,0 +1,323 @@
...
+                            "required" : ["name", "email"]
+                        }
+                    }
+                }
+            }
+        }
+
+        super(IncomingApplication, self).__init__(raw, struct=struct, construct_silent_prune=True, expose_data=True)
+
+    def custom_validate(self):
```
emanuil-tolev: We would ideally refactor form, api and CSV reapplication validation (the common parts at least) into some form of validation library that can be easily used by the api, the forms and the custom ingestion code. Not now, but we should do fairly soon in a project so we don't end up in a bit of a pickle if validation requirements change. Or we could leave the refactor until they do change of course.

portality/lib/dataobj.py
```diff
@@ -172,184 +178,731 @@ class DataSchemaException(Exception):
...
+            path = key
+            wrapper = DataObj
+
+        # try to set the property on othe internal object
+        if path is not None:
+            wasset = self._set_internal_property(path, value, wrapper)
+            if wasset:
+                return
+
+        # fall back to the default approach of allowing any attribute to be set on the object
```
emanuil-tolev: So if you try to set something that is not a dynamically defined property ("shortcut" to a path) or part of the struct, we would just allow the setting to take place?

What would be the use case for this? When I wrote the previous version the reason it didn't allow you to set a property that wasn't defined in the struct was because with these objects you don't get autocomplete, which means typos are more likely, so I wanted to immediately stop you from assigning to something that isn't going to end up in self.data .

portality/lib/dataobj.py
```diff
@@ -172,184 +178,731 @@ class DataSchemaException(Exception):
...
+                elif isinstance(val, list):
+                    return [wrapper(v, expose_data=self._expose_data) for v in val]
+
+            # otherwise, return the raw value if it is not None, or raise an AttributeError
+            if val is None:
+                raise AttributeError('{name} is not set'.format(name=path))
+
+            return val
+
+        # if the struct contains a reference to the path, always return something, even if it is None - don't raise an AttributeError
```
emanuil-tolev: and this can be set in the struct as this get__default definition I've seen, is that right?

emanuil-tolev:
Very nice. I'm a bit concerned about the size of the dataobj code and we'd probably need to comment the library a bit more as we use it, but as you've been using it successfully on other projects, :+1: . I just had 1 question about setting any attr: https://github.com/DOAJ/doaj/pull/841#discussion_r37070606 and we can merge after that's answered.

portality/api/v1/data_objects/application.py
richard-jones: Yes, quite right.  I think we're probably close to another major bit of refactoring, but I'm not quite sure what it is yet.  Possibly need to spend a bit more time with the Gang of Four again ...

portality/lib/dataobj.py
richard-jones: This is how python objects actually work, which is the reason I've done this.  Without this, it is not possible for a sub-class to define its own properties.

portality/lib/dataobj.py
richard-jones: Yup, exactly

richard-jones:
The code is large, agreed, but it does a large number of things.  Not all the code is relevant for the API, but it will be when we get to the model refactoring.  In octopus I have a fairly decent suite of tests which try all the various angles, so it is well tested.  This is one of those problems where, basically, the solution /is/ difficult :)

PR #843, opened 2015-08-16
Complete treatment of application object validation

richard-jones:
Thought it would be useful to make another PR, as I've now done a full treatment of the validation requirements for the application model, and this will be useful in the implementation of the journal crud.

Note the addition of custom coerce functions in the constructor, which use a newly created string_canonicalise dataobj coerce routine, giving us parity with the behaviour of the submission form.

### DIFF
--- a/doajtest/unit/test_api_crud_application.py
+++ b/doajtest/unit/test_api_crud_application.py
@@ -23,12 +23,11 @@ class TestCrudApplication(DoajTestCase):
         data = ApplicationFixtureFactory.incoming_application()
         ia = IncomingApplication(data)
 
-        # FIXME: this bit doesn't yet work - dataobj refactor required
         # make another one that's broken
-        #data = ApplicationFixtureFactory.incoming_application()
-        #del data["bibjson"]["title"]
-        #with self.assertRaises(DataStructureException):
-        #    ia = IncomingApplication(data)
+        data = ApplicationFixtureFactory.incoming_application()
+        del data["bibjson"]["title"]
+        with self.assertRaises(DataStructureException):
+            ia = IncomingApplication(data)
 
         # now progressively remove the conditionally required/advanced validation stuff
         #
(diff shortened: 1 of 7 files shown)


doajtest/unit/test_api_crud_application.py
```diff
@@ -23,12 +23,11 @@ def test_01_incoming_application_do(self):
...
-        # FIXME: this bit doesn't yet work - dataobj refactor required
         # make another one that's broken
-        #data = ApplicationFixtureFactory.incoming_application()
-        #del data["bibjson"]["title"]
-        #with self.assertRaises(DataStructureException):
-        #    ia = IncomingApplication(data)
+        data = ApplicationFixtureFactory.incoming_application()
+        del data["bibjson"]["title"]
+        with self.assertRaises(DataStructureException):
+            ia = IncomingApplication(data)
```
emanuil-tolev: :+1:

portality/api/v1/data_objects/application.py
```diff
@@ -203,28 +233,17 @@ def __init__(self, raw=None):
...
-                            "required" : ["name", "email"]
-                        }
-                    }
                 }
             }
         }
 
-        super(IncomingApplication, self).__init__(raw, struct=struct, construct_silent_prune=True, expose_data=True)
+        mycoerce = deepcopy(self.DEFAULT_COERCE)
+        mycoerce["persistent_identifier_scheme"] = dataobj.string_canonicalise(["None", "DOI", "Handles", "ARK"], allow_fail=True)
```
emanuil-tolev: Would it be a good idea to get these from the form choices and string them together in a list dynamically? Rather than specifying explicitly here.

emanuil-tolev:
Just 1 question about possibly pulling in the form choices for string canonicalisation from the formcontext modules.

emanuil-tolev:
But not blocking, so merged. Feel free to make changes responding to the comment in the next PR

portality/api/v1/data_objects/application.py
richard-jones: yes, definitely

PR #845, opened 2015-08-17
[hotfix] expose API and make auth information for API easier to find

emanuil-tolev:
# Hotfix, do not merge here

Make the API link public, fix API base URL to be the right one and put in a few sentences about auth which are a bit more specific and lead to the right area now.

### DIFF
--- a/doajtest/bootstrap.py
+++ b/doajtest/bootstrap.py
@@ -7,6 +7,9 @@ logging.getLogger("requests").setLevel(logging.WARNING)
 def prepare_for_test():
     core.app.config['ELASTIC_SEARCH_DB'] = core.app.config['ELASTIC_SEARCH_TEST_DB']
 
+    # Ensure all features are enabled so we don't fail feature-specific tests
+    core.app.config['FEATURES'] = core.app.config['VALID_FEATURES']
+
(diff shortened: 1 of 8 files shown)


richard-jones:
Apparently the build failed, although I can't see from a quick peek at the code what would have caused that

Steven-Eardley:
I think this is failing the build because the API feature is disabled, so it can't reach those routes. Because this is based on ```production``` and not ```develop```, it doesn't have the adjustments I made to the tests to activate all features (in the commit here https://github.com/DOAJ/doaj/commit/26376f88b9ba2c50fa6c184a77abb16af0a1c5cb) to pass the API tests.

Steven-Eardley:
:+1: to your changes; But be aware, cherrypicking this commit means we'll also be releasing my changes to the API error handler code. This has been reviewed before, and should improve the error messages for the Discovery API that's been released, so I'll go ahead and release the hotfix now.

emanuil-tolev:
> But be aware, cherrypicking this commit means we'll also be releasing my changes to the API error handler code

Yep, that's what I thought but it looked like the errors would continue to be raised, so why not.

Thanks for releasing!

PR #851, opened 2015-08-24
Application CRUD API

richard-jones:
Ok, here's full coverage of the application CRUD API.  There are a number of key changes here:

* addition of retrieve, update and delete routes
* Addition of OutgoingApplication model
* Refactoring of the structs into 3 parts: the base struct for incoming applications, the required fields for incoming applications, and the extended struct for outgoing applications.  Separating the required fields into their own struct allows us to apply it to incoming data, but not to outgoing data (which, while there is still legacy data, we don't want to hold to the same restrictions).
* the ability to merge two dictionaries, where one adheres to a struct and the other does not, such that the latter augments the former only in areas not covered by the struct.  This allows us to merge an incoming application and an existing Suggestion to retain all the old suggestions administrative info.
* various minor tweaks to data object code, including recursive merge of structs (omitted by mistake from previous versions) and better handling of non-structured data objects (which we only really use in testing).
* a load of tests


emanuil-tolev:
:+1: - it had conflicts that I fixed so just waiting for build to finish again and can be merged.

Issue #854, opened 2015-09-09
Bug: lost the colour on the facet removal crosses


dommitchell:
I am certain these haven't always been white! They're near invisible now.

![crosses](https://cloud.githubusercontent.com/assets/5955396/9760633/18aeb2f6-56f7-11e5-9161-3d6027ac56ba.jpg)

emanuil-tolev:
Wow, that's so weird.

I found the culprit, as part of the new social media icon on the page we had to upgrade the FontAwesome library so we could get the icon without, like, recreating it ourselves in the style of the others. Unfortunately that seems to have had this interesting effect. The cross is supposed to be a certain shade of blue.

emanuil-tolev:
Although I'm struggling to understand what that has to do with this icon, since this is an icon from Bootstrap, a core style library we use. It doesn't come from FontAwesome.

emanuil-tolev:
Hahah, I got it, the old version of FontAwesome would actually override Bootstrap icons. The Bootstrap ones don't take custom colours at all, but we were unwittingly using FontAwesome ones, which look the same, except you can colour them. How's that for a subtle bug :).

I've got a hotfix up for review. It's simple, but I want it reviewed since there are many ways to skin this particular cat and I want to make the change maintainable so we have an easy way of making future changes like this, rather than heap on style hacks. I think I've done it a good way.

Steven-Eardley:
@emanuil-tolev Seriously good catch! Looks like I should've eyeballed more of the site when I upgraded the library. Thanks for taking care of this.

Steven-Eardley:
Fix on live now, @dommitchell

dommitchell:
Great! Closing.

PR #855, opened 2015-09-09
[hotfix] Fix invisible close button on filters

emanuil-tolev:
# HOTFIX, DO NOT MERGE HERE

For #854 

There are many ways I could've done this including just removing .icon-white and .icon-remove with Javascript on page load, and then replacing with .fa .fa-remove . But that feels hackier, I think the HTML should contain the intent (otherwise as a dev I will one day wonder why the hell are my bootstrap icons coloured when that's impossible and I don't seem to be using FontAwesome ones - the override is too easy to miss).

I wanted to do it the right way which to me seems to be to use the callback functionality and just render custom filters, identical in every way except the remove icon.


Steven-Eardley:
:+1: I like this solution - you've fixed the problem with CSS without bodging the CSS.

PR #859, opened 2015-09-11
Feature/api crud articles

Steven-Eardley:


portality/api/v1/crud/applications.py
```diff
@@ -109,7 +110,7 @@ def update(cls, id, data, account):
 
         # finally save the new application, and return to the caller
         new_ap.save()
-        return ap
+        return new_ap
```
emanuil-tolev: Nice catch in applications. Does a test need to be added or change as well so that we make sure we return the new application to the user rather than the old object? I'm not saying you should add it in this PR, just if we should make a note to.

portality/api/v1/crud/articles.py
```diff
@@ -1,5 +1,111 @@
...
+        am = ia.to_article_model()
+
+        # Check we are allowed to create an article for this journal
+        if not XWalk.is_legitimate_owner(am, account.id):
+            raise Api404Error()                 # not found for this account
+
+        # if the caller set the id, created_date, or last_updated, then we discard the data and apply our
+        # own values (note that last_updated will get overwritten anyway)
+        am.set_id()
+        am.set_created()
```
emanuil-tolev: Do we also need to overwrite the "journal" section of the article bibjson? Except start_page, end_page, volume and number (issue number), the other stuff (like title, license, country) will need to be copied in from the journal I think. Even if we let people specify them, we should ignore them. It's not possible (or shouldn't be) to upload an article without a journal being already present, so we can always perform this overwrite. It is also only happening on create and update, on the way into the system.

portality/view/forms.py
```diff
@@ -904,25 +904,25 @@ def __call__(self, form, field):
...
     eissn = SelectField("Journal ISSN (online version)", [ThisOrThat("pissn")], choices=[]) # choices set at construction
  
-    volume = TextField("Volume Number", [validators.Optional()])
-    number = TextField("Issue Number", [validators.Optional()])
-    start = TextField("Start Page", [validators.Optional()])
-    end = TextField("End Page", [validators.Optional()])
+    volume = StringField("Volume Number", [validators.Optional()])
+    number = StringField("Issue Number", [validators.Optional()])
+    start = StringField("Start Page", [validators.Optional()])
+    end = StringField("End Page", [validators.Optional()])
```
emanuil-tolev: All these form changes I assume make it work the same as before, they're just bc TextField is deprecated? And the DataRequired change .. for a similar reason?

emanuil-tolev:
v-v-v-v-very nice! A few questions which would be nice to know about, and as long as the forms work the same way as before I'm very happy to merge. There's no other blocking stuff, maybe some things we can note to improve before the final release of phase 2.

portality/api/v1/crud/applications.py
Steven-Eardley: Yes, my next PR for the API is going to be another look at Api tests - I'll add that to my list.

portality/view/forms.py
Steven-Eardley: Yes, those are the deprecated fields, and since this form is still in use, I thought I'd make the change. Yes, they behave in the same way, same for the validators. At least, I've been changing them whenever I come across them so far and nobody's noticed!

portality/api/v1/crud/articles.py
Steven-Eardley: Yes, I think you're right there. I'll amend this.

Steven-Eardley:
I also have a question for you - should Updates accept PATCH or PUT (or both)? @richard-jones has used PUT, so I did as well.

emanuil-tolev:
Good point, but let's just go with PUTs for now. They are quite conventional in web apps as an idempotent operation to update stuff. Besides we accept the whole record rather than a partial, which is more put-ty than patch-y.

emanuil-tolev:
We decided not to give people journal info at all actually. I'll merge this to get the other changes and the workflow in, and ideally we'll amend this endpoint so it doesn't expose journal data at all in the article (beyond ISSN).

Issue #849, opened 2015-08-20
Bug when manually assigning roles on account creation vs. roles added automatically


dommitchell:
There's a bug in the system which means that every new user created manually gets the api role. Even if you remove that role and assign another before hitting submit, only the removed api role sticks. You have to go to the new record, remove API and add the role you want.

![api1](https://cloud.githubusercontent.com/assets/5955396/9382999/cf22febe-4745-11e5-950d-4992375dc699.jpg)

![api2](https://cloud.githubusercontent.com/assets/5955396/9383010/d7511134-4745-11e5-9f81-5c375d6882c1.jpg)

![api3](https://cloud.githubusercontent.com/assets/5955396/9383015/db92c22e-4745-11e5-952c-a736ed1db6e7.jpg)

API role should only be added to user accounts that have a publisher role allocated to them, whether those accounts are created automatically upon acceptance of an application *or* manually.

How does this work for the user accounts of reapplications?

emanuil-tolev:
> How does this work for the user accounts of reapplications?

They're "publisher" aren't they?

emanuil-tolev:
#846 deals with "only assign api role automatically to users with role publisher" but you seem to have found some other sort of bug where you specify a certain role or multiple roles, but it gets overwritten when the account should have the "api" role added (currently doing that for every new account).

dommitchell:
> How does this work for the user accounts of reapplications?
> 
> They're "publisher" aren't they?

They are but they have existing accounts. Actually, I just remembered that we will apply the api role to all publisher accounts. All good.

Steven-Eardley:
@dommitchell The fix for this is now live.

In brief, I've fixed the functionality such that the 'api' role is auto-filled as a suggestion on the form, but when it's removed it'll actually be removed from the account! It will also pay attention to which other role(s) you enter in the form now.

That sounds about right to me - I've no idea what I was thinking before!

dommitchell:
And it works like a dream! Closing!

PR #870, opened 2015-09-21
Hotfix/2015 09 19 2000 new acc role bug

Steven-Eardley:
**Hotfix**

for issue #849

### DIFF
--- a/doajtest/unit/test_api_account.py
+++ b/doajtest/unit/test_api_account.py
@@ -31,7 +31,7 @@ class TestClient(DoajTestCase):
 
     def test_01_api_role(self):
         """test the new roles added for the API"""
-        a1 = models.Account.make_account(username="a1_user", name="a1_name", email="a1@example.com", roles=["user"], associated_journal_ids=[])
+        a1 = models.Account.make_account(username="a1_user", name="a1_name", email="a1@example.com", roles=["user", "api"], associated_journal_ids=[])
         a1.save()
         time.sleep(1)
 
@@ -41,18 +41,9 @@ class TestClient(DoajTestCase):
 
         # Check we can retrieve the account by its key
         a1_retrieved = models.Account.pull_by_api_key(a1_key)
+        print a1
+        print a1_retrieved
         assert a1 == a1_retrieved
-
-        # Check all roles gain the api role on creation
-        a2 = models.Account.make_account(username="a2_user", name="a2_name", email="a2@example.com", roles=["admin"], associated_journal_ids=[])
-        a3 = models.Account.make_account(username="a3_user", name="a3_name", email="a3@example.com", roles=["publisher"], associated_journal_ids=[])
-        a4 = models.Account.make_account(username="a4_user", name="a4_name", email="a4@example.com", roles=["editor"], associated_journal_ids=[])
-        a5 = models.Account.make_account(username="a5_user", name="a5_name", email="a5@example.com", roles=["associate_editor"], associated_journal_ids=[])
-        assert a1.has_role("api")
-        assert a2.has_role("api")
-        assert a3.has_role("api")
-        assert a4.has_role("api")
-        assert a5.has_role("api")
 
         # Check that removing the API role means you don't get a key
         a1.remove_role('api')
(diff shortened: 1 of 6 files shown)


emanuil-tolev:
Right, so no matter what you do it just gets an "api" role. Will it get an "api" role now? E.g. if I go create an account, will it still at least suggest that the new account should get the "api" role by pre-filling the role field with 1x "api" entry for me?

Steven-Eardley:
Looks like it's still not quite right. I knew it needed tests!

Steven-Eardley:
Okay, now I've removed the rather obnoxious adding of API role in the model whenever you create a user. Instead, it just suggests it in the UI by pre-filling the register form. I've also had to update the createuser script to let you know the role is usually supposed to be there (but after the fact, like the existing warning when there's no role at all).

The changes to tests are mainly to reflect this change in approach - new users aren't automatically given API access by ```models.Account.make_account``` any more, so it needs to be explicit if that's called.

doajtest/unit/test_api_account.py
```diff
@@ -41,30 +41,21 @@ def test_01_api_role(self):
 
         # Check we can retrieve the account by its key
         a1_retrieved = models.Account.pull_by_api_key(a1_key)
+        print a1
+        print a1_retrieved
```
emanuil-tolev: I suppose you didn't mean to keep these 2 print statements :)

Steven-Eardley: Certainly didn't! Good heavens this has been sloppy from the outset :-/

portality/models/account.py
```diff
@@ -152,12 +149,19 @@ def prep(self):
     @property
     def api_key(self):
         if self.has_role('api'):
-            return self.data.get('api_key', None)
+            # Return the stored api key if it's there, generate one if not (there should be one present if the role is)
```
emanuil-tolev: Hmmm. Generating a key if there isn't one is cool .. but it won't really work until the object is .save()-ed. I don't think we should .save() the object on generate_api_key() either.

Is it really necessary to generate one if there isn't one *on read access*? Just return None. It's the caller's responsibility to then .generate_api_key() and .save() . My concern is that this way, say I'm debugging somebody whose key got deleted by an admin. I go into the console and I type account.api_key .. and I get a key! Except it won't work and it will change every time I refresh the object, since it's not persisted. This will also allow you to do ```self.api_key``` on line 128 instead of ```self.data.get('api_key', None)``` - that's already a code smell which tells you the getter isn't really a getter as you're not able to use it as such.

So it'll change the last line of the new test you added above - after deleting an api key, the key should in fact be None :).

Unless I'm missing some kind of use case here where generating the key on read access it required? I'd really look for a way to call .generate_api_key() explicitly in the view function (or whatever the use case that is using the Account model is) instead.

portality/view/account.py
```diff
@@ -250,6 +259,7 @@ def register():
         account = models.Account.make_account(
             username=form.w.data,
             email=form.n.data,
+            roles=[r.strip() for r in form.roles.data.split(',')]
         )
         account.set_password(form.s.data)
         account.save()
```
emanuil-tolev: GH not allowing me to comment on line 257 (new) so just leaving it here.

```form = RegisterForm(request.form, csrf_enabled=False, roles='api')```

That's an interesting way of adding a default value to a field - I think I've seen that, but probably years ago. So you just pass [name of field] = value as further keyword arguments and if the form constructor isn't specifically written to handle them, it just prefills the fields with the keyword args? That's quite cool.

portality/models/account.py
Steven-Eardley: My thought at the time was that returning ```None``` usually means you don't have API access, and the missing key is an accident so should be fixed by generating a new one. You're right about debugging though, it'll be a bit of a moving target as it generates new ones until save.

So if someone expects to have API access but their key can't be validated, do we expect them to login and generate themselves a new one? I guess that's straightforward. Anyway, there shouldn't be many opportunity for an account with the api role to be decoupled from its api_key record in data.

as a side note, why can't I put a function call in the dict.get() method? when I tried to write ```self.data.get('api_key', self.generate_api_key()) it always ran the generate, whereas if the default arg was ```None``` it would return the key from the dict.

portality/view/account.py
Steven-Eardley: Isn't this how the ```formcontext``` pre-fills the data? I wasn't aware I was doing anything novel here!

emanuil-tolev: Novel to me :). I think WTForms has had support for a long time but I'm used to just using 2 args with it (which can be a pain) so that was probably more so me learning about this!

portality/models/account.py
emanuil-tolev: > Anyway, there shouldn't be many opportunity for an account with the api role to be decoupled from its api_key record in data.

Yeah, I don't think there's currently a way for a key to be deleted, even admins can't do it. At worst a developer can go in and remove it by hand, but there's a lot of havoc you can wreak that way anyway.



> My thought at the time was that returning None usually means you don't have API access, and the missing key is an accident so should be fixed by generating a new one.

If we want to do that, we need to do it in the controller (/ view function / helper) that discovers the problem, and .save() the object. It isn't really possible to authenticate against a key that is stored only in memory, the user will never know what their key is even if you generate one on the fly.

I don't think even that is necessary though, if your key is not working you would go to your account and copy it from there, if it still doesn't work you'd contact DOAJ for support.

The way to remove API permissions from somebody is to remove the api role - they can keep the key. All our private routes should check for the API role.

emanuil-tolev: > as a side note, why can't I put a function call in the dict.get() method? when I tried to write self.data.get('api_key', self.generate_api_key()) it always ran the generate, whereas if the default arg was None it would return the key from the dict.

Hmm. Sorry, no idea beyond putting it in a temporary var.

Steven-Eardley:
@emanuil-tolev Is this where we want it now?

emanuil-tolev:
Yep. Sorry, it doesn't email when you add a commit, just replies. Feel free to finish / deploy.

PR #872, opened 2015-09-22
Feature/formcontext email refactor

Steven-Eardley:
I've moved the email sending functions out of ```formcontext.py```. No new functionality added.

### DIFF
--- a/portality/formcontext/emails.py
+++ b/portality/formcontext/emails.py
@@ -1,0 +1,213 @@
+from flask import url_for
+import json
+
+from portality import models, app_email
+from portality.core import app
+from portality.dao import Facetview2
+
+
+def send_admin_ready_email(application, editor_id):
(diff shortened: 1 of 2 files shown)


emanuil-tolev:
Hmm, cool. Maybe redundant question, but have you tried that these still work? I can't find a fast way to see if they are just copy/pasta operations into a new file. But it'd be really silly if an email dropped off completely just bc of a small runtime problem that we don't get alerted about.

Steven-Eardley:
All emails in that file apart from ```send_received_email()``` (to the applicant when an application is submitted) are tested in ```/doajtest/unit/test_review_emails.py``` - I can add a test for that one if it'd make you happier :)

emanuil-tolev:
Probably wise to do if you have time, but not a blocking problem. Thanks :)

PR #873, opened 2015-09-23
add journal retrieval to crud

emanuil-tolev:
Finally! @Steven-Eardley @richard-jones 

Important things:

1. Introduce journals, duh. Tests, api class, view func.

    EDIT: I've taken the Application struct that Richard wrote (since it has more type casts) and tacked on fields like oa_end that only journals have. Also journals and applications now expose both subjects and keywords.

2. Refactor applications a bit to use common logic with journals. Had to make a common Outgoing object since it was just the best way to share the logic I could come up with, and it allows more sharing in the future. Sorry about the name of the common object, I thought for 5 min but couldn't come up with "what's the parent of journal and application conceptually". Suggestions very welcome.

3. Introduce parity between incoming and outgoing applications with additional safeguards so people can't overwrite info they shouldn't be able to overwrite. Though they can supply that info now. Otherwise it's very annoying to update applications via the API (GET, modify what you want, delete some fields so it fits our reqs, PUT; rather than just GET, modify what you want, PUT).

### DIFF
--- a/doajtest/fixtures/article.py
+++ b/doajtest/fixtures/article.py
@@ -37,6 +37,10 @@ class ArticleFixtureFactory(object):
     def make_article_source():
         return deepcopy(ARTICLE_SOURCE)
 
+    @staticmethod
+    def make_article_apido_struct():
+        return deepcopy(ARTICLE_STRUCT)
+
 ARTICLE_SOURCE = {
     "id" : "abcdefghijk_article",
     "admin" : {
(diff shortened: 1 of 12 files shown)


emanuil-tolev:
Oopsies, tests. Fixing.

emanuil-tolev:
Not sure how but it doesn't look like "apc" of the bibjson is getting set correctly during tests - in the model, not the data object. When creating a model from the test journal source, .bibjson().apc is an empty dict despite "apc" clearly being present". This is the last error, I'll have to pick this up tomorrow unless somebody has an idea of how this could be. Obviously I've not touched how the models work, and I've not touched that part of the journal fixture or the tests that are failing, so I'm a bit confused atm.

emanuil-tolev:
@richard-jones @Steven-Eardley ready for review, tests fixed. Lesson of the week, .update does not recursively update keys inside dictionaries, it just overrides the top-level dict.

emanuil-tolev:
@Steven-Eardley @richard-jones I've got a PR ready for API in/out parity - it does need a few tests added, but I'd like to get it in front of you to review the logic as soon as I can. It's based on this branch unfortunately so this one needs to be merged first before there's even any point in putting up another PR with the same diff (and more).

Steven-Eardley:
It looks okay to me, and passes tests, but I don't really know about the overall dataobj stuff.

PR #874, opened 2015-09-24
Removed a few errant print statements.

Steven-Eardley:


portality/view/query.py
```diff
@@ -152,7 +152,6 @@ def query(path='Pages'):
         if associate_filter:
             terms.update(_associate_filter())
 
-        print terms
```
emanuil-tolev: omg

PR #876, opened 2015-09-25
[hotfix] change default search operator to AND

emanuil-tolev:
# HOTFIX, DO NOT MERGE HERE

A developer has noticed that our search API uses default OR operator (the ES default) but the search page uses AND (facetview specifies it as default). He thinks AND produces much more relevant results and I agree. Also it should behave similarly to the web search. And finally you can override the operator if needed by giving us "word1 OR word2" as a query.

### DIFF
--- a/portality/api/v1/discovery.py
+++ b/portality/api/v1/discovery.py
@@ -290,7 +290,8 @@ class ApplicationQuery(object):
                     },
                     "query" : {
                         "query_string" : {
-                            "query" : self.qs
+                            "query" : self.qs,
+                            "default_operator": "AND"
                         }
                     }
                 }
--- a/portality/api/v1/discovery.py
+++ b/portality/api/v1/discovery.py
@@ -249,7 +249,8 @@ class SearchQuery(object):
                     },
                     "query" : {
                         "query_string" : {
-                            "query" : self.qs
+                            "query" : self.qs,
+                            "default_operator": "AND"
                         }
                     }
                 }
--- a/portality/templates/doaj/api_docs.html
+++ b/portality/templates/doaj/api_docs.html
@@ -18,6 +18,12 @@
 
     <h2 id="search_api">Search API</h2>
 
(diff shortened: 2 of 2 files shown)


Steven-Eardley:
This makes sense to me, I think. :+1: - but do the corresponding docs need altering with this PR?

emanuil-tolev:
Yeah, we, uh, don't have docs on this :). I added some on it + a reiteration of Elasticsearch query string syntax in the beginning of the search docs (as well as under start of advanced usage section).

emanuil-tolev:
If this still looks good I'd probably like to roll it out quickly, since, well, the guy is trying to integrate with the API so let's see what he makes. I think he runs a site(s) so more useful feedback may come.

Steven-Eardley:
Yup, ready to go. I'm convinced.

Steven-Eardley:
the default operator was set to AND site-wide some time ago, but the bit of code I was accustomed to using to build queries (```Facetview2``` class in ```portality/dao.py```) still has the default parameter as ```OR```, which is familiar when I use ES directly.

emanuil-tolev:
> the default operator was set to AND site-wide some time ago, but the bit of code I was accustomed to using to build queries (Facetview2 class in portality/dao.py) still has the default parameter as OR, which is familiar when I use ES directly.

Hmm. I'm not sure what we should do there tbh. It kind of sucks to have the JS part of the site use AND and the Python part use OR by default, but the actual ES default *is* OR. @richard-jones what do you think?

I am leaning towards we should use AND in python as well (not to mention we pretty much never use string queries in code). If this doesn't break any existing code (proven via passing tests and analysing the codebase - are there any queries that would be affected at all? I think not) I'd do that.

In terms of how, adding "default_operator": "AND" to every query string query (and only to those) in portality/dao.py or the relevant dependency would be the route to go for.

emanuil-tolev:
I'll roll this hotfix out now.

PR #884, opened 2015-10-01
Feature/api inout parity

emanuil-tolev:
@Steven-Eardley @richard-jones Ahoy! Possibly final code PR for CRUD API. A bit more in docs coming, and should be done!

We already have parity in applications (in/out) and journals are only out atm. So this adds parity to articles. This was a bit tricky because of journal info. Unfortunately it seemed quite difficult to go with the proper solution of supplying a journal identifier to people (e.g. for journal info see X) because some mutable properties like start page, end page, volume and issue make a bit more sense in the journal part of the article **and** more importantly people would've been able to specify journal_id on the way in. Which we could ignore, but it could get quite confusing (how do you assign an article to a journal then?).

So I've avoided that can of worms by keeping the data object quite close to the internal data structure for this iteration of the API. People just have properties they can set, and properties we will ignore. Journal-intrinsic properties like title, license, country etc. will be ignored on article create/update. start_page, end_page, volume and number (issue) are allowed to change.

Also fixes a little boo-boo in portality/api/v1/data_objects/article.py:215 as well, which meant start and end pages were never retrieved for an article. Hooray tests.

### DIFF
--- a/doajtest/fixtures/article.py
+++ b/doajtest/fixtures/article.py
@@ -36,6 +36,15 @@ class ArticleFixtureFactory(object):
     @staticmethod
     def make_article_source():
         return deepcopy(ARTICLE_SOURCE)
+    
+    @staticmethod
+    def make_incoming_api_article():
+        template = deepcopy(ARTICLE_SOURCE)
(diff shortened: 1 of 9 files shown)


Steven-Eardley:
Looks good.

PR #886, opened 2015-10-02
Removed the in_doaj from the journal csvs, which was slightly meaning…

Steven-Eardley:
…less since they were all in the doaj.

# Hotfix


emanuil-tolev:
:+1: we'll deploy on Mon. Don't now, Friday deploy not good deploy :).

Issue #877, opened 2015-09-28
Have we lost page number ordering on ToCs?


dommitchell:
https://doaj.org/toc/1823-4690/10/8

Start page: 1009. 1035, 1054 etc, 958, 972

Steven-Eardley:
Oh dear, that's unfortunate. They're appearing first because ```1``` comes before ```9```. I'll have a look at ensuring it considers entire number!

emanuil-tolev:
Makes you wonder, doesn't it, a problem that's existed for as long as computers have differentiated between strings and numbers (using the ASCII table at least), yet there's still no way to fix it generally.

Steven-Eardley:
This one was an excellent brain-teaser. 

The crux of the problem was ``` _sort_articles()``` in ```portality.models.article.py``` was always getting ```None``` for the start_page, then everything was considered unsorted:

```python
for art in articles:
    sp = art.get("bibjson",{}).get("start_page", None)

    # can't sort anything that doesn't have a start page
    if sp is None:
        unsorted.append(art)
        continue
    ...
```

We were trying to get the start_page from the bibjson but it wasn't structured as this function expects - the query providing the data is one that pulls out individual fields, so we have to access the start_pages field like this instead:

```python
sp = art.get("bibjson.start_page", [None])[0]
```

Because the results we are reading from look like this:
```json
"hits": {
    "total": 3,
    "max_score": null,
    "hits": [
      {
        "_index": "doaj",
        "_type": "article",
        "_id": "",
        "_score": null,
        "fields": {
          "bibjson.journal.number": [
            "Issue 1"
          ],
          "bibjson.abstract": ["Abstract"],
          "bibjson.title": [
            "The Title"
          ],
          "bibjson.journal.volume": [
            "2014"
          ],
          "bibjson.end_page": [
            "5102"
          ],
          "bibjson.start_page": [
            "5093"
          ],...
```

The reason we were getting semi-sensible results all this time was that the ElasticSearch query includes  ```"sort": "bibjson.start_page"```, so it was pre-sorting the articles, but not doing a complete job of it. We were being a little unfair to ES, really, because we do store the start_pages as strings in case it's something like 'ix'.

Steven-Eardley:
Fix is live

dommitchell:
Perfect!

PR #888, opened 2015-10-03
Fixed article ordering on TOC pages by getting the right field from t…

Steven-Eardley:
…he search results in the sort function - previously it was always getting None for the start_page, and falling back to ES ordering.

# Hotfix

For issue #877.


emanuil-tolev:
:+1:

Issue #801, opened 2015-07-10
Bring facet labels and statuses in line with each other


dommitchell:
So the final list in the Admin view of either an Application or a Reapplication will look like this:

'reapplication in progress' changes to 'reapplication pending' (only visible to admin and publishers)
reapplication submitted (only visible to admin and publishers)
pending (only visible to asseds, editors, admin)
in progress (only visible to asseds, editors, admin)
completed (NEW - only visible to asseds, editors, admin)
ready (only visible to editors and admin)
accepted (only visible to admin)
rejected (only visible to admin)
on hold (only visible to admin)

The facet under the Application tab in the Admin view will look the same, with the following mapping:

'reapplication' changes to 'reapplication pending'
'submitted' changes to 'reapplication submitted'
pending
in progress
completed (NEW)
accepted
rejected
on hold
ready

greboun:
the facets in the admin interface for applications are **reapplications**, 
,**submitted** , pending, in progress , ready, accepted, rejected,  on hold.
In this **reapplications**=**reapplications pending** with the publisher, **submitted**= reapplications submitted by a publisher the others are statusses with the editorial team
In the dropdown of the application form we have **reapplication submitted**, **reapplication in progress**, pending,in progress , ready, accepted,rejected on hold.
![image](https://cloud.githubusercontent.com/assets/12031556/8617403/bf1119e6-2703-11e5-93b7-ad4d0f6ce9d5.png)

**reapplication in progress** means that the reapplication is in progress with the publisher, and not yet submitted. _We should change this to **reapplication pending**_ in line with the facet view
Once the reapplication has been submitted the maned will change the status to **pending** and allocate to an editor..(S)he starts working and the reapplications gets status **in progress*
NOTE from hereon we can only see that we work with a reapplication if we have put in a remark in the notes section, or indirectly when there already is a username filled in in the form.
ANY OTHER WAY YOU CAN THINK OF TO DISTINGUISH REAPPLICATIONS FROM NEW APPLICATIONS IN THE FORM??

dommitchell:
> ANY OTHER WAY YOU CAN THINK OF TO DISTINGUISH REAPPLICATIONS FROM NEW APPLICATIONS IN THE FORM??

@greboun all reapplication records get a 'View original record' link at the top.

dommitchell:
@greboun can you complete the task description so it clearly states what the changes are that we want? Thanks!

greboun:
what we want is in the application view dropdown
[...]

PR #891, opened 2015-10-08
Feature/workflow final changes

Steven-Eardley:
These are changes to the workflow project based on feedback.

### DIFF
--- a/portality/formcontext/choices.py
+++ b/portality/formcontext/choices.py
@@ -103,12 +103,6 @@ class Choices(object):
         ('ready', 'Ready'),
     ]
 
-    _application_status_publisher = [
-        ('', ' '),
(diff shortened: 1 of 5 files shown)


portality/formcontext/choices.py
```diff
@@ -100,10 +100,15 @@ class Choices(object):
     ]
 
     _application_status_editor = _application_status_base + [
-        ('on hold', 'On Hold'),
         ('ready', 'Ready'),
     ]
 
+    _application_status_publisher = [
```
Steven-Eardley: Hmm, giving the publishers a list of allowed statuses doesn't do anything, it turns out - the publisher doesn't see the application status on the form, as I had imagined from the description of #801. I'll simplify this in another commit as it's misleading just now.

PR #893, opened 2015-10-11
Feature/crud api docs

emanuil-tolev:
Hi @richard-jones, @Steven-Eardley - this is pretty much the final version of the main part of the CRUD docs. Specifically we are now able to generate Swagger Spec out of our DataObject structs! So for any DataObject, we get the interactive interface and example (where it says Model: on https://doaj.org/api/v1/docs) just by writing the struct for it.

It does need a few tests to make sure the right docs are being produced (and continue to be throughout future changes).

Additionally I've taken out all the Swagger YAML, there's a new decorator that allows us to actually specify the Swagger JSON ourselves - whether that be dynamically generated or loaded from a file on app start-up, this is now separated from the API code.


emanuil-tolev:
Oh, also to add: use the new dynamic swag spec generation on articles and applications. The code is extremely generic (it's in the DO library) so this should just be a function call.

portality/api/v1/crud/journals.py
```diff
@@ -5,6 +5,33 @@
...
+                    "description": "<div class=\"search-query-docs\">DOAJ journal ID. E.g. 4cf8b72139a749c88d043129f00e1b07 .</div>",
+                    "required": True,
+                    "type": "string",
+                    "name": "journal_id",
+                    "in": "path"
+                }
+            ],
+            "tags": ['CRUD']
+        }
+        template['responses']['200']['schema']['title'] = 'Journal schema'
```
Steven-Eardley: If you're defining the template here anyway, can't you just plop the data straight in above all at once?

Steven-Eardley:
It's looking good! Nice to see the massive doc comments factored out, too.

portality/api/v1/crud/journals.py
emanuil-tolev: Good point, I refactored the template into common.py, into the CrudApi class. Might need to move it up to the Api class even. Now all the various endpoints actually share the template, deepcopy it, and add their stuff.

emanuil-tolev:
Hey @Steven-Eardley @richard-jones  - this is ready for a final review and merging now. Apologies for the size of the PR. The best thing to do IMO is to checkout this branch, run it locally and head over to http://localhost:5004/api/v1/docs to see if anything seems grossly out of whack (after all the end-game of this is to produce new usable docs), or alternatively if everything is dandy.

I'd welcome comments on how the documentation is produced code-wise as well, but obviously it is a pretty big introduction of lots of functionality, so I'd understand if you prefer to just glance over it as it's pretty self-contained. (Not encouraging that, but I do want it merged this week if it seems good :).)

portality/api/v1/crud/articles.py
```diff
@@ -42,7 +71,7 @@ def create(cls, data, account):
 
         # Check we are allowed to create an article for this journal
         if not XWalk.is_legitimate_owner(am, account.id):
-            raise Api404Error()                 # not found for this account
+            raise Api403Error()
```
emanuil-tolev: I thought it was too confusing to return a 404 for a **related** resource. Like, I send a POST create request and I get a 404 .. what? What wasn't found? A 404 works on retrieve and update and such, but not really on create - should just use 403.

greboun:
Hi Emanuil

I have been looking at the API doc  (https://doaj.org/api/v1/docs#)
I have one question
Can publishers upload (re) applications via the API and will they not need
to do this through a CSV anymore?
https://github.com/DOAJ/doaj/blob/develop/README.md#journalapplication-data-model

TOM @greboun

Dr. Tom Olijhoek
Editor-in-Chief  DOAJ
Support us: www.doaj.org/support

emanuil-tolev:
Hi @greboun . I think they might be able to. I am trying it out to make sure on the test site.

PR #894, opened 2015-10-12
Hotfix: detect and strip Byte Order Mark from UTF8 csvs

richard-jones:


emanuil-tolev:
:+1: we should merge locally and roll out.

PR #898, opened 2015-10-12
fixes to article ingest pipeline to defend against es-refresh race-co…

richard-jones:
…ndition bugs in rapid re-import of matching articles.  Plus tests.

This works by adding the following:

* a "differentiate" keyword on the DAO save() method which forces last_updated to be modified such that it is different to any previous version of the record if necessary.  Basically this means that if a record is saved twice in a second, it will block until the next second ticks over before saving.
* a block() method on the DAO which monitors an id for a last_updated time that can be /searched/ for (not pulled) which is equal to or greater than the desired time.
* calling save(differentiate=True) on articles when an XML doc is processed (and nowhere else in the codebase)
* calling block() after large loops over documents/result sets

The reasons for doing it this way, and not in the esprit style blocking save is that we don't want to do this for every save, only on the last save of many, and since the records are often pulled out of iterators, it's difficult to know when that is.  So instead, we separate the save from the block, and then just block the thread itself, based on the id/last_updated of the last saved object.

I realise, also, that we could do this with version numbers, and that would get rid of the need to differentiate on save, but I already had experience with this above approach, and know it works well, and I don't think the difference of approach really makes much difference, especially in this use case.

### DIFF
--- a/doajtest/unit/resources/article_uploads.xml
+++ b/doajtest/unit/resources/article_uploads.xml
@@ -28,7 +28,7 @@
         <affiliationsList>
             <affiliationName affiliationId="1"/>
         </affiliationsList>
-        <fullTextUrl/>
+        <fullTextUrl>http://doaj.org/testing/url.pdf</fullTextUrl>
         <keywords>
             <keyword>Introduction</keyword>
             <keyword>Littérature amérindienne</keyword>
--- a/doajtest/unit/test_article_upload.py
+++ b/doajtest/unit/test_article_upload.py
@@ -299,3 +299,28 @@ class TestArticleUpload(DoajTestCase):
 
         found = [a for a in models.Article.find_by_issns(["1234-5678", "9876-5432"])]
         assert len(found) == 0
+
+    def test_09_duplication(self):
+        j = models.Journal()
+        j.set_owner("testowner")
+        bj = j.bibjson()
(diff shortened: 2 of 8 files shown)


richard-jones:
Incidentally, in testing I was able to reproduce the duplication, and then introduce a fix which prevented it, so I'm quietly confident that this will put a lid on this issue.

emanuil-tolev:
> I realise, also, that we could do this with version numbers, and that would get rid of the need to differentiate on save, but I already had experience with this above approach, and know it works well, and I don't think the difference of approach really makes much difference, especially in this use case.

Hahah, yes, of course, it was just a suggestion. I don't think we should necessarily break consistency just to do the same thing differently, unless the speed difference really matters. it was more of a "maybe in the future we should adopt this approach, it seems better and gives us more flexibility in terms of letting conflicts bubble up to the user in a web context". :+1: on pull request. If I can't today, I'll roll it out tomorrow. @Steven-Eardley feel free as well.

Just note that this is not related to a specific project and can go out ASAP, so if it's not me rolling it out, make sure to create a hotfix branch out of Richard's feature/article_ingest_fixes, but without any ```develop``` stuff we might not be ready to release (so git cherry-pick, not git merge, the feature branch into a hotfix branch).

emanuil-tolev:
Done as a hotfix, this is now live

PR #907, opened 2015-10-21
Hotfix/2015 10 20 api articles retrieve auth

Steven-Eardley:
Updates to Articles endpoint, tests, and the corresponding swagger.

```#swag #yolo```

### DIFF
--- a/portality/api/v1/crud/articles.py
+++ b/portality/api/v1/crud/articles.py
@@ -107,7 +107,7 @@ class ArticlesCrudApi(CrudApi):
 
         # at this point we're happy to return the article if it's
         # meant to be seen by the public
-        if ar.get('admin').get('in_doaj', False):
+        if ar.is_in_doaj():
             return OutgoingArticleDO.from_model(ar)
 
         # as long as authentication (in the layer above) has been successful, and the account exists, then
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: portality/api/v1/crud/articles.py [hunk 1]
```diff
@@ -95,20 +95,26 @@ def retrieve_swag(cls):
...
 
-        # is the article id valid
+        # is the article id valid?
         ar = models.Article.pull(id)
         if ar is None:
             raise Api404Error()
 
+        # at this point we're happy to return the article if it's
+        # meant to be seen by the public
+        if ar.get('admin').get('in_doaj', False):
```
emanuil-tolev: ```ar.is_in_doaj()``` probably better idea. If you read it you'll see that it also assumes ```{}``` if the first get returns nothing, otherwise the second one will cause a 500 error. But seeing as a method already does it .. :)

emanuil-tolev:
> #swag #yolo

:D :D :D I did not see this initially.

Steven-Eardley:
> ar.is_in_doaj() probably better idea.

I'll change both Journals and Articles, then.

emanuil-tolev:
Uh oh. Yes, thanks.

Steven-Eardley:
Changes made

emanuil-tolev:
:+1: merge on ok build!

Issue #746, opened 2015-05-29
Javascript widget - Simple search


emanuil-tolev:
Simple search widget that will point people towards /search on DOAJ, no embedded search results.

emanuil-tolev:
Potential load issues, but #747 would be far worse. Thoughts on server load placed there.

Nimphal:
This is now on test, the code sample on test can be embedded wherever you like for testing pleasure :)

Nimphal:
Actually, it seems to be having some https issues on test, so not ready yet.

Nimphal:
Okay, not it is ready and an example can be viewed here https://testdoaj.cottagelabs.com/widgets

emanuil-tolev:
@dommitchell looking good at URL above?

dommitchell:
Thanks! I like it!! It looks great.

First thing I did is try to embed it in my wordpress blog and it didn't work. I know that, for example, Slideshare has Wordpress specific embed code for Wordpress. Is the code you provide at https://testdoaj.cottagelabs.com/widgets universal?

Nimphal:
How did you include it in your wordpress? Did you edit your theme? Or did
you paste into a blogpost? If the latter, then you'd need something like
this to make it work https://wordpress.org/plugins/html-javascript-adder/ .
WordPress does not take kindly to random JS scripts in blogposts generally.
The widget is as general as they get, just some JS to be loaded. It works
when included in HTML pages (as you can see on the example page, it's
included as the example code shows). WordPress is not a standard platform
in that case and requires a bit more legwork to make custom widgets work.

2015-10-21 13:31 GMT+01:00 dommitchell <notifications@github.com>:

dommitchell:
> Did you edit your theme? Or did you paste into a blogpost?

Yup, just pasted it in. Thanks for the guidance!

I think that this widget is good to go live whenever the time comes. Thanks @Nimphal !

emanuil-tolev:
The live documentation page is pointing to the test server's widget, it should point to the production one.

Steven-Eardley:
Should the we consider a transparent background (or a border) for the DOAJ logo in the widgets? This is how it looks when I chuck the simple search into a hero unit:

<img width="307" alt="screen shot 2015-10-29 at 20 56 32" src="https://cloud.githubusercontent.com/assets/1446355/10831831/be460d68-7e7f-11e5-8383-96a061231d2a.png">

dommitchell:
[...]

PR #914, opened 2015-10-28
Update widget docs

Nimphal:
For #746


emanuil-tolev:
I like how the domain names were correct but not the filenames :+1:

PR #915, opened 2015-10-28
[hotfix] article deduplication hotfix

emanuil-tolev:
# HOTFIX, DO NOT MERGE HERE

This is an attempt to get multiple duplicates for an article in one go. Currently for some reason the code in production will return the article itself when asked for duplicates (?!) and thus there are some .. complications when you ask for multiple duplicates of one article.

Technically (by the tests) the duplication detection is working correctly for multiples.

1. create article A with URL lala
2. create article B with URL lala
3. create article C with URL lala

get_duplicate(A, all_duplicates=True) will return B and C.

~~However, when running the new portality/scripts/remove_duplicate_articles.py **ALL** of them get deleted - A, B and C are all gone. I'm not sure why, but obviously that needs to be figured out before this goes on live. @Steven-Eardley if you have time tomorrow a bit of pairing might be helpful if I still haven't figured it out. I think it might have to do with~~

- ~~state of the index - we are deleting from a list we are reading from (not using scroll search - so we might need to use scroll search)~~
- ~~when we delete an article that's not reflected in Python memory. So I get to A and ask for duplicates. I get B and C. I delete them. But then I get to B and ask for duplicates. I get A and C, so I delete them. Now they're all gone. This is just conjecture, I may be wrong.~~

### DIFF
--- a/doajtest/unit/test_article_match.py
+++ b/doajtest/unit/test_article_match.py
@@ -155,8 +155,63 @@ class TestArticleMatch(DoajTestCase):
         expected = sorted([a, a2])
         xwalk = article.XWalk()
         l = xwalk.get_duplicate(z, all_duplicates=True)
+        assert isinstance(l, list), l
+        assert l
+        l.sort()
+        assert expected == l
+
+    def test_05_with_doi_instead(self):
+        # make ourselves a couple of example articles
+        a = models.Article()
+        b = a.bibjson()
+        b.title = "Example A article with a fulltext url"
+        b.add_identifier('doi', "10.doi")
+        a.save()
+
+        # wait for a bit to create good separation between last_updated times
+        time.sleep(2)
+
+        a2 = models.Article()
+        b2 = a2.bibjson()
+        b2.title = "Example B article with a fulltext url"
+        b2.add_identifier('doi', "10.doi")
+        a2.save()
+
+        # wait for a bit to create good separation between last_updated times
+        time.sleep(2)
+
+        # create an article which should not be caught by the duplicate detection
+        not_duplicate = models.Article()
+        not_duplicate_bibjson = not_duplicate.bibjson()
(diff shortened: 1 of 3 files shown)


emanuil-tolev:
I believe I've tackled the original article being deleted. @Steven-Eardley could you give this script a go?

To test:

```python
# Take stock of number of articles in index. Mine was 101.
from portality import models
duplicates = []
from doajtest.fixtures import ArticleFixtureFactory
a = models.Article(**ArticleFixtureFactory.make_article_source())
a.save()
a = models.Article.pull('abcdefghijk_article')
a.id  # should go without error
id_ = a.makeid(); duplicates.append(id_); a.set_id(id_); a.save(); a = models.Article.pull('abcdefghijk_article')
id_ = a.makeid(); duplicates.append(id_); a.set_id(id_); a.save(); a = models.Article.pull('abcdefghijk_article')
id_ = a.makeid(); duplicates.append(id_); a.set_id(id_); a.save(); a = models.Article.pull('abcdefghijk_article')
duplicates  # review
# run script
# check number of articles in index. Expect 102 for success.
# check script output. Interested in the *first* article the script sees 
# - that should have remained in the index, the others - gone.
models.Article.pull('abcdefghijk_article')  # could be None, could be the article
models.Article.pull(duplicates[0])  # could be None, could be the article
models.Article.pull(duplicates[1])  # could be None, could be the article
models.Article.pull(duplicates[2])  # could be None, could be the article

# The point is, only one of the 4 lines above should output an object in iPython.
# If so, success.
```

To run script:

```python portality/scripts/remove_duplicate_articles.py -y -g```

-g for ghost (do not snapshot deletes), -y for agreeing to run it and having thought about whether you need to save to a log file or not.

emanuil-tolev:
And review the code of course.

Steven-Eardley:
Works like a charm! :+1: ready to go

Issue #785, opened 2015-06-30
 Both Managing Editors and Editors should be able to assign or remove AssEds to/from applications


dommitchell:
Both Managing Editors and Editors should be able to assign or remove AssEds to/from applications

richard-jones:
Before approving this for TnM just a note that this is a non-trivial bit of work.  Probably would take upwards of a day to do.  Ok, to add to the list?

dommitchell:
Yes please!

dommitchell:
@richard-jones wouldn't this be better as part of the workflow project that @Steven-Eardley is doing?

richard-jones:
I've done a review of the work to be done here, and remember why this was a bit fiddly.  It's to do with the managing editor's update workflow.  We need to support the following actions in the ManEdApplicationFormContext:

* Addition of an Editor Group where there was previously no group.  This should then enable the Editor select box to allow the selection of an editor from that group

* Removal of an Editor Group, where there was previously a group.  This should clear the Editor select box entirely, and set the value to the empty string.

* Change an Editor Group from one group to another.  This should set the Editor value to the empty string and allow the selection of a new editor from that group.

Probably the best way to achieve this is to introduce some more jquery into the maned form, which will monitor the editor_group field for changes, and carry out the appropriate modifications to the editor select box.  This will include an ajax call to a new endpoint that we'll need to create which will give back the editors in a given editor group, which will then be used to populate the select box.

It may be necessary to validate the editor against the editor_group on form submission, in case there's a problem with the javascript in the front end.  So check that the specified editor actually appears in the specified editor_group, and return an error if not.

I don't think that any changes are required to the Editor's form context - they can already modify the assigned editor, and they can't change the editor group.

@Steven-Eardley - does that all make sense to you?

Steven-Eardley:
> I don't think that any changes are required to the Editor's form context - they can already modify the 
> assigned editor, and they can't change the editor group.

Isn't the above functionality all this issue requests for the Managing Editors too? Group management within the forms seems above and beyond!

richard-jones:
@dommitchell - we've been talking more about this, and actually have realised we may have got the wrong end of the stick with regard to what is needed here.  Could you tell me which of the following you are after:

1/ Managing Editors should be able to change the Associate Editors assigned to an application

2/ Editors should be able to add users with the role "associate_editor" or "editor" to their editorial group (or remove them).

dommitchell:
[...]

Issue #830, opened 2015-08-04
Evidence that matching doesn't always work for article uploads


dommitchell:
https://doaj.org/toc/2166-6482/29/4
https://doaj.org/toc/2166-6482/29/3
https://doaj.org/toc/2166-6482/29/2
https://doaj.org/toc/2166-6482/29/1

In #711 @richard-jones says:
"there's no doubt that the matching algorithm is working correctly - I was able to do a mock match for one of these and found the other." https://github.com/DOAJ/doaj/issues/711#issuecomment-127292406 and yet we can see that there are circumstances when this does not work correctly.

In this particular example, the original XML uploads were affected by the recent upload stoppage so perhaps something there is causing this. Thoughts?

dommitchell:
More evidence from Hindawi:
"Yes, uploading files is working fine, thank you. But I noticed that some of the re-uploaded files appear duplicated on your website! I think there is a bug occurs when parsing the metadata XML on your systems."

https://doaj.org/toc/1563-5147/2014
101473 3 times
101626 3 times
101808 3 times

https://doaj.org/toc/1687-4129/2014
102621 3 times
103418 3 times

dommitchell:
More from Scielo, uploaded on 14th August:

dommitchell:
Some more articles uploaded yesterday, 17th August

dommitchell:
I think we have found out why duplication is occurring:

Refreshing the XML upload page causes articles to be duplicated. "after uploading any record to your side it appears as "pending" in our admin page; if we refresh "F5" this page while there are pending records, this will cause the records to be duplicated. The repetition depends on the refresh count! If I refreshed the page 2 times the record will appear 3 times (1 originally submitted + 2 times due to F5 clicks). "

emanuil-tolev:
So they're not getting asked "do you want to submit again", they're just refreshing the status page? Obviously if they're ignoring the warning the browser gives upon form resubmission, they're ... going to resubmit the same file. There is code to prevent the processing of duplicated records though, so even if they literally come back and upload the same file 10 min later, both will be processed and only the first should take effect.

richard-jones:
We need to do the following things:

1/ add redirect on form submission
2/ Remove "refresh" requests from ingestarticles.py
3/ Introduce esprit's blocking save feature on final article submitted in a file

emanuil-tolev:
A spot of trouble with this, I can't get my local one to process things properly for some reason. Not forgotten about it :)

[...]

PR #919, opened 2015-11-02
Hotfix/issue 785 maneds assign associates

Steven-Eardley:
## Hotfix

For issue #785 - Managing Editors are now able to assign editors as well as editor groups on the Application and Journal forms. Javascript has been added to the html to read the editor options whenever a new editor group is added, and the choice is validated in the formcontext.

This should probably be demoed on the test server before full release.

### DIFF
--- a/portality/formcontext/formcontext.py
+++ b/portality/formcontext/formcontext.py
@@ -289,13 +289,12 @@ class PrivateContext(FormContext):
         if apply_notes_by_value:
             self.target.set_notes(tnotes)
 
-    def _populate_editor_field(self):
+    def _populate_editor_field(self, editor_group_name):
         """Set the editor field choices from a given editor group name"""
-        egn = self.source.editor_group
-        if egn is None:
+        if editor_group_name is None:
             self.form.editor.choices = [("", "")]
         else:
-            eg = models.EditorGroup.pull_by_key("name", egn)
+            eg = models.EditorGroup.pull_by_key("name", editor_group_name)
             if eg is not None:
                 editors = [eg.editor]
                 editors += eg.associates
@@ -623,7 +622,8 @@ class ManEdApplicationReview(ApplicationContext):
         self.form.application_status.choices = choices.Choices.application_status("admin")
 
         # The first time the form is rendered, it needs to populate the editor drop-down from saved group
-        self._populate_editor_field()
+        egn = self.form.editor_group.data
+        self._populate_editor_field(egn)
 
 class EditorApplicationReview(ApplicationContext):
     """
@@ -742,7 +742,8 @@ class EditorApplicationReview(ApplicationContext):
             self.form.application_status.choices = choices.Choices.application_status("editor")
 
         # get the editor group from the source because it isn't in the form
-        self._populate_editor_field()
+        egn = self.source.editor_group
+        self._populate_editor_field(egn)
 
 
 class AssEdApplicationReview(ApplicationContext):
@@ -1233,7 +1234,8 @@ class ManEdJournalReview(PrivateContext):
 
     def _set_choices(self):
         # The first time this is rendered, it needs to populate the editor drop-down from saved group
-        self._populate_editor_field()
+        egn = self.form.editor_group.data
+        self._populate_editor_field(egn)
 
     def finalise(self):
         # FIXME: this first one, we ought to deal with outside the form context, but for the time being this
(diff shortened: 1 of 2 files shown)


portality/formcontext/formcontext.py
```diff
@@ -1191,6 +1213,21 @@ def source2form(self):
         self._set_choices()
         self._expand_descriptions(["publisher", "society_institution", "platform"])
 
+    def pre_validate(self):
```
emanuil-tolev: Is this validator exactly the same as the one above at https://github.com/DOAJ/doaj/pull/919/files#diff-a08df963abb007b5bc03439f44aaa7b9R470 ?

portality/formcontext/formcontext.py
```diff
@@ -1208,11 +1245,19 @@ def patch_target(self):
 
 
     def _set_choices(self):
```
emanuil-tolev: I suppose the same is true here - same as the one on line 603 above?

portality/templates/formcontext/maned_application_review.html
```diff
@@ -127,4 +127,45 @@ <h3 class="heading-aligned-with-fields">Details of person making this applicatio
...
+$("#editor_group").change(function() {
+    var ed_group_name = $("#s2id_editor_group").find('span').text();
+    $.ajax({
+        type : "GET",
+        data : {egn : ed_group_name},
+        dataType: "json",
+        url: "{{url_for('admin.eg_associates_dropdown')}}",
+        success: function(resp)
+        {
+            // Get the options for the dropdown fron our ajax request
```
emanuil-tolev: little typo :)

portality/templates/formcontext/maned_journal_review.html
```diff
@@ -141,4 +141,45 @@ <h3 class="heading-aligned-with-fields">Old journal information</h3>
 <script type="text/javascript">
     var notes_deletable = true;
 </script>
+
+<script>
+// When the editor group field is changed, refresh the options for editor
+$("#editor_group").change(function() {
```
emanuil-tolev: Is this exactly the same as the one on the application review? If so this is quite easily factored out to a JS file to a named function (in any current JS file that's loaded everywhere, or feel free to make a new one and stick it in _js_includes.html under the correct conditions) and then called on .change.

So this becomes ```$("#editor_group").change(function() {editor_group_changed();}``` .

If there are only small differences between the journal and application .change handlers you can even parametrise them at this point, passing them as function args.

emanuil-tolev:
Just a few questions around maintainability. Let me know if factoring out the same/similar code in the Python section is a pain - perhaps we can look at using composition there instead of inheritance, and this might not be that hard to do. For the moment I'd really make it at least with inheritance, even though that can be a bit clunky, as having to change it in 2 places afterwards would not be ideal.

If neither converting to composition, nor doing it within the current inheritance model is feasible, then leaving a comment in each Python function pointing to its sibling is probably a good idea, so that whoever changes one, at least has a chance of reading the comment and changing the other.

The JS should definitely be refactorable into one piece, but I'm happy to help on exactly how!

I'll be at an event tomorrow but will try to look inbetween.

* * *

Also one final thing - for now we have the dedupe script doing #830 running .. quite slowly (1s per duplicate delete after all). So I'd avoid hotfixing if possible - technically it shouldn't affect a running environment and script but I'd rather not find out right now :).

Steven-Eardley:
Yeah, I just duplicated it because the scope was just expanded to Journals today! I'll do a nicer refactor to share the code - it is identical, I was just being lazy.

portality/templates/formcontext/maned_journal_review.html
Steven-Eardley: hmm, factoring this out into a static file isn't as easy as I'd hoped, because I rely on defining the script in the template to use ```url_for()``` as part of my ajax request.

Steven-Eardley:
@emanuil-tolev this is ready for another look. I've fixed the code duplication issue, plus fixed a bug where the drop-down and the autocomplete could become out of sync.

emanuil-tolev:
Awesome, looks good now :+1: . I'm thinking of temporarily stopping the script in #830 so we should be able to roll this out soon.

Steven-Eardley:
Okay, I'll put it on the test server for now.

emanuil-tolev:
You can release if you like, job stopped.

Steven-Eardley:
Cheers! I've just remembered to do the cache bust after checking it works on test.

Issue #920, opened 2015-11-05
Contact us link on API page returns error


dommitchell:
When not logged in, the Contact Us link under Authenticated Routes on https://doaj.org/api/v1/docs returns an error: https://doaj.org/api/v1/%7Bcontact_us_url%7D

Steven-Eardley:
This is live now. I also fixed the CSS which was causing some discrepancies in font sizes. (bonus!)

dommitchell:
Great!

PR #922, opened 2015-11-05
Hotfix/2015 11 05 jinjified swag api info

Steven-Eardley:
Should fix issue #920

Also, fixed some CSS wobbliness when going to the API page.


emanuil-tolev:
Wow, awesome. So long it still works (the page) which I assume you tried (I can't just now :(), release!

Issue #708, opened 2015-05-05
Add Tooltip to APC facet


dommitchell:
We'd like to add a tooltip to the APC facet to allow people to click out to this link: https://doajournals.wordpress.com/2015/05/11/historical-apc-data-from-before-the-april-upgrade/.  I understand this means adding a component as per the discussion here. #706

dommitchell:
Adding this for tnm assessment

richard-jones:
I'm not sure I totally understand what's being asked for.  Can you walk me through what you imagine happens when the user interacts with that facet?

dommitchell:
Hmm I'm not entirely sure. It was something @emanuil-tolev suggested. 

Either way, there needs to be a link in the facet label (like a question mark perhaps: "What's this?") that, when clicked, takes people out to the blog post above.

richard-jones:
How about something like the text "learn more", which is a link out to the desired page?

Would this text need to be visible even when the facet is collapsed, or only if it is opened?  If the latter, I'd probably position this next to the control buttons.

We could make this an option in fv2 so that you can add such a link to any of the facets.

dommitchell:
Yes, something like that sounds great. It doesn't need to be visible when the facet is collapsed.

dommitchell:
What's the due date on this one?

richard-jones:
I've now put a PR in for this one.  The behaviour is that there's a "learn more" link that appears by the APC facet, and when you click it it gives some blurb and then tells you to see a blog post for more details.

If you check this out when it is available on the test server, and let me know if you want the text of the tip changed.

dommitchell:
Please change 'learn more' to 'What do these figures mean?'

Please change: 'Our current knowledge about APCs for journals in DOAJ. For more information see this blog post (opens in new window).' to 'The data here may not be complete. For more information see this blog post (opens in new window).'

richard-jones:
Made a hotfix PR for this

Steven-Eardley:
Deployed on live now!

greboun:
Looks good
This one we really needed
Thanks



Dr. Tom Olijhoek
Editor-in-Chief  DOAJ
Support us: www.doaj.org/support

dommitchell:
Perfect!

PR #947, opened 2015-11-24
upgrade to fv2 version which allows tooltip link to be set, and updat…

richard-jones:
…e configuration for public search

Hotfix.

As per final changes to #708


emanuil-tolev:
:+1: can be deployed - I can't just now but will do tomorrow if Steve doesn't chance upon it

PR #954, opened 2015-11-26
Add Bulk Article API

emanuil-tolev:
This PR has been updated to include in-memory deduplication of incoming bulk article create requests.


emanuil-tolev:
@richard-jones @Steven-Eardley This PR has been updated to include the in-memory deduping - would you mind reviewing this again?

It's been up for review for a long time and there's been no other comments so I assume everything else is OK. The new salient bit is at https://github.com/DOAJ/doaj/pull/954/files#diff-1d8d41eb7549fffc0a31afdc7351973cR31 , and you know where to look for tests (there are now 2 unit tests for article creation, one with multiple different articles and one with duplicates).

Steven-Eardley:
It looks good, but we'll need to rethink the api_key in the functional tests - running them requires finding a valid one before running the test.

PR #967, opened 2015-12-14
fix journal resolution in article page route

richard-jones:
hotfix


Steven-Eardley:
yup. that's a bug. deploying.

Issue #921, opened 2015-11-05
Clear form entry field on hidden questions


dommitchell:
original issue: **Unable to save form**

Invokes 'Problem with submitted form error' but no indication on the form where or what the error is:

https://doaj.org/admin/suggestion/003cd6bc210344a09333c641dba95a47

Reported by @greboun

Steven-Eardley:
Well, the error is on question 31, which is hidden! Say 'yes' for the preceding question and you see "N/A" is in the text box. This is related to a fix I recently put in at @greboun's request, issue #902

Steven-Eardley:
That's quite tricky to find - we should perhaps clear that text box when the question is hidden, so there isn't a chance for nasty data to get in there.

dommitchell:
@greboun I'll leave this to you to look at first. We can discuss later.

greboun:
I think it would be best to clear the text box atomatically when the
Question is hidden like you suggest
Thanks for sorting this out

Dr. Tom Olijhoek
Editor-in-Chief  DOAJ
Support us: www.doaj.org/support

Steven-Eardley:
The fix is live now - you can check that hiding questions will clear the hidden text box.

dommitchell:
Great! Thank you!

PR #971, opened 2016-01-04
JavaScript changes to clear hidden fields on application form, preven…

Steven-Eardley:
…ting states where errors occur in hidden blocks and can't be easily rectified. Issue #921

Replaced a deprecated call in forms.py
Code style changes in forms,py so that pycharm would let me amend this commit message.

## Hotfix


emanuil-tolev:
:+1: Steve, deploying it now.

Issue #747, opened 2015-05-29
Javascript widget - controllable widget


emanuil-tolev:
No freetext box, but is programmatically controllable via Javascript (passing in parameters at least) to show embedded search results from the public API.

emanuil-tolev:
Potential source of high load. Some thoughts on the matter:

* * *

@dommitchell re: the load issue, it's actually quite hard to know that. I expect we can only reasonably see how the load is while our pilot friends are using the newly minted widget - that is to say without doing heavier-duty scalability testing, which can take quite some effort to set up. Currently my plan regarding load is:

see how pilot publishers "load" the site
from that, make some judgements on how to improve scalability (e.g. can we simply buy a third data server, can we segregate API requests somehow) if that needs to happen
let it go live if there are no major concerns, and keep monitoring the load - if there are concerns we'll discuss them during the project.
ultimately we can always shut the widgets down in emergencies. A widget means the code for it is actually downloaded from doaj.org and then that code also queries DOAJ - stop serving the widget, stop all the widget requests. Of course that'll break it on widget users' websites.
The alternative is to simulate the kind of load the widget will generate which is just too much effort for a widget. At the end of the day it's only a smallish part of the API project. Of course if load from it becomes a sticking point we'll have to discuss options of handling that - and don't forget that beyond what we can do, we can always simply ask those whose sites are too big for us to handle, to remove the widget temporarily while we work with you to decide if we want increased capability + exposure on the very high traffic site, or go without that exposure and expenditure. That way smaller widget users can keep using it and we deal specifically with problematic users, it's not an all or nothing.

Steven-Eardley:
1. first step - copy current widget

2. functionality:
take a JS variable DOAJ_WIDGET_QUERY = "string as per Discovery API docs"
take a JS variable DOAJ_WIDGET_DOCTYPE = journal|article
display a facetview after loading all necessary libraries - JS libraries can be loaded in the JS part of the widget (use current jQuery loading as example) and CSS can be loaded in the HTML part of the widget - the less CSS files the faster the display on other websites

3. design:
        * we don't know outright how much space we have - but we might not have to think about this if facetview design works OK on smaller spaces. If it doesn't work OK we will need a new FV theme "smallwidth" or something.
        + fixed width? 200px?
        + minimum width approach? i.e. assume we at least x, but allow it to expand. Recommend a width
    * go for fixed width first, then see if we can enhance it easily or allow steps of width (140, 200, 260 etc. BS spans)
[...]

PR #973, opened 2016-01-05
Fixed toc links on widget facetview,

Steven-Eardley:
env-ified the links to static resources in widget,
Added env detection to facetview theme so that the correct widget file version can be generated
#747 

## Hotfix

### DIFF
--- a/portality/static/doaj/js/doaj.facetview.theme.js
+++ b/portality/static/doaj/js/doaj.facetview.theme.js
@@ -493,15 +493,35 @@ function humanDate(datestr) {
 //////////////////////////////////////////////////////
 // fixed query widget generation
 /////////////////////////////////////////////////////
+
+var doajenvmap = {
+    "http://localhost:5004" : "dev",
+    "https://testdoaj.cottagelabs.com" : "test",
+    "https://stagingdoaj.cottagelabs.com" : "staging"
+};
+
+function doajDetectCurrentEnv(){
+    // Return env, if one of recognised locations, default to production.
+    return doajenvmap[document.location.origin] || "production";
+}
+
 function doajGenFixedQueryWidget(widget_fv_opts){
-    // Put the html code here which will set the options and embed the widget
+    // Generates code for the fixed query widget
+
     var source = elasticSearchQuery({"options" : widget_fv_opts, "include_facets" : widget_fv_opts.include_facets_in_url, "include_fields" : widget_fv_opts.include_fields_in_url});
 
     // Code to get a version of jQuery
     var jq = '<script type="text/javascript">!window.jQuery && document.write("<scr" + "ipt type=\\"text/javascript\\" src=\\"http://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js\\"></scr" + "ipt>"); </script>';
 
+    // Get the env to serve the correct version of the widget file
+    var env_suffix = "";
+    var doaj_env = doajDetectCurrentEnv();
+    if (doaj_env != "production") {
+        env_suffix = "_" + doaj_env;
+    }
+
     // Code to configure the widget
     var frag = '<script type="text/javascript">var doaj_url="'+ document.location.origin + '"; var SEARCH_CONFIGURED_OPTIONS=' + JSON.stringify(source) + '</script>';
-    frag += '<script src="' + document.location.origin +'/static/widget/fixed_query.js" type="text/javascript"></script><div id="doaj-fixed-query-widget"></div></div>';
+    frag += '<script src="' + document.location.origin +'/static/widget/fixed_query' + env_suffix + '.js" type="text/javascript"></script><div id="doaj-fixed-query-widget"></div></div>';
     return jq + frag
 }
--- a/portality/static/widget/fq_widget_depends_compiled.js
+++ b/portality/static/widget/fq_widget_depends_compiled.js
@@ -1026,12 +1026,12 @@ function dateHistogramShowLess(options, context, facet) {
 }
 
 var licenceMap = {
-    "CC BY" : ["https://doaj.org/static/doaj/images/cc/by.png", "http://creativecommons.org/licenses/by/3.0/"],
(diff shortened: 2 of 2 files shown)


Steven-Eardley:
@emanuil-tolev adding env stuff to facetview felt a bit hacky, but it's only as bad as hard-coding the URLs in the various widget files we have.

I had started this process of env-ing the widget, then decided it was easier / neater to fix it to the live site, but I appreciate you getting it to the point where I could finish it off - usability wise it's much better to have the data coming from the origin.

portality/static/widget/fq_widget_depends_compiled.js
```diff
@@ -1026,12 +1026,12 @@ function dateHistogramShowLess(options, context, facet) {
 }
 
 var licenceMap = {
-    "CC BY" : ["https://doaj.org/static/doaj/images/cc/by.png", "http://creativecommons.org/licenses/by/3.0/"],
-    "CC BY-NC" : ["https://doaj.org/static/doaj/images/cc/by-nc.png", "http://creativecommons.org/licenses/by-nc/3.0/"],
-    "CC BY-NC-ND" : ["https://doaj.org/static/doaj/images/cc/by-nc-nd.png", "http://creativecommons.org/licenses/by-nc-nd/3.0/"],
-    "CC BY-NC-SA" : ["https://doaj.org/static/doaj/images/cc/by-nc-sa.png", "http://creativecommons.org/licenses/by-nc-sa/3.0/"],
-    "CC BY-ND" : ["https://doaj.org/static/doaj/images/cc/by-nd.png", "http://creativecommons.org/licenses/by-nd/3.0/"],
-    "CC BY-SA" : ["https://doaj.org/static/doaj/images/cc/by-sa.png", "http://creativecommons.org/licenses/by-sa/3.0/"]
```
emanuil-tolev: Wouldn't this break it? This code executes on the host site, right, so if you say /some_asset.png isn't it going to try to find that on host-site.com/some_asset.png ?

portality/static/doaj/js/doaj.facetview.theme.js
```diff
@@ -493,15 +493,35 @@ function humanDate(datestr) {
 //////////////////////////////////////////////////////
 // fixed query widget generation
 /////////////////////////////////////////////////////
+
+var doajenvmap = {
```
emanuil-tolev: This seems alright, it's only in the facetview theme rather than the vendor library itself. That's partly the point of having a theme, so you can do custom things with it. You already had the fragment "// Code to configure the widget" below in the FV DOAJ theme.

portality/static/widget/fq_widget_depends_compiled.js
```diff
@@ -3269,372 +3269,3 @@ function journal_toc_id(journal) {
 
     return toc_id;
 }
-
```
emanuil-tolev: what's up with the big delete of fv call handling functions below, are they just not used by the widget's faceview config?

portality/static/widget/fq_widget_depends_compiled.js
Steven-Eardley: see line 3021 - I prepend the doaj_url - this is the same solution I have been using elsewhere for static resources.

I.e. the staging site widget will no longer request assets from the live site.

emanuil-tolev:
> I had started this process of env-ing the widget, then decided it was easier / neater to fix it to the live site, but I appreciate you getting it to the point where I could finish it off - usability wise it's much better to have the data coming from the origin.

I think I made some assumptions here which weren't true, as you had found while you worked on it. Sorry about that, should have given the issue more thought.

portality/static/widget/fq_widget_depends_compiled.js
emanuil-tolev: Cool, sounds reasonable then. I do have one question though ... do we also need to make these changes to the normal JS files besides the compiled file? The compiled file is really supposed to be just a reflection of the collection of our normal multi-file JS code, though I appreciate it's the thing that will actually get loaded for the widget. I'm apprehensive to let the separate-file-codebase and this big file diverge, or rather I give priority to the separate files since I can pretty much mechanically regenerate this.

portality/static/widget/fq_widget_depends_compiled.js
Steven-Eardley: yeah, I discovered a weird difference between the various facetviews in doaj. public.journalaricle.facetview.js defines the rendering function before facetview declaration, whereas most of the others e.g  /admin.*\.js/ use rendering callbacks for the components of the result objects. 

the widget copied the public facetview, so I inherited the inferior workflow, and the closures below aren't required.

portality/static/widget/fq_widget_depends_compiled.js
Steven-Eardley: That ship has pretty much sailed I'm afraid. There are numerous bodges in this compiled file...
We should not add these to the normal file - as a whole, this compiled js needs some variables that aren't relevant or present in other places facetview is expected. I didn't want to change how facetview is designed.

portality/static/doaj/js/doaj.facetview.theme.js
Steven-Eardley: good

Steven-Eardley:
it's ended up looking like a better piece of work, so I'm happy!

portality/static/widget/fq_widget_depends_compiled.js
emanuil-tolev: :+1:

portality/static/widget/fq_widget_depends_compiled.js
emanuil-tolev: Hmm, fair enough. Ok, we can compare if it comes to upgrading it. We need to get it out so people can use it, and if it does prove popular enough to warrant a redesign with the next redesign of the site we will have to compare the contents of this file, or simply rewrite it using the same process we used this time. Hopefully made a lot easier by just looking at changes to this file with git.

emanuil-tolev:
:+1: , merge and release then!

Issue #824, opened 2015-08-01
XML article ingestion issue with a few files


emanuil-tolev:
cc @richard-jones @Steven-Eardley 

Not sure if bug or not.

```
importing  /home/cloo/appdata/doaj/upload/7e767b74ff2b4bfc983b196db7d5e30a.xml
server error {u'status': 500, u'error': u'IllegalArgumentException[Document contains at least one immense term in field="bibjson.abstract.exact" (whose UTF8 encoding is longer than the
 max length 32766), all of which were skipped.  Please correct the analyzer to not produce such terms.  The prefix of the first immense term is: \'[60, 33, 45, 45, 91, 105, 102, 32, 10
3, 116, 101, 32, 109, 115, 111, 32, 57, 93, 62, 60, 120, 109, 108, 62, 32, 32, 60, 119, 58, 87]...\', original message: bytes can be at most 32766 in length; got 39625]; nested: MaxByt
esLengthExceededException[bytes can be at most 32766 in length; got 39625]; '}
```

```
importing  /home/cloo/appdata/doaj/upload/9d2ec13033e84ed0b0aaa4c04614be07.xml
bad request {u'status': 400, u'error': u'RemoteTransportException[[Nth Man: the Ultimate Ninja][inet[/10.131.168.23:9300]][indices:data/write/index]]; nested: IllegalArgumentException[
Document contains at least one immense term in field="bibjson.abstract.exact" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped.  Please correct the 
analyzer to not produce such terms.  The prefix of the first immense term is: \'[60, 33, 45, 45, 91, 105, 102, 32, 103, 116, 101, 32, 109, 115, 111, 32, 57, 93, 62, 60, 120, 109, 108, 
62, 32, 60, 111, 58, 68, 111]...\', original message: bytes can be at most 32766 in length; got 38141]; nested: MaxBytesLengthExceededException[bytes can be at most 32766 in length; go
t 38141]; '}
  File "/home/cloo/repl/apps/doaj/src/doaj/portality/scripts/ingestarticles.py", line 202, in <module>
    result = article.ingest_file(handle, format_name=upload.schema, owner=upload.owner, upload_id=upload.id)
  File "/home/cloo/repl/apps/doaj/src/doaj/portality/article.py", line 624, in ingest_file
    results = xwalk.crosswalk_doc(doc, article_callback=cb, limit_to_owner=owner, fail_callback=article_fail_callback)
[...]

PR #992, opened 2016-02-23
[hotfix] refactor article ingest script and DAO save and block methods

emanuil-tolev:
Related to #824 and general system maintenance.


richard-jones:
Ok, great, all merged and pushed, ready for running on live

Issue #982, opened 2016-02-02
Article metadata duplication: bibjson.subject


Steven-Eardley:
Each time `Article.add_journal_metadata()` is run, duplicate subjects are added. They are appended from the journal, rather than checking if they exist first. This has been confirmed in data from the live index.

To fix:
- code change to check for existence in `GenericBibJSON.add_subject()` before append
- write a migration to remove existing duplicates from `Article.bibjson.subject`

richard-jones:
This indicates that we need to introduce de-duplication into the bibjson model - putting under tnm, as although a bug is also 18 months old!

richard-jones:
@Steven-Eardley - this is resolved, right?  Can we close?

Issue #964, opened 2015-12-09
2619 articles return 500 Internal Server error


dommitchell:
http://bit.ly/1OT479U

These are all article of the license type 'CC by-nc' which is an old license type and shouldn't be on the site. Note that there are 2 CC BY-NC listing under license now:

CC BY-NC (350137)
CC by-nc (2619)

Any idea what is going on?

emanuil-tolev:
A bit of an idea. It'll crystallise soon enough. cc @richard-jones @Steven-Eardley did we not have to deal with a similar error on article pages before?

If not we'll investigate more thoroughly (though I think we have enough info to replicate). Below is the error log, and below that, the article's JSON record in Elasticsearch so we can replicate locally.

Looks like we might need to escalate that job which will refresh article metadata based on the journal info - this article has issns so it's reachable via a journal ToC and the discovery API, but has no "journal" section in the JSON. This is I believe planned as part of the Harvester project since it deals with refreshing article metadata automatically too. Nonetheless, to fix this we could create a separate script from the current portality/scripts/journalinfo.py which looks for articles without a "journal" section or which have a "journal" but no "journal.bibjson" (just in case) - and updates them. Let me know if you have any thoughts about doing it differently or if I'm wrong about the cause of this.

```
2015-12-09 13:52:28,556 - portality.core - ERROR - Exception on /article/000c38480c3d4d1383444cf8b24bc85d [GET]
Traceback (most recent call last):
  File "/home/cloo/repl/production/doaj/local/lib/python2.7/site-packages/flask/app.py", line 1687, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/cloo/repl/production/doaj/local/lib/python2.7/site-packages/flask/app.py", line 1360, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/cloo/repl/production/doaj/local/lib/python2.7/site-packages/flask/app.py", line 1358, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/cloo/repl/production/doaj/local/lib/python2.7/site-packages/flask/app.py", line 1344, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/cloo/repl/production/doaj/src/doaj/portality/view/doaj.py", line 278, in article_page
    return render_template('doaj/article.html', article=article, journal=journal)
[...]

Issue #941, opened 2015-11-16
Article cleanup script


richard-jones:
Building on the existing journalinfo.py script, we need a script which does the following for each article:
1. Checks that there is a corresponding journal
   - If there is not, the article should be deleted
2. Ensure that the article in_doaj status is the same as the journal's
3. Applies the journal's information to the article metadata as needed

In order for this to be run regularly it needs to be performant.  The following changes should be made to how the script works:
1. All journal records requested should be cached, so they are only looked up once
2. scroll search over articles (might already do this, but check)
3. Don't write articles that don't need writing (i.e. where they haven't changed)
4. Bulk write articles
5. Bulk delete articles

richard-jones:
In addition, we're going to need to add an extra field to the "journal" portion of the article model: "issn".

This is so that the journal can provide the article with the issn to use in generating the toc, in the case that the article also contains another issn which isn't known by the journal record.  In that latter case, this can cause the article to provide the incorrect toc url in the search interface.

When we've done that, we then need to modify the search interface to read the issn for the toc from the journal portion of the record.

Steven-Eardley:
@richard-jones how much does it matter that keys are unicode strings? Sometimes when ```Article.get_journal_metadata()``` is run, it replaces unicode keys ```[{u'type': u'CC BY', u'title': u'CC BY'}]``` with non-unicode keys from the Journal object ```[{'type': u'CC BY', 'title': u'CC BY'}]```

This is triggering my 'a change has been written' bit, although I think they are unicode when I next retrieve them, after saving the Article object. So the cycle continues.

I'm only seeing this within licenses from Journals, so far.

Steven-Eardley:
Related issue to this script: bug fix #982

richard-jones:
There's no strong need for the keys to be unicode - we don't use any special characters in the keys so they should be pretty much interchangeable between unicode and str.  So, if you want to force it to be one way or the other, that's good with me!

Steven-Eardley:
>When we've done that, we then need to modify the search interface to read the issn for the toc from the journal portion of the record.

Would it be sufficient to change the Article's ```index.issn``` to the new one from ```journal.issn``` when the index is built?

richard-jones:
I'm not sure.  Thoughts below...

We need the journal.issn field to link to a single known issn for the journal, that we guarantee will link successfully, but the index.issn contains all known issns for the article, for searching purposes.  That means, I think, that journal.issn is a string and index.issn is a list.

[...]

Issue #296, opened 2014-09-09
JournalInfo script to only take action if necessary


richard-jones:
So, when syncing journal info to articles, only update the article if the journal info changes.

Also, would be useful to check for articles whose parent journals can no longer be found, and warn about them.

richard-jones:
@Steven-Eardley - a related issue to your journal info update script.  I don't think there's anything new here, except the email to the feedback address.  I'm not sure we actually want to do that for every article - perhaps the script can produce a digest of orphaned articles, and log them or mail them somewhere at the end?

Steven-Eardley:
Issue superseded by #941, and the script checks orphaned articles, deletes them, and logs their IDs to file.

PR #993, opened 2016-02-28
Feature/iss941 article cleanup script

Steven-Eardley:
## CAUTION - this contains migrations which edit all articles

This is a combination of code additions, refactors and new scripts / migrations.

### Code changes:
* fixed cause of #982, where article.add_journal_metadata() was adding duplicate subjects
* refactor of article.py to separate the finding of related journal into its own function
* added property to articles: ```journal_issn```, which is the journal's preferred ISSN
* article.add_journal_metadata() now finds the ```journal_issn``` and can take a journal as an argument.
* new function in ```util.py``` to normalise dictionary keys as unicode (to better compare them)

### Migration:
* new migration ```migrate/subjects/remove_duplicate_subjects.py``` - remove duplicate subjects #982.
* documentation changes to existing migrations, and fixed a bug in ```upgrade.py```

### Scripts
* new script for regular use, ```/scripts/article_cleanup_sync.py```, which is a successor to the ```journalinfo``` script. This replaces and augments article's metadata by finding its associated journal, copying some properties, and (optionally) deleting those articles for which no journal could be found.

### Esprit
* requires new version which can handle bulk deletes.

#941, #296, #964


Steven-Eardley:
## ON DEPLOY

I strongly recommend trying this in test / staging first.

* run the ```remove_duplicate_subjects.py``` first (this can be read-only to check results)
* try ```article_cleanup_sync.py``` in read-only, then with ```-w -p``` flags, which writes changes (including deletes) and runs ```prep()``` on unchanged articles too to get updates to their index

Steven-Eardley:
@emanuil-tolev @richard-jones PR for your perusal.

portality/migrate/subjects/remove_duplicate_subjects.py
```diff
@@ -0,0 +1,85 @@
...
+        try:
+            article_model = models.Article(_source=a)
+
+            # Remove the duplicates
+            subjects = unicode_dict(article_model.bibjson().subjects())
+            newsub = []
+
+            for sub in subjects:
+                if sub not in newsub:
+                    newsub.append(sub)
```
emanuil-tolev: ```newsub = list(set(subjects))```?

emanuil-tolev: I can't think of why what you've done wouldn't work, I just know of list(set(orig_list)) as the goto for list deduplication in Python

Steven-Eardley: Subjects are non-hashable

portality/migrate/subjects/remove_duplicate_subjects.py
```diff
@@ -0,0 +1,85 @@
...
+                updated_count += 1
+                if write_changes:
+                    article_model.data['bibjson']['subject'] = newsub
+                    article_model.prep()
+                    write_batch.append(article_model.data)
+            else:
+                same_count += 1
+
+        except ValueError:
+            # Failed to create model (this shouldn't happen!)
```
emanuil-tolev: log the id of the article and keep a count of failed model init()s?

portality/migrate/subjects/remove_duplicate_subjects.py
Steven-Eardley: e.g. ```list(set([{'a' : 1, 'b': 2}, {'a' : 1, 'b': 2}]))```

portality/migrate/subjects/remove_duplicate_subjects.py
Steven-Eardley: will do.

portality/models/article.py
```diff
@@ -271,6 +292,9 @@ def _generate_index(self):
         issns += cbib.get_identifiers(cbib.P_ISSN)
         issns += cbib.get_identifiers(cbib.E_ISSN)
 
+        # get the issn from the journal bibjson
+        issns.append(cbib.journal_issn)
```
emanuil-tolev: how does this relate to the two lines above it? .get_identifiers() for PISSN and EISSN . Sounds like this would produce a list of duplicate ISSNs, e.g. if PISSN is 0123-4567 and there is no EISSN, then executing lines 292-296 (as per your new code) will result in an issns list of ['0123-4567', '0123-4567'] as the PISSN will be returned by both .get_identifiers(cbib.P_ISSN) and cbib.journal_issn. Perhaps another deduplication? Or removing lines 292 and 293 as they're potentially superseded by 296?

portality/scripts/article_cleanup_sync.py
```diff
@@ -0,0 +1,137 @@
...
+batch_size = 1000
+
+# Cache the Journals here, so we don't keep searching for them in the index
+journal_cache = {}
+
+
+def cleanup_articles(write_changes=False, prep_all=False):
+
+    # Connection to the ES index
+    conn = esprit.raw.make_connection(None, 'localhost', 9200, 'doaj')
```
emanuil-tolev: can you use the elasticsearch settings for host, port and index from app.config (portality/settings.py) instead? The live app server actually refers to "gateway:9200" when it talks to ES. The gateway machine does the round-robin of ES nodes this way.

portality/scripts/article_cleanup_sync.py
```diff
@@ -0,0 +1,137 @@
...
+            # Try to find journal in our cache
+            bibjson = article_model.bibjson()
+            pissns = bibjson.get_identifiers(bibjson.P_ISSN)
+            eissns = bibjson.get_identifiers(bibjson.E_ISSN)
+            allissns = list(set(pissns + eissns))
+
+            assoc_journal = None
+            for issn in allissns:
+                try:
+                    assoc_journal = journal_cache[issn]
```
emanuil-tolev: :+1: very nice

portality/scripts/article_cleanup_sync.py
```diff
@@ -0,0 +1,137 @@
...
+                except KeyError:
+                    continue
+
+            # Cache miss; ask the article model to try to find its journal
+            if assoc_journal is None:
+                assoc_journal = article_model.get_journal()
+
+                # Add the newly found journal to our cache
+                if assoc_journal is not None:
+                    for issn in allissns:
```
emanuil-tolev: Is this necessarily true? allissns is derived from the article record. Some of the issns in there could be out of date. Should we only record the journal (in the cache) against the issns present in the *journal* record we just fetched, rather than all issns of the article when we found the journal through any 1 of them?

portality/scripts/article_cleanup_sync.py
```diff
@@ -0,0 +1,137 @@
+"""
+For each article in the DOAJ index:
+    * Checks that it has a corresponding journal, deletes it otherwise (if -d set)
```
emanuil-tolev: There is no -d argument. -w writes and deletes, or no -w does not do anything.

portality/scripts/article_cleanup_sync.py
Steven-Eardley: Good point!

portality/upgrade.py
```diff
@@ -55,10 +55,7 @@ def do_upgrade(definition, verbose):
             tasks = tdef.get("tasks", None)
             if tasks:
                 for func_call, kwargs in tasks.iteritems():
-                    if args:
```
emanuil-tolev: can you tell what this used to refer to? I can't find an args variable, there's just kwargs...

emanuil-tolev: your change seems sensible, but it kind of looks like this should have been failing with "name args is not defined" - or else it did get used for something

portality/util.py
```diff
@@ -125,7 +125,18 @@ def normalise_issn(issn):
...
         content = f.read()
-    return content
\ No newline at end of file
+    return content
+
+
+def unicode_dict(d):
+    """ Recursively convert dictionary keys to unicode """
+    if isinstance(d, dict):
+        return dict((unicode(k), unicode_dict(v)) for k, v in d.items())
```
emanuil-tolev: there are dictionary comprehensions in Python nowadays as well, but I presume this is already-tested code so best stick to it :)

portality/migrate/subjects/remove_duplicate_subjects.py
emanuil-tolev: :+1:

portality/scripts/article_cleanup_sync.py
Steven-Eardley: oops, I did change that. I changed it to -w which handles all writes and deletes. I was wondering, however,  as I wrote the instructions above whether we might want to write updates but not do deletes. Should I put the -d flag back in?

portality/upgrade.py
Steven-Eardley: yup, it was something silly I did. It actually find args from the argparser below, so it does stuff so long as you have provided an argument...

portality/util.py
Steven-Eardley: It works for me!

portality/models/article.py
Steven-Eardley: There are now 3 ISSNs in each article record. PISSN and EISSN have been recorded in ```Article.bibjson()``` which written on the article's creation. These are the article's own ISSNs.

My new work has added ```journal_issn``` which is a copy of the preferred ISSN from the journal record, and added to the journal section of the article's bibjson. So, this is an additional way of identifying the article, via a link to its parent Journal.

My hope by appending it to the Article's index of ISSNs is to make it more likely to be findable by a Journal which is looking for its articles. I'm not completely convinced I've done enough for this to be the case.

portality/scripts/article_cleanup_sync.py
Steven-Eardley: I'm not sure I see what you mean - I'm caching the journal by the ISSNs the article has used to find it.

portality/scripts/article_cleanup_sync.py
emanuil-tolev: I can't foresee us wanting to keep orphaned articles, but for flexibility I would put it in. However, I would default to deleting if -w is specified.

It is going to be a little tricky to manage the interaction of the 2 flags and a potential source of bugs, so I'm wondering if it really is worth introducing. Don't put in the effort to introduce it for now - let's talk about it a bit tomorrow, and then it can be introduced if needed.

emanuil-tolev: I've put it in the agenda.

portality/scripts/article_cleanup_sync.py
emanuil-tolev: Yes, but will all the ISSNs that the article used to find the journal *actually belong* to that journal? (In edge cases where one of the ISSNs on the article record is wrong / old etc.)

Therefore should we only cache it by its own recorded ISSNs?

portality/upgrade.py
emanuil-tolev: :+1:

portality/util.py
emanuil-tolev: that's the spirit

portality/scripts/article_cleanup_sync.py
Steven-Eardley: Ah! possibly! @richard-jones ?

emanuil-tolev: also added to agenda for tomorrow

portality/models/article.py
Steven-Eardley: Also for discussion

portality/scripts/article_cleanup_sync.py
Steven-Eardley: Discussed - no change. done.

portality/scripts/article_cleanup_sync.py
Steven-Eardley: done

Steven-Eardley:
Thanks for the excellent code review @emanuil-tolev - I think it's ready now

portality/migrate/subjects/remove_duplicate_subjects.py
emanuil-tolev: Is this one pushed? I just see except ValueError: continue

It's ok if you've decided not to do that too, but since you've noted "done" I'm just wondering

portality/models/article.py
```diff
@@ -271,6 +292,13 @@ def _generate_index(self):
         issns += cbib.get_identifiers(cbib.P_ISSN)
         issns += cbib.get_identifiers(cbib.E_ISSN)
 
+        # get the issn from the journal bibjson
+        if type(cbib.journal_issns) is list:
```
emanuil-tolev: ```isinstance(cbib.journal_issns, list)``` is the canonical way to check (also includes child classes, whereas this line is doing an exact match, i.e. won't work for list-like-collections). See http://stackoverflow.com/a/152596/1154882 if you'd like a bit more background.

emanuil-tolev: Although yours does read very nicely.

portality/scripts/article_cleanup_sync.py
emanuil-tolev: So we decided not to put in a -d flag, yes, but you still need to remove the reference to it from the module docstring :)

emanuil-tolev:
Some very minor stuff left I think @Steven-Eardley

portality/migrate/subjects/remove_duplicate_subjects.py
Steven-Eardley: you should see this change! I can when I use the 'view full changes' link. I might not've done it for both scripts though, so will do so.

portality/models/article.py
Steven-Eardley: thanks for showing me that! ```basestring``` is useful to remember, too

portality/migrate/subjects/remove_duplicate_subjects.py
emanuil-tolev: I'm just looking at the latest diff https://github.com/DOAJ/doaj/pull/993/files line 44 of file portality/migrate/subjects/remove_duplicate_subjects.py . Our long discussion in between notwithstanding, lines 43-45 are just:

```
+except ValueError:
+            # Failed to create model (this shouldn't happen!)
+            continue
```

Issue #961, opened 2015-12-04
Sending email to wrong contact name


dommitchell:
There is a bug in the confirmation emails going out for reapplications.

/start/
Dear Fernando César Lima Leite,
Congratulations! Your reapplication for Revista Ibero-Americana de Ciência da Informação has been successful and the journal has been accepted back into DOAJ. Please go to the new journal record and review to check for accuracy. In particular, please check that the ISSN information is correct.
/end/

We don't know where Fernando is coming from because the name of the contact in the reapplication record is Murilo Bastos da Cunha
https://doaj.org/admin/journal/607fd93912b04b60bd23320ebf047ac2

Can you please verify where you are pulling the name field from and correct it so that it pulls in the name of the journal contact in Question 9 of the reapplication form?

Steven-Eardley:
@dommitchell the system is doing the right thing - see my email entitled 'doaj issue 961 - publisher name incorrect'. The account the journal is owned by is named Fernando; you could assign the journal to a different account for Murilo if you like.

dommitchell:
Thanks! Closing and opening new, separate issue

dommitchell:
Now I think about this: where are we pulling the name information from? We don't store that anywhere in the user database....

Steven-Eardley:
Of the 8074 accounts in DOAJ, 7308 have a name field. You can see the email I mentioned above for a sample of what we store for accounts. Under the normal workflow we create a new account when a journal is accepted, using the contact information from the application form. So we do store the names, but don't have an interface for changing it on existing accounts.

The situation reported in this issue has arisen because we intentionally send reapplication notifications to the *existing* account, rather than allowing a new account to take over an existing journal when they reapply.

dommitchell:
Yah, I am just wondering where those names came from originally. They must have been imported when we migrated in 2012/3. As you say, there is nowhere to change them in the current interface and there is also nowhere that they are displayed:  eg https://doaj.org/account/00015113

Steven-Eardley:
They continue to come from the application form; the journal contact fields. We just don't display them anywhere else, or allow anyone to change them!

emanuil-tolev:
This continues to be a problem as "journal accepted" emails are targeted at account.name, but there's no interface to update those names. We've been asked to amend some names already. It would be good to either:

- add an interface to change the names
- stop using account-level names in the journal acceptance process and just email whoever was on the application contact when the journal is accepted

emanuil-tolev:
[...]

PR #995, opened 2016-03-04
Hotfix/2016 03 04 acc name form

Steven-Eardley:
Part of the changes for #961 - allowing users (and admins) to change account's name.


Steven-Eardley:
Looks like this:

screen shot 2016-03-04 at 13 55 14

emanuil-tolev:
Does this work as-is then? No changes to the Python form objects to add .name field or whatever? Or is this just the 1st piece, like the UI.

Steven-Eardley:
It works as-is: see lines 57 and [58](https://github.com/DOAJ/doaj/blob/develop/portality/view/account.py#L58) in accounts.py: all keys from the POST (apart from the disallowed ones) are added to the data. This probably does mean I could add my own field and data to my account if I wanted!

I verified it working in my local index - adding text to that box does add a ```.name``` attribute, and the text is retrieved on subsequent loads. Plus, it saves properly and is visible in the index.

emanuil-tolev:
> It works as-is: see lines 57 and 58 in accounts.py: all keys from the POST (apart from the disallowed ones) are added to the data.

Hahah, what is this form handling .. it's very different to the other forms. Fair enough though!

Issue #749, opened 2015-05-29
Integration with Google Analytics


emanuil-tolev:
- [ ] Server-side integration with GA to log requests to all API endpoints.

- [ ] Log requests to the OAI endpoint too.

emanuil-tolev:
This doesn't fully fit into the phase 1 schedule but adding it so we can have a go if we've got time (I've got a bit). Also to note, if possible and easy it would be nice to send who the user making the request is to google analytics (as some sort of custom data) in addition to the usual anon info they use. This is obviously only going to work on authenticated routes.

emanuil-tolev:
When this is done also do one pass of nginx logs to determine how many requests we got against what API endpoints before instituting analytics.

PR #997, opened 2016-03-08
[hotfix] add analytics integration to API

emanuil-tolev:
For #749

### DIFF
--- a/portality/decorators.py
+++ b/portality/decorators.py
@@ -2,6 +2,7 @@ from functools import wraps
 from flask import request, abort, redirect, flash, url_for, render_template
 from flask.ext.login import current_user
 from flask_login import login_user
+import UniversalAnalytics
 
 from portality.api.v1.common import Api401Error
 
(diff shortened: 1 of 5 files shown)


portality/decorators.py
```diff
@@ -88,3 +89,64 @@ def decorated_view(*args, **kwargs):
...
+
+    Then we get a hit asking for application with id '12345'.
+    This will result in an event with category "API Hit", action "Retrieve
+    application" and label "12345".
+
+    We can also choose to not record custom data from the args passed
+    to the view function during a request. This might be appropriate
+    in cases such as creating an application.
+    """
+    # TODO extract these into env vars, but right now they're not env vars
```
emanuil-tolev: blocker - don't merge until this is done

portality/decorators.py
```diff
@@ -88,3 +89,64 @@ def decorated_view(*args, **kwargs):
         return fn(*args, **kwargs)
 
     return decorated_view
+
+def track_analytics(event_category, event_action, record_value_of_which_arg=''):
```
emanuil-tolev: Consider some tests for this, considering how important it is and how involved the abstract decorator code is.

emanuil-tolev:
@Steven-Eardley @richard-jones this is updated now to make decorator more generic (for evtl octopus integration) and factor out config var, making it safe to use in dev/test envs without danger of polluting live analytics data.

emanuil-tolev:
poke :). upon merging this will go to Test, where I will try and configure it to log analytics specifically for the test site. If that goes well it'll be released and we'll see the real effects.

Issue #1003, opened 2016-04-05
More facetview icon issues


Steven-Eardley:
We override the facetview2 theme's render function for terms filters so that we can provide a blue fontawesome close icon; the same needs to be done with a histogram filter.

From the (current live) public search page, notice the cross is white for 'year':
<img width="664" alt="screen shot 2016-04-05 at 16 48 55" src="https://cloud.githubusercontent.com/assets/1446355/14288446/a7bf08d2-fb4e-11e5-97e9-87256de8b72f.png">

Steven-Eardley:
Will fix along with Toc work #983 which also uses histograms filters.

Steven-Eardley:
Fixed, deployed.

PR #1004, opened 2016-04-05
Prerequisites for the facetview2 ToC are here so that they can be in …

Steven-Eardley:
…place before the UI changes, meaning it will seamlessly release when that happens.

These are invisble changes which can safely be released before the fv toc.

Bonus: fixed issue with white icon on histogram filters in all facetviews #1003


Steven-Eardley:
@emanuil-tolev @richard-jones fyi

emanuil-tolev:
:+1:

PR #1011, opened 2016-04-21
add script which removes applications which are in status 'reapplicat…

richard-jones:
…ion' and withdraws the corresponding journal/articles from DOAJ

This script needs to be run on a recent clone of the live data into test, before we can deploy it to live.

Therefore, recommend not to merge this until after (or at the same time as) the continuations PR.

To run the script, we need to run the following:

    python portality/scripts/reapp_remove_pending.py -o <output file>


Steven-Eardley:
oops, deployed. Don't run this until comments above addressed.

PR #1018, opened 2016-05-13
upgrade facetview version to fix bug with escapeHtml

richard-jones:
Another minor fix for the continuations work.  There is a small bug with facetview2 that results in an error when displaying some records.  I can't reproduce this locally, so it must just be a difference in the data, but I know what the problem is and this should fix it.  It might throw up other problems when we push it to test, but we'll handle those when they come up.


emanuil-tolev:
This will get deployed automatically in ~10 min from now, 14:50 BST.

richard-jones:
Amazing, thank you!

On 13 May 2016 at 14:40, Emanuil Tolev <notifications@github.com> wrote:

> This will get deployed automatically in ~10 min from now, 14:50 BST.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/DOAJ/doaj/pull/1018#issuecomment-219045832>
>



-- 

Richard Jones,

Founder, Cottage Labs
w: http://cottagelabs.com

PR #1020, opened 2016-05-16
update fv2 for more bugfixes

richard-jones:
This upgrade for fv2 fixes the latest issue on the test server, which is just another render error on a field which contains an empty value.


emanuil-tolev:
should be up at ~17:10 BST

PR #1026, opened 2016-05-21
add filter to public search results to strip out unwanted fields

richard-jones:
hotfix

### DIFF
--- a/portality/settings.py
+++ b/portality/settings.py
@@ -207,7 +207,7 @@ DEFAULT_SORT = {
 }
 
 QUERY_ROUTE = {
-    "query" : {"role": None, "default_filter": True},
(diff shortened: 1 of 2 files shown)


richard-jones:
Incidentally, all unit tests pass locally, so something wrong with the codeship here, I think.

emanuil-tolev:
:+1: but can't roll out today myself I'm afraid

Steven-Eardley:
I can deploy it today, and I'll check the build on codeship shortly.

Steven-Eardley:
Codeship was failing due to SSH keys being denied. Maybe they've changed them. Anyway, I've set it to authenticate through my account rather than emanuil's and the builds are working again.

Issue #1033, opened 2016-05-25
Article overwriting not working again: causes duplication


dommitchell:
http://bit.ly/1OIORdZ
http://bit.ly/1sMN7fI

I asked the publisher to correct some broken characters and the new versions didn't overwrite the old version as it should have. Can you investigate?

richard-jones:
I've had a quick poke around the tests to make sure that url and doi matching are working, and everything seems to be in order.

Can you give me a bit more detail on what the publisher was doing?  Is this a bulk XML upload, or an individual article metadata entry, or API request?

Is there any chance they can send us the source XML data, if it was an upload?

dommitchell:
User account 1110662x, and they used XML upload feature. Files emailed to you.

richard-jones:
Ok, I've found out what's going on and provided a hotfix.

The issue was that for accounts with >1024 issns associated with them, deduplication was failing due to a technical error with the deuplication query.

@dommitchell - while we get this deployed, the advice in the mean time is for all very large accounts to avoid uploading updates

emanuil-tolev:
The hotfix is deployed. Not 100% sure how this is to be reviewed.. have one of the affected accounts upload (a small number of articles) again and see if no dupes occur?

dommitchell:
Asking HIndawi...

dommitchell:
Hindawi confirms that this has worked

dommitchell:
Leaving this here as I htink that this has become an issue again: http://bit.ly/2aGmhzG

These 10 articles are identical to their 10 predecessors.

Steven-Eardley:
It seems there are two identical versions which were created at different times on the same day. 

https://doaj.org/admin/admin_site_search?source=%7B%22query%22%3A%7B%22filtered%22%3A%7B%22filter%22%3A%7B%22bool%22%3A%7B%22must%22%3A%5B%7B%22term%22%3A%7B%22_type%22%3A%22article%22%7D%7D%5D%7D%7D%2C%22query%22%3A%7B%22query_string%22%3A%7B%22query%22%3A%222466%5C%5C-3913%22%2C%22default_operator%22%3A%22AND%22%7D%7D%7D%7D%2C%22sort%22%3A%5B%7B%22index.unpunctitle.exact%22%3A%7B%22order%22%3A%22desc%22%7D%7D%5D%2C%22from%22%3A0%2C%22size%22%3A10%7D

e.g.

```
"hits": [
{
"_index": "doaj",
[...]

PR #1037, opened 2016-06-10
make Article.duplicate break up the query into several queries if too…

richard-jones:
… many issns are provided

HOTFIX

This fixes the duplication problem seen in #1033 - the problem was that for accounts with > 1024 ISSNs associated with them, the ES query was throwing an error due to a size limit.

### DIFF
--- a/portality/models/article.py
+++ b/portality/models/article.py
@@ -3,6 +3,7 @@ from portality.models import GenericBibJSON, Journal
 from copy import deepcopy
 from datetime import datetime
 from portality import xwalk
+from portality.core import app
 
 import string
 from unidecode import unidecode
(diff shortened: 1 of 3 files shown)


portality/models/article.py
```diff
@@ -17,20 +18,47 @@ def duplicates(cls, issns=None, publisher_record_id=None, doi=None, fulltexts=No
...
+        terms_limit = app.config.get("ES_TERMS_LIMIT", 1024)
+        issn_groups = []
+        lower = 0
+        upper = terms_limit
+        while lower < len(issns):
+            issn_groups.append(issns[lower:upper])
+            lower = upper
+            upper = lower + terms_limit
+
+        if issns is not None and len(issns) > 0:
```
emanuil-tolev: len(issns) > 0 should be enough, there's a type check and enforcement on line 18 above.

emanuil-tolev: ah forget it, the rest of the PR is fine, so merging and deploying

emanuil-tolev:
:+1:

Issue #1039, opened 2016-06-13
Inaccuracy in OAi-PMH


dommitchell:
Please asses this one for tnm. Would this be an easy thing to enhance the accuracy of? What are our options?

From a user:
/start/
I noticed that article records in Dublin Core OAI Article format (OAI_DC) include the dc:rights element. According to the documentation this is “the journal's content re-use policy. Will be one of the Creative Commons licences”.

Many OA journals offer authors a choice of Creative Commons licenses. This means that there can be discrepancies between an article’s actual re-use rights and the re-use rights described in the OAI-PMH record. For example, this Elsevier article is published under a CC BY license but the dc:rights information provided by DOAJ lists it as CC BY-NC-ND:
http://www.sciencedirect.com/science/article/pii/S2212096315000261
https://doaj.org/oai.article?verb=GetRecord&identifier=oai:doaj.org/article:3273ed9d74864a279f580c450a4700d6&metadataPrefix=oai_dc

Within the article record it is not made clear that this rights element applies at a journal level, and not necessarily to the article itself, which could cause problems for people using these records – especially if the DOAJ record lists an article as having a less restrictive rights policy than it actually does. Are you aware of this, and do you have any plans to improve the accuracy of this element, or plans to remove this element in the future? From a publisher’s perspective it’s concerning that there may be inaccurate rights information associated with article records. The current DOAJ upload XML schema does not allow for article-level rights information to be provided.

It is worth noting that records provided in OAI_DOAJ format do not contain an equivalent element.
/ends/

emanuil-tolev:
Previously DOAJ has not held article-level licence information. I suppose that's the major decision to make rather than a technical challenge - either start accepting and holding such information, or make it clear the current field is journal-level only.

dommitchell:
Right! That's a bigger issue

> or make it clear the current field is journal-level only.

How could we do that in the context of the feed? Just through documentation? Not actually in the feed itself?

emanuil-tolev:
> How could we do that in the context of the feed?

Good question. cc @richard-jones who's written and ingested a lot of OAI feeds.

richard-jones:
Consider moving the rights information to provenance: http://dublincore.org/documents/dcmi-terms/#terms-provenance

richard-jones:
Done in upcoming PR

richard-jones:
This one, with the associated documentation, is now on the test server @dommitchell

dommitchell:
[...]

Issue #786, opened 2015-07-01
OAI earliest date question


dommitchell:
/start/
The following request:
https://doaj.org/oai?verb=Identify
executed today, returns:
<earliestDatestamp>2015-06-10T15:10:50Z</earliestDatestamp>

However, the following request, which I'd expect to return all records in the database:

https://doaj.org/oai?verb=Listrecords&metadataPrefix=oai_dc&from=2015-06-10T15:10:50Z

only returns completeListSize="289", which clearly is far less than the expected 10000+ journals!

On the other hand, without date range:

https://doaj.org/oai?verb=Listrecords&metadataPrefix=oai_dc

returns completeListSize="10344", which seems more plausible.

I suspect something similar is wrong with oai.articles. I hope you can fix this, as our OAI harvester relies on this earliest date.
/end/

richard-jones:
Certainly used to work, so I expect that something else we changed for another project has had a knock-on effect, so this will just require an update.

richard-jones:
The exact reason for this has eluded me - it seems that there's some kind of issue with how dates are handled in statistical and terms facets which causes them to be different from the values actually stored, with regard to the time zone.  Basically, the earliest date was being reported as an hour later than the actual earliest date, causing a mismatch in the results.

I've altered the query to get the dates reliably, and that's in an upcoming PR

dommitchell:
I don't know how to text these. @richard-jones will test

dommitchell:
Seen it on the call. Closing.

PR #1059, opened 2016-07-29
Feature/oai fixes

richard-jones:
A couple of OAI fixes for tnm: #1039 and #786

### DIFF
--- a/portality/models/oaipmh.py
+++ b/portality/models/oaipmh.py
@@ -11,15 +11,10 @@ class OAIPMHRecord(object):
                 ]
             }
         },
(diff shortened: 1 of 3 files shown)


emanuil-tolev:
:+1: . This will go on the test server. Do tell when you'd like it released to live.

emanuil-tolev:
Actually, one question I remembered too late: can we add any tests for the problems described in the two issues this is fixing?

emanuil-tolev:
Should this be released now? Seems like the kind of thing that should've been a hotfix (it's an OAI fix after all).

PR #1065, opened 2016-08-15
Feature/autocomplete

richard-jones:
This adds a mechanism to switch between autocomplete algorithms depending on the incoming string, and supports filtering by one field (e.g. in the index.* document space) and faceting on another, allowing us to more effectively use the prefix query, and to switch seamlessly to the wildcard query.

### DIFF
--- a/portality/migrate/autocomplete/autocomplete.json
+++ b/portality/migrate/autocomplete/autocomplete.json
@@ -1,0 +1,7 @@
+{
+	"types": [
+		{
+			"type": "journal"
+		}
+	]
+}
--- a/doajtest/unit/test_models.py
+++ b/doajtest/unit/test_models.py
@@ -874,15 +874,15 @@ class TestClient(DoajTestCase):
 
         time.sleep(2)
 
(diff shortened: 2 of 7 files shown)


richard-jones:
Oh, in order to deploy this the migrate script needs to be run. See

    migrate/autocomplete/README.md

for details

Steven-Eardley:
Looks good to me.

portality/migrate/autocomplete/autocomplete.json
```diff
@@ -0,0 +1,7 @@
+{
+	"types": [
+		{
+			"type": "journal"
```
emanuil-tolev: Don't applications (type suggestion) also have these auto-complete dropdowns and therefore require a self._generate_autocompletes() to be called on them as well?

Bit unusual to comment on a merged PR - just want to make sure we've done everything consistently.

emanuil-tolev:
:+1: - I had it on my todo to come back to review this properly when I have time just because Linus' law.

PR #1067, opened 2016-08-15
Feature/no editor

richard-jones:
Adds data model extension and facets for browsing by whether journals and applications have editors/editor groups

In order to deploy, migrate must be run, see

    migrate/no_editor/README.md


Steven-Eardley:
Rebased onto more recent develop to fix conflict. Note:  I will **run this migration only** because this subsumes the one in #1065

PR #1070, opened 2016-08-16
enhance article import logic to reject articles where one of the issn…

richard-jones:
…s does not match a journal which already has 2 issns


Steven-Eardley:
Yep, makes sense.

Issue #1068, opened 2016-08-16
Delete 10008 Hindawi articles


dommitchell:
Duplicates caused by issue #1033 https://github.com/DOAJ/doaj/issues/1033#issuecomment-225295757

Can you delete these for us?

dommitchell:
Email me for the CSV file.

dommitchell:
Notes from call: email list to Richard.

richard-jones:
The code to do this has now been merged, and deployed on test.  We should probably run it on test to ensure that it works.  I have the list of article ids to remove, so @emanuil-tolev if I send these over to you, can you just execute the script and check it functions.  The command is:

    python portality/scripts/articlerm.py -c <path to csv>

emanuil-tolev:
Steve executed this on test. @richard-jones if you are/were happy with the results, the script is now on live. If you send the IDs over I'll run this on live.

dommitchell:
Can this one be updated please?

Steven-Eardley:
@dommitchell this is still assigned to me - can we close? (I hope so, this was over a year ago!)

dommitchell:
I never got confirmation that this had actually been done. Was it done?

Steven-Eardley:
It was not done. I imagine It was assigned to me after a conversation not logged here so that I'd run it rather than emanuil as noted above.

Steven-Eardley:
It should still be possible using the original script and CSV file:

[ArticleIDs.csv.txt](https://github.com/DOAJ/doaj/files/1507046/ArticleIDs.csv.txt)

Steven-Eardley:
It's worth noting, these are going to be deleted by https://github.com/DOAJ/doaj/issues/1296 eventually anyway. When we eventually run it in production, that is.

dommitchell:
So perhaps not worth duplicating effort then? If you are certain that #1296 will catch them all then I am happy for this to be closed.

Steven-Eardley:
I'm satisfied it will work for them - I've checked a couple of the IDs in the test server's log for the dedupe script, and they feature. Actually, it may be the better course of action - these IDs in the CSV are the more recent version with better metadata (abstract vs no abstract), the script would delete the older versions instead and keep these.

They trigger both DOI and full text, since they are genuine duplicates:
```
[...]

PR #1074, opened 2016-08-23
enhance articlerm script to also take a csv/text list of article ids …

richard-jones:
…to delete

When this is merged, I can supply a csv of the article ids that we need to remove for issue #1068


Steven-Eardley:
Looks good

Issue #962, opened 2015-12-07
Reapplication accepted email should go to journal contact listed in reapplication record.


dommitchell:
Please change it so that when a reapplication is accepted, the confirmation email goes to the Journal Contact listed in the accepted reapplication record (Questions 9 and 10) and *not* the contact in the account or the original application.

richard-jones:
I've been taking a look at this, and it is actually surprisingly fiddly. Or rather, getting the right user experience is.  Here's the current setup:

1. A "suggester" creates a new application, and they provide a "contact"
2. When the application is accepted, the "suggester" gets a user account created for them (if one didn't already exist) and an account activation email is sent
3. Then a Congratulations email is sent to the "suggester", referencing the account creation email, and giving them maintenance instructions

If we switch to sending email (3) to the "contact" rather than the "suggester", then the content of the email will be wrong.  There are a couple of options, that I can see:

1. Send an email to the "contact" just informing them.  Also send the maintenance instructions to the "suggester"
2. Create an account for the "contact" instead or as well, and send them the maintenance instructions instead.

I think probably (1) is the right way to go.  There are a couple of email templates that matter:

https://github.com/DOAJ/doaj/blob/develop/portality/templates/email/publisher_application_accepted.txt
https://github.com/DOAJ/doaj/blob/develop/portality/templates/email/publisher_reapplication_accepted.txt

@dommitchell - how do you want to approach it?

dommitchell:
I agree. (1) is the way to go. I can edit the letter templates, no problem.

richard-jones:
Ok, I've implemented this in my feature branch so that now on acceptance an email is sent to both the owner and the contact.

@dommitchell - Before I create a PR, do you think you could do me the text for each of the 4 email templates?  You can tweak them later, but I'd like to get them as close to correct as possible for the initial version, so that when we merge it with the live version it's immediately in the right state.

I need text for the following situations:

* Owner's email, when it's an application
* Owner's email, when it's a reapplication
* Contact's email, when it's an application
* Contact's email, when it's a reapplication

You can use the existing application email as a template:

https://github.com/DOAJ/doaj/blob/develop/portality/templates/email/publisher_application_accepted.txt

You'll have access to all the same variables (publisher_name, etc)

dommitchell:
Leaving the texts here. Please let me know if they will work.

[Reapplication_owner_email.txt](https://github.com/DOAJ/doaj/files/426888/Reapplication_owner_email.txt)
[...]

PR #1076, opened 2016-08-28
Feature/journal contact email

richard-jones:
For issue #962

On application accepted, 2 emails are now sent: one to the journal contact and one to the account owner.

### DIFF
--- a/portality/formcontext/formcontext.py
+++ b/portality/formcontext/formcontext.py
@@ -465,6 +465,40 @@ class ApplicationContext(PrivateContext):
             app.logger.error(magic + "\n" + repr(e))
             raise e
 
+    def _send_contact_approved_email(self, journal_title, publisher_name, email, reapplication=False):
+        """Email the journal contact when an application is accepted """
+        url_root = request.url_root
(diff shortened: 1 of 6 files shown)


portality/formcontext/formcontext.py
```diff
@@ -582,7 +618,13 @@ def finalise(self):
             # create the user account for the owner and send the notification email
             try:
                 owner = self._create_account_on_suggestion_approval(self.target, j)
-                self._send_application_approved_email(j.bibjson().title, owner.name, owner.email, self.source.current_journal is not None)
+                journal_contacts = ""
+                for contact in j.contacts():
+                    names = contact.get("name")
+                    journal_contacts = ", ".join(names)
```
emanuil-tolev: Do you mean `journal_contacts` += `", ".join(names)`? It appears here that the latest contact's name will be the only one shown, as the variable is being overwritten in this loop. So the last iteration of the loop will be all that remains.

richard-jones:
Ok, I've added a fix for the name concatenation problem

Issue #996, opened 2016-03-07
Add collapsible groups of DOAJ members to /members static page


emanuil-tolev:
Consortia are already listed on https://doaj.org/members . If help is needed to identify the exact list of consortia if you can't tell by the name (there should be 15 in total), ask @jbarnsby .

Consortia are listed under their respective (main) country where they (chiefly) operate. These are and will continue to be listed by DOAJ admins who edit the /members page. The idea here is to add the ability to collapse/expand a list each consortium's members.

For example, currently we have:

#### Russia
* NEICON - National Electronic Information Consortium

We want to end up with

#### Russia
* <a href="#expand">[+]</a> <a href="http://neicon.ru/" target="_blank">NEICON - National Electronic Information Consortium</a>

The [+] is a link which expands / collapses the list of members. It might be decorated with a dotted underline to distinguish it from link following immediately after - the name of the consortium, which will link to its website (as it does right now).

Ideally this will be done as unobtrusively as possible with JS. I would consider simply adding a class like "expandable" to the ```<li>``` element containing the consortium name, and adding the [+] entirely via Javascript. That way, if JS doesn't work for some reason, you don't get lots of inoperative [+] links. This also makes it relatively easy for DOAJ admins to operate.

I'd recommend the following HTML structure for listing the consortium members (just bc I had to check guidelines on nested lists):

```
<ul class="the usual country-level list listing various organisations, including our consortium(s) in the country">
<li class="expandable the-element-for-1-consortium">
        <span class="expand-collapse inserted-via-JS">+</span> <a href="consortium website">Consortium 1</a>
        <ul class="one_consortiums_members">
            <li>Poppleton Library (these could also be links to individual websites, DOAJ admins know how to add/edit them)</li>
            <li>Poppleton University Press</li>
        </ul>
</li>
```

It'd be good to get one example actually, so get 1 consortium and a list of its members. When development is finished this will become the first example consortium which DOAJ admins like @jbarnsby can copy to other consortia.

emanuil-tolev:
This has been agreed to proceed under T&M for a few/several hours of work. We need an example (I've asked for one) and probably a stricter estimate just to help guide the work internally to CL, but besides that this should be ready to be picked up.

jbarnsby:
Consortium example (listed under USA):
Virtual Library of Virginia (VIVA)

Members:
Bridgewater College
Christopher Newport University
College of William & Mary
Eastern Shore Community College
Eastern Virginia Medical School
George Mason University
[...]

PR #1077, opened 2016-08-28
add javascript for automatically collapsing and making interactive li…

richard-jones:
…st elements which are double nested

For issue #996


emanuil-tolev:
Clean, short and sweet. Would only need extending if there are nested lists which do not indicate a consortium (so `<ul><super-li that is not a consortium><ul><nested li><nested li 2> ... <nested li n></ul></li></ul>`).  At the moment it binds "is a nested list" to "is a consortium".

I don't believe this is a problem at the moment, but let's see. I'll merge this into develop, at which point the page at https://testdoaj.cottagelabs.com/members should change. Here's the page as it is before the merge: https://web.archive.org/web/20160829095226/https://testdoaj.cottagelabs.com/members , so we see if any lists are collapsed when they should not be (presumably not).

Issue #1045, opened 2016-06-23
Add distinction for Reapplication records. Facet?


dommitchell:
We need a new facet in admin to distinguish between reapplications and applications. 

Priority please.

emanuil-tolev:
@richard-jones @Steven-Eardley maybe based on whether there is a linked journal record present? Don't recall if "reapplication" is directly recorded as an attribute anywhere. Sounds like it might have to be made one. index.reapplication: false | true .

emanuil-tolev:
Any reasons *not* to do this, or to not do it in a particular way (e.g. a new field in the index part of each application record)?

emanuil-tolev:
@dommitchell @greboun are we happy with an "Application type" facet with currently 2 values: "reapplication" and "new application"?

dommitchell:
@emanuil-tolev Yep!

emanuil-tolev:
This should be available on live now.

emanuil-tolev:
(and on test)

dommitchell:
Smashing!

dommitchell:
Can we have it added to the Editor view as well? 

https://doaj.org/editor/group_applications?source=%7B%22query%22%3A%7B%22match_all%22%3A%7B%7D%7D%2C%22from%22%3A0%2C%22size%22%3A10%7D

emanuil-tolev:
Should be live.

dommitchell:
Perfect!

PR #1078, opened 2016-08-28
add reapplication facet to editor view

richard-jones:
HOTFIX

Adds the reapplication facet to the editor view, as per #1045


emanuil-tolev:
:+1: I will merge a bit later and deploy

Issue #1061, opened 2016-08-03
Please create outputs for the following queries


dommitchell:
All journals with the status 'Rejected'. Should include: Title, ISSN, Date applied, Last updated date and, if possible, the Notes for the following countries or group of countries:
1) India

2) Latin America:
Argentina
Bolivia
Brazil
Chile
Colombia
Costa Rica
Cuba
Dominican Republic
Ecuador
El Salvador
French Guiana
Guadeloupe
Guatemala
Haiti
Honduras
Martinique
Mexico
Nicaragua
Panama
Paraguay
Peru
Puerto Rico
Saint Barthélemy
Saint Martin
Uruguay
Venezuela

3) Middle East & Persian Gulf:
Egypt
Jordan 
Lebanon
Syria
Iraq
Iran
Saudi Arabia
Yemen
Oman
Kuwait

4) Russian Federation

5) East Africa:
Kenya
Tanzania
Ethiopia
Somalia
South Sudan
Sudan
Nigeria  (English speaking!)
Uganda

6) North Africa
Morocco
Algeria
Tunisia
Libya

7) West Africa:
Senegal
Gambia
Guinea Bissau
Cape Verde
Guinea
Mauretania
Mali
Niger
Chad
Burkina Faso
Ivory coast
Liberia
Ghana
Togo
Cameroon
Central African Republic
Congo
Rwanda
Burundi
Equatorial Guinea
Gabon

8) Southern Africa:
South Africa
Namibia
Zimbabwe
Malawi
Zambia
Angola
Mozambique
Botswana
Madagascar

richard-jones:
Sorry, I'm not totally sure I understand what you're after here.  Can you give me a bit more detail?

dommitchell:
I'd like a list of all rejected journals from those countries, including title, issn, date rejected and, if possible, notes. There should be 8 files, 1 for each region

dommitchell:
UPDATE: produce a list of rejected journals as CSV export.

richard-jones:
Working on this, I suddenly realised we might be talking at cross-purposes.

Which of the following are you looking for:

1. Journals which are in_doaj=False
2. Applications which have status "rejected"

dommitchell:
> Applications which have status "rejected"

dommitchell:
Done!

dommitchell:
Not done. Closed too soon. But sample that @richard-jones sent me is approved

richard-jones:
A PR has been made for this

[...]

PR #1083, opened 2016-09-01
enable querying suggestions by status, and producing a csv report on …

richard-jones:
…that

For issue #1061 

Once this has been merged, can it be run on test just to make sure it functions there, and then deployed to live, and then run again to produce the final output to be sent to @dommitchell 

To run the script do

    python portality/scripts/rejected_reapplications.py -o <output file path>


emanuil-tolev:
> portality/scripts/rejected_reapplications.py

Filename is actually rejected **app**lications (not *reapp*lications), fyi for myself or Steve - there's a typo.

emanuil-tolev:
I'll send file by email so you can confirm it looks ok.

PR #1096, opened 2016-10-11
Code improvements to reporting: renamed a few variables which were co…

Steven-Eardley:
…lliding with builtins / imports, caught an exception when the role added to provenance record is not an editor role.


richard-jones:
+1

I haven't merged it yet since the checks are still running, but it looks fine to merge to me

Issue #1085, opened 2016-09-13
Journal CSV: missing information


dommitchell:
Column J in the journal CSV (Journal article processing charges (APCs)) only shows APC Yes. 'No' and 'No information' are missing.

https://doaj.org/csv

Can you fix?

dommitchell:
A fix to produce 'No information' was done on the facet so this just needs to be made the same in the CSV too.

dommitchell:
Can I have an update on this one please?

Steven-Eardley:
I've made a fix this afternoon, should be released soon, once reviewed.

dommitchell:
@Steven-Eardley  Not seeing that in test. Where can I review?

Steven-Eardley:
@dommitchell sorry I meant code review - @emanuil-tolev @richard-jones there's a PR up for this.

emanuil-tolev:
@dommitchell I think it's up on the test site now, could you try it? https://testdoaj.cottagelabs.com/csv

dommitchell:
Looks great! Can go to live.

Steven-Eardley:
Released now, it should re-generate soon and the changes will be visible on live.

Steven-Eardley:
Confirmed on live, closing.

PR #1106, opened 2016-10-21
Utilise the yes/no/no info APC charge data placed in the index for th…

Steven-Eardley:
…e journalcsv. For #1085.


emanuil-tolev:
If you merge without deploy, Dom could check it in test. I'm sure it will be fine and it can be rolled out live later today.

Issue #1100, opened 2016-10-14
Add concept of a "system" user, which is reserved


emanuil-tolev:
"system" can't be created as a name via any user creation mechanism:
- web UI
- script
- ?

PR #1108, opened 2016-10-25
Add reserved system user

emanuil-tolev:
#1100 

Plus a tiny bit of cleanup - some default args were empty lists (no-no), and a few unused imports.


emanuil-tolev:
It can't be added - that's why the validator is reused in the model. The script will exit with a validation exception when it tries to initialise the model. (I tried it.)

PR #1109, opened 2016-10-25
add missing tests and add enhancements to Create new user page

emanuil-tolev:
#1100 

Oops, apologies!! We shouldn't be reducing the coverage on the codebase, but increasing it!

I forgot to commit this file to the earlier PR and didn't see.

### DIFF
--- a/doajtest/unit/test_reserved_usernames.py
+++ b/doajtest/unit/test_reserved_usernames.py
@@ -1,0 +1,40 @@
+from doajtest.helpers import DoajTestCase
+
+from nose.tools import assert_raises
+
+from werkzeug.datastructures import MultiDict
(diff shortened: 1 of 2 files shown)


emanuil-tolev:
Further to the missing tests, I've fixed the "create new user" page:

image


Also creating a new user with another username still works via that page. No new further tests to add - my tests use a fake form, so apply equally to both application and create new user forms. As a consequence it would be nice if either of you can think of any *more* forms/pages which allow user creation!

emanuil-tolev:
Tried to double check code for any uses of "username", "owner" and similar, found none. I'm sure we'll hear about it soon enough if a user creation mechanism starts 500-ing without a nice error.

Issue #972, opened 2016-01-05
Hotfixes to application form


greboun:
ON HOLD UNTIL REAPPLICATION SHUTDOWN

We would like to see the following changes as listed in #137 prioritized:
as per our discussion on 1/5/2016 'remove' can be replaced by 'Hide from public view'
- Question 47: remove the 'No' option
- Question 51: remove OAKlist. OAKlist has been decommissioned.
-  Question 52: remove 'Other'.
-  Question 54: remove 'Other'.

One more question
could the admin /editor view of the application form be easily edited to highlight the questions that determine acceptance / rejection into DOAJ?    eg **text bold faced**  or **numbering in bold**

richard-jones:
When doing this we need to bear in mind that some applications may have these values set, so we need a migration strategy, or a legacy data strategy.

Propose that we first check to see how many records will be affected, then make a decision what to do.  If those records are rejected, for example, we might just want to delete them.

@greboun with regard to your question - yes, this would be possible - we'd provide custom descriptions to the fields that were relevant.  If you want us to go ahead with that also under tnm, could you open a new issue, and list the questions you want to appear in bold.  This doesn't have to wait until the reapplication shutdown, so we can do that now, independently of the above.

Assigning this to you, for review - just unassign yourself when you have decided what to do about the above.

emanuil-tolev:
> When doing this we need to bear in mind that some applications may have these values set, so we need a migration strategy, or a legacy data strategy.



> Question 47: remove the 'No' option

There are 4,154 applications which have selected "No". A number (72) are Accepted, so we do need to think what to do here. Maybe set some kind of text in the "Other" box after removing the "No" option.



> Question 51: remove OAKlist. OAKlist has been decommissioned.

OK, shall we just remove the OAKlist answer from all 132 who have answered this? In addition, if it was their only answer, we will set their Deposit Policy answer to "None" (tech note, do this by deleting .deposit_policy completely - I checked that this is what the form does on setting the None answer).

> Question 52: remove 'Other'.
> Question 54: remove 'Other'.

OK, again, shall we just drop the values that people have put in here over time? And set the answers to "No" (since it wasn't a Yes to begin with). These are required questions so we need to set something, otherwise the application objects become incomplete and you won't be able to save them (even if editing unrelated information, or trying to reject them).

emanuil-tolev:
> Question 47: remove the 'No' option
[...]

PR #1118, opened 2016-10-31
[hotfix] hotfixes for application form

emanuil-tolev:
# hotfix, do not merge here

# don't actually deploy this until it's tested on testdoaj.cottagelabs.com. See issue #972


emanuil-tolev:
Merged into develop so it can deploy to test. Leaving this PR open until the hotfix is also merged into master.

In the meantime, other hotfixes can continue being merged into master and develop as usual, so we're not blocked from doing deployments.

PR #1128, opened 2016-11-08
DO NOT MERGE Feature/1120 to 1123 background processor

richard-jones:
I have now rebased this and extended my code to incorporate huey.  We should look to merge this down to develop quite soon, as it will be necessary to develop the integration with the UI.


emanuil-tolev:
Build fail? May need some more setup now we have huey, I will check. Also title still says do not merge - I assume we are actually looking to merge now.

doajtest/unit/test_background.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def test_01_with_user_success(self):
+        acc = models.Account()
+        acc.set_id("testuser")
+        acc.save()
+        time.sleep(2)
+
+        task = BackgroundFixtureFactory.get_task(acc.id)
+        BackgroundApi.execute(task)
+        time.sleep(2)
```
emanuil-tolev: So this is just to let ES update. Since it's difficult to specify a blocking save here (you would have to change what BackgroundApi.execute does just for the benefit of a faster unit test) it's OK to leave this line as-is. Where we do use `.save()` explicitly in tests we should probably make it `.save(differentiate=True)`.

emanuil-tolev: Don't have blocking saves yet (new PR available for that), ignore this comment.

emanuil-tolev:
Just build fail now, otherwise :+1: from me.

Issue #1130, opened 2016-11-18
Review DOAJ bot guidelines


emanuil-tolev:
1. Does the sitemap look reasonable? doaj.org/sitemap.xml
2. Create a robots.txt - didn't have time to think of one back in 2013 and now JS-enabled spiders are wreaking havoc on the performance of the system. We want them to index content, but /search is totally unnecessary.
3. Fix https://doaj.org/toc - it's a non-javascript list of all journals in DOAJ, but the links appear to have gotten broken during a previous update. Write an automated test to ensure a further code change doesn't cause the same problem.

@ClaraDOAJ @greboun This is related to the DOAJ scalability quote and us trying to get a better handle on spiders and crawlers using the site. I would say it's a T&M task, but it should not take long (no more than 2-3 hours) and I imagine it would be pretty high priority work, especially if it helps us understand future scalability better. Correct me if I'm wrong here though.

PR #1131, opened 2016-11-18
[hotfix] modify sitemap, add robots.txt and fix /toc + test perf. improvements

emanuil-tolev:
#1130

### DIFF
--- a/portality/view/doaj.py
+++ b/portality/view/doaj.py
@@ -163,7 +163,7 @@ def autocomplete(doc_type, field_name):
 
 @blueprint.route("/toc")
 def list_journals():
-    js = models.Journal.all_in_doaj(page_size=1000, minified=True)
+    js = models.Journal.all_in_doaj(page_size=1000)
     return render_template("doaj/journals.html", journals=js)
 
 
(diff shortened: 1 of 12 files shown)


emanuil-tolev:
@Steven-Eardley would you have a minute to review this this week? Richard's probably going to be working on other things. Hopefully it's not too bad.

Steven-Eardley:
Will do, tomorrow morning I expect

portality/view/doaj.py
```diff
@@ -163,7 +163,7 @@ def autocomplete(doc_type, field_name):
 
 @blueprint.route("/toc")
 def list_journals():
-    js = models.Journal.all_in_doaj(page_size=1000, minified=True)
```
Steven-Eardley: Why the demise of minified? Apart from the old ```all_in_doaj``` code being confusing?

emanuil-tolev: Two things for me:

1. it was not being used anywhere else
2. it did not return model objects - just JSON ones with a few values (the point of minified). We actually need to use a method in the template serving this endpoint, `.toc_id()`, so in fact we do end up needing the whole object. I just didn't see any point in it after that - but I may have missed something!

Steven-Eardley: I think I'm with you there, thanks

PR #1137, opened 2016-11-27
Feature/1136 sitemap

richard-jones:
This converts the sitemap to be a scheduled task.

One thing to check:  I've commented out the anacron invocation of the old script, so the new job will need to run on the same schedule.  At the moment I've set it monthly, as I wasn't sure how the anacron config worked.

### DIFF
--- a/deploy/anacrontab-production-apps
+++ b/deploy/anacrontab-production-apps
@@ -10,5 +10,5 @@ PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 7       10      cron.weekly      nice run-parts --report /etc/cron.weekly
 @monthly        15      cron.monthly nice run-parts --report /etc/cron.monthly
 
-1       5     doaj-sitemap    su - cloo -c 'export NOW=`date +\%F_\%H\%M\%S` ; export DOAJENV=production ; echo -e "\n\nstart timestamp $NOW" >> /home/cloo/cron-logs/doaj-sitemap.log && nice -n 19 /home/cloo/repl/production/doaj/bin/python /home/cloo/repl/production/doaj/src/doaj/portality/scripts/sitemap.py >> /home/cloo/cron-logs/doaj-sitemap.log 2>&1'
+#1       5     doaj-sitemap    su - cloo -c 'export NOW=`date +\%F_\%H\%M\%S` ; export DOAJENV=production ; echo -e "\n\nstart timestamp $NOW" >> /home/cloo/cron-logs/doaj-sitemap.log && nice -n 19 /home/cloo/repl/production/doaj/bin/python /home/cloo/repl/production/doaj/src/doaj/portality/scripts/sitemap.py >> /home/cloo/cron-logs/doaj-sitemap.log 2>&1'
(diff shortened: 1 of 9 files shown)


emanuil-tolev:
> One thing to check: I've commented out the anacron invocation of the old script, so the new job will need to run on the same schedule. At the moment I've set it monthly, as I wasn't sure how the anacron config worked.

Anacron runs daily. The sitemap's regenerated every day.

deploy/anacrontab-production-apps
```diff
@@ -10,5 +10,5 @@ PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 7       10      cron.weekly      nice run-parts --report /etc/cron.weekly
 @monthly        15      cron.monthly nice run-parts --report /etc/cron.monthly
 
-1       5     doaj-sitemap    su - cloo -c 'export NOW=`date +\%F_\%H\%M\%S` ; export DOAJENV=production ; echo -e "\n\nstart timestamp $NOW" >> /home/cloo/cron-logs/doaj-sitemap.log && nice -n 19 /home/cloo/repl/production/doaj/bin/python /home/cloo/repl/production/doaj/src/doaj/portality/scripts/sitemap.py >> /home/cloo/cron-logs/doaj-sitemap.log 2>&1'
+#1       5     doaj-sitemap    su - cloo -c 'export NOW=`date +\%F_\%H\%M\%S` ; export DOAJENV=production ; echo -e "\n\nstart timestamp $NOW" >> /home/cloo/cron-logs/doaj-sitemap.log && nice -n 19 /home/cloo/repl/production/doaj/bin/python /home/cloo/repl/production/doaj/src/doaj/portality/scripts/sitemap.py >> /home/cloo/cron-logs/doaj-sitemap.log 2>&1'
```
emanuil-tolev: just delete that line, that's what git is for :)

deploy/huey.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+huey_consumer.py portality.tasks.consumer.main_queue
```
emanuil-tolev: I wonder if we can reuse this across environments, so in production sidekiq will start this script, which will start huey. That way we all start huey the same way. I can foresee there might be trouble with that .. but we'll see.

For now this is fine.

portality/tasks/sitemap.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+        Submit the specified BackgroundJob to the background queue
+
+        :param background_job: the BackgroundJob instance
+        :return:
+        """
+        background_job.save()
+        generate_sitemap.schedule(args=(background_job.id,), delay=10)
+
+@main_queue.periodic_task(crontab(month="*", day="1", hour="0", minute="0"))
```
emanuil-tolev: I'd say day="*" is needed here.

Separately, if you'd like to keep it doing exactly what it was doing before, you'd have to set hour="8" as well, but you don't have to - I don't have a strong opinion about when to run it during 1 day.

richard-jones:
ok, I've added a "schedule" function to the huey_redis.py file, which takes an action name, and returns a crontab for the configured time.

portality/tasks/redis_huey.py
```diff
@@ -1,4 +1,13 @@
-from huey import RedisHuey
+from huey import RedisHuey, crontab
 from portality.core import app
 
 main_queue = RedisHuey('doaj', host=app.config['HUEY_REDIS_HOST'], port=app.config['HUEY_REDIS_PORT'])
+
+def schedule(action):
+    cfg = app.config.get("HUEY_SCHEDULE", {})
```
emanuil-tolev: If I do schedule('lalalala') it's going to return * * * * * and run every minute. I think this is going to be a bit misleading and could result in unintentional runs (e.g. a typo of "shcedule_sitemap" means it will now run every minute rather than every day"). Could you make this function actually fail if the requested schedule does not exist?

So simply on lines 7 and 8

cfg = app.config['HUEY_SCHEDULE']  # you're not going anywhere without any configs if you're calling this function
action_cfg = cfg[action]  # the action you've passed as an arg to the schedule() function has to have a matching schedule, else fail

emanuil-tolev: All good otherwise though.

PR #1142, opened 2016-12-07
[hotfix] static pages update

emanuil-tolev:


emanuil-tolev:
just fyi

emanuil-tolev:
so we don't have any code merges into production and develop done solely by 1 person, not visible to others or in github

PR #1150, opened 2016-12-19
Feature/1138 articleupload

richard-jones:
Review version of the article ingest code.  This moves most of the article ingest functionality into the task and handles the possible asynchronous file system.

There is a set of tests now for all this functionality, but it will be good to test this on the dev server like a real user would use it.

### DIFF
--- a/portality/view/publisher.py
+++ b/portality/view/publisher.py
@@ -9,7 +9,7 @@ from portality import models, article
 from portality.view.forms import ArticleForm
 from portality.formcontext import formcontext
 from portality.util import flash_with_url
-from portality.tasks.ingestarticles import process_file_upload
+from portality.tasks.ingestarticles import IngestArticlesBackgroundTask, BackgroundException
 
 import os, requests, ftplib
 from urlparse import urlparse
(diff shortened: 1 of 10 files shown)


portality/view/publisher.py
```diff
@@ -92,16 +93,26 @@ def upload_file():
...
     
     if url is not None and url != "":
-        return _url_upload(url, schema, previous)
+        flash("File reference successfully received - it will be processed shortly", "success")
+        return render_template('publisher/uploadmetadata.html', previous=previous)
     
     flash("No file or URL provided", "error")
     return render_template('publisher/uploadmetadata.html', previous=previous)
 
+"""
```
emanuil-tolev: Is this code supposed to be commented out? It is actually modified below in a couple of places but the whole function is inactive.

richard-jones: Yes, it's not needed, just left it in for reference at the time, it can now be removed.

emanuil-tolev:
Noted that we need the s3fs set up on test ASAP to finish this one now.

emanuil-tolev:
Also I need the huey always eager code for my tests - I have actually copied it over into my branch, but can now do a merge from develop, so thanks for that.

PR #1151, opened 2016-12-19
Feature/background reporting

richard-jones:
Refactoring to the reporting background task.  As this was the first background task we ported, it was slightly behind our current conventions, so that is now fixed, and tests are added.


emanuil-tolev:
Conflict with develop, probably on huey testing setup

PR #1153, opened 2016-12-20
Feature/1152 jobs facetview

richard-jones:
Adds the UI for browsing background jobs

### DIFF
--- a/portality/static/doaj/js/available_facetviews/admin.background_jobs.facetview.js
+++ b/portality/static/doaj/js/available_facetviews/admin.background_jobs.facetview.js
@@ -1,11 +1,106 @@
 jQuery(document).ready(function($) {
 
+    function toggleMoreInformation(options, context) {
+        // toggle the abstracts
+        $('.more_information', context).hide();
+        $(".more_information_action", context).unbind("click").click(function(event) {
+            event.preventDefault();
+            var el = $(this);
+            $('.more_information[data-id="' + el.attr("data-id") + '"]').slideToggle(300);
+            return true;
+        });
+    }
+
+    function backgroundJobResult(options, resultobj) {
+
+        var firstRow = "";
+        if (resultobj.action) {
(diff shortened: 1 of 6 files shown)


Steven-Eardley:
At the moment, the IDs list goes a long way off to the side, and the selection_query hasn't been parsed:


screen shot 2016-12-28 at 11 23 29

portality/static/doaj/js/available_facetviews/admin.background_jobs.facetview.js
```diff
@@ -45,6 +45,9 @@ jQuery(document).ready(function($) {
             paramsBlock += "<strong>Parameters:</strong><br>";
             for (var key in resultobj.params) {
                 var val = resultobj.params[key];
+                if (val instanceof Array) {
```
richard-jones: I've never used instanceof before, so I'm not sure of the compatibility of it.  There is a $.isArray function in jquery, which makes me suspicious of instanceof.  Anyone know?  A quick google hasn't turned up anything obviously pro or against this approach.

portality/static/doaj/js/available_facetviews/admin.background_jobs.facetview.js
```diff
@@ -53,7 +56,7 @@ jQuery(document).ready(function($) {
         if (resultobj.reference) {
             paramsBlock += "<strong>Reference:</strong><br>";
             for (var key in resultobj.reference) {
-                var val = resultobj.reference[key];
+                var val = JSON.stringify(resultobj.reference[key]);
```
richard-jones: This puts quote marks around the value when it's a string, which might be confusing, so I may add a catch for arrays and objects and only use JSON.stringify in those cases.

richard-jones:
I've merged this for convenience in other bits of the project.  We can add any further fixes to it through other branches.

emanuil-tolev:
looks good anyway

Issue #1071, opened 2016-08-19
Please enable Owner in the Admin view for Applications


dommitchell:
https://doaj.org/admin/applications?source=%7B%22query%22%3A%7B%22match_all%22%3A%7B%7D%7D%2C%22from%22%3A0%2C%22size%22%3A10%7D

For all views.

emanuil-tolev:
Enable as in...? Be able to facet on that, add it to each search result, or be able to search inside that specific field?

Or all of them :)?

dommitchell:
Ooops. Just want that information to be visible in the search results display. No need to facet it, or be able to search on it. Just want it displayed, when that info is present,

Steven-Eardley:
The changes to the admin applications facetview are live now.

richard-jones:
@ClaraDOAJ - could you take a look at this one and close if you are happy it's complete

ClaraDOAJ:
How can I look at the applications as owner? Sorry- I don't know how to do it.

richard-jones:
Looking back at the notes above, it looks like what Dom wanted was just the Owner field to appear (linked) in the search results.  If you look at the application search in the admin area, you should be able to see applications which have been assigned to accounts have that data:

https://doaj.org/admin/applications?source=%7B%22query%22%3A%7B%22match_all%22%3A%7B%7D%7D%2C%22from%22%3A0%2C%22size%22%3A10%7D

Looks like it's done, then, but just check it's what you need.

ClaraDOAJ:
Understood. It can be closed! Many thanks!

Clara Armengou
DOAJ - Community Manager

PR #1155, opened 2017-01-03
Removed a deprecated field (confirmed absent from real data) from app…

Steven-Eardley:
…lications facetviews. Added a result field for the application owner in the admin facetview. Bumped version, because this is a JS update. #1071 

# HOTFIX


Steven-Eardley:
Also, please note it is 2017, despite what I have indicated.

Issue #1089, opened 2016-09-29
Bulk change of URLs from http to https


dommitchell:
A publisher has asked if we can change all their URLs from http to https across all their entries.

The publisher is Hindawi so we're talking about > 500 journals.

The URLs all currently start with:

http://www.hindawi.com/journals/.....

and should be changed to https://www.hindawi.com/journals/....

Can you do this automatically for us?

dommitchell:
Technically doable. Bit time-consuming probably. Probably have to write a migrate script. 

DM to provide a list of all the ISSNs or a search criteria.

emanuil-tolev:
Yes, could do under TnM. Have to consider articles, journals and (re?)applications.

dommitchell:
> Have to consider articles, journals and (re?)applications.

Only applies to journals and (re)applications.

Steven-Eardley:
I'm looking to take this on using a migrate script - can you supply the ISSNs or searches which cover the journals and applications from Hindawi?

ClaraDOAJ:
Hi @Steven-Eardley, I could. It is 564 entries. I have the search but I am not sure you are able to see it if I paste it here.

https://doaj.org/admin/journals?source=%7B%22query%22%3A%7B%22filtered%22%3A%7B%22filter%22%3A%7B%22bool%22%3A%7B%22must%22%3A%5B%7B%22term%22%3A%7B%22admin.in_doaj%22%3A%22T%22%7D%7D%5D%7D%7D%2C%22query%22%3A%7B%22query_string%22%3A%7B%22query%22%3A%22hindawi%22%2C%22default_operator%22%3A%22AND%22%7D%7D%7D%7D%2C%22from%22%3A0%2C%22size%22%3A564%7D

Steven-Eardley:
Thanks @ClaraDOAJ - When checking over the search you provided, I found they're not all owned by the same account. I'll send an updated search by email as it contains non-public information.

We can take our time and get things right, I can proceed in the meantime on the assumption we'll have search criteria to use, and write the script accordingly.

Steven-Eardley:
@ClaraDOAJ This has been run on the test machine now. Can you give its effects on the Hindawi journals a quick check before I run on live?

ClaraDOAJ:
Hi @Steven-Eardley I see that there are some entries that do not have the https. See second entry. 
[...]

PR #1156, opened 2017-01-06
Hotfix/2017 01 04 iss1089 edit by query

Steven-Eardley:
#1089 - there are lots of URLs in these objects, this script runs over the JSON dumped objects, uses a regex to change ```http://www.hindawi``` to ```https://www.hindawi``` wherever it occurs, then bulk saves them. It can be run on journals or suggestions if supplied with the correct query.

It's pretty low-level, so there's some danger if this is used incorrectly. A safer but slower approach would be editing all the URLs via the formcontext.

Also, it could be handy to write the counterpart to this; roll back records to a previous version based on a query.


portality/migrate/1089_edit_field_by_query/regex_http_to_https.py
```diff
@@ -0,0 +1,67 @@
...
+        exit(1)
+
+    if not args.type:
+        print 'One or more type parameters are required.'
+        exit(1)
+
+    print 'Starting {0}.'.format(datetime.now())
+
+    # Connection to the ES index
+    conn = raw.make_connection(None, 'localhost', 9200, 'doaj')
```
emanuil-tolev: This won't work on production I'm afraid, it overrides the ELASTICSEARCH_HOST variable to use a different hostname, "gateway". It's best if the scripts respect the configuration variables from settings anyway.

Esprit should have support for the all-in-one Elasticsearch connection url (i.e. with schema, host and port all in one string) that DOAJ defines. This, or something close to it, should work:

```python
conn = esprit.raw.Connection(app.config.get("ELASTIC_SEARCH_HOST"), app.config.get("ELASTIC_SEARCH_DB"))
```

emanuil-tolev: That is in fact the line used by a recent migration, `portality/migrate/972_appl_form_changes/appl_form_changes.py:13`, so it should work as-is.

emanuil-tolev:
When merging, it will be best to merge the hotfix branch into develop first, without merging it into production. Then we can test at our leisure on testdoaj.cottagelabs.com, and only fully merge it into production with git flow when we are about to deploy it.

richard-jones:
Yup, as long as we can test this in the test environment first, all looks fine to me

PR #1157, opened 2017-01-08
Bulk journal and application editing

emanuil-tolev:
This PR covers bulk editing for journals and applications, and tests. There are also bits of refactoring done to various parts of the system to allow cleaner bulk editing code.

The one outstanding problem is with editing applications that contain certain data, details to follow.

It's finished now, with a refactor of back-end bulk edit controllers taking place. A number of further bugs were discovered and fixed.


emanuil-tolev:
Here is a job I just ran locally:

suggestion_bulk_edit by emanuil status: complete
Job ID: f60f71a6f9294990aeb51cf184402873
Job Created: 8 Jan 2017 at 14:34
Job Last Updated: 8 Jan 2017 at 14:34

More Information
Parameters:
suggestion_bulk_edit__editor_group -- 
suggestion_bulk_edit__note -- 
suggestion_bulk_edit__ids -- d60860741575475cbff4a1c6312a0fd6, b747d0d3fe62460e9018a5990becdc52
suggestion_bulk_edit__application_status -- on hold
Reference:
selection_query -- {"query": {"filtered": {"filter": {"bool": {"must": [{"term": {"admin.application_status.exact": "pending"}}]}}, "query": {"match_all": {}}}}}
Audit Messages:
2017-01-08T14:34:37.813956Z -- Job Started
2017-01-08T14:34:37.822381Z -- Setting application_status to for suggestion d60860741575475cbff4a1c6312a0fd6
2017-01-08T14:34:37.823271Z -- Data validation failed while bulk editing suggestion d60860741575475cbff4a1c6312a0fd6 : {"processing_charges_currency": ["Not a valid choice"], "submission_charges_currency": ["Not a valid choice"]} The data from the fields with the errors is: {"processing_charges_currency": "None", "submission_charges_currency": "None"}
2017-01-08T14:34:37.849981Z -- Setting application_status to for suggestion b747d0d3fe62460e9018a5990becdc52
2017-01-08T14:34:37.851699Z -- Data validation failed while bulk editing suggestion b747d0d3fe62460e9018a5990becdc52 : {"processing_charges_currency": ["Not a valid choice"], "submission_charges_currency": ["Not a valid choice"]} The data from the fields with the errors is: {"processing_charges_currency": "None", "submission_charges_currency": "None"}
2017-01-08T14:34:37.852399Z -- Job Run Completed
2017-01-08T14:34:37.852782Z -- Cleanup Started
2017-01-08T14:34:37.871545Z -- Job Cleanup Completed
2017-01-08T14:34:37.872724Z -- Job Finished

It indicates a form validation failure (i.e. formcontext.validate() returning False) for two applications it was asked to set a new application_status on.

The trouble is, I can open these records as a managing editor and save them without any errors. I.e. I open, say, http://localhost:5004/admin/suggestion/b747d0d3fe62460e9018a5990becdc52 (the second ID). I can hit "Save" without editing anything and it saves successfully.

We have a situation where the form obtained from the formcontext can't be saved, but it saves fine in the UI.

The fields the error message refers to are hidden in the UI form because they are optional for those particular applications. They are only required if the answers to "does this journal have processing charges?" and "does this journal submission charges?" are "Yes". If the answers are "No", then the problem fields should be optional, and their values should not be validated.

This validation skipping seems to happen when the form is submitted via the UI, but not when it's submitted via by bulk application editing task.

I copied the Formcontext invocations from the UI form processing code, so not sure why this is coming about. It's probably to do with the use of Formcontext in an unusual way - we have to pretend the user has opened the form, changed it and is now trying to save it.

https://github.com/DOAJ/doaj/pull/1157/files?diff=unified#diff-2d3a9d72695e7d22551697540adf70ebR73 : line 73 to line 99 of this file contain the use of Formcontext.

The next step is probably to write a test that reproduces the behaviour described above (i.e. set processing_charges to "No" but set processing_charges_currency to "None" [string], same for submission_charges).

richard-jones:
Ok, so this is kind of good in a way, insomuch as it means we need to make sure that the code for the UI and the code for the background task work through the same interface.  We just need to debug that interface and tighten it up so that we get the same behaviour in both cases.

emanuil-tolev:
We've done this and the commit has been pushed up.

I need to finish some refactoring of the back-end controller, then this will be ready for a final review - tomorrow afternoon hopefully.

emanuil-tolev:
@richard-jones sorry I haven't pushed this up yet, I will do so tomorrow. Ran into some difficulties with refactoring the admin.py bits we've written. But I've formed a solution which is better than what's in the PR right now, it's just a question of changing the JS a bit to use the new URLs I've defined and testing it manually (since we've not got JS tests).

emanuil-tolev:
PR now finished. A refactor of back-end controllers for bulk editing was done for easier maintenance and to prepare for bulk delete ability (coming separately). A number of further bugs were also discovered and fixed.

richard-jones:
Ok, this looks fine to me.  I'm happy just to merge this and then continue development, and fix any issues that come up as we go on.  Shall I merge?

Issue #1129, opened 2016-11-14
Set up S3 File server and S3 Fuse on test


emanuil-tolev:
+ instructions to wire up any dev machine to S3 test

emanuil-tolev:
Some possible issues with this I've been researching last week.

I've selected https://github.com/s3fs-fuse/s3fs-fuse out of several alternatives due to POSIX compatibility (so no leaky abstractions - you can do to it what you can do to any other local file).

But ... the limitations section:

> Generally S3 cannot offer the same performance or semantics as a local file system. More specifically:

> - random writes or appends to files require rewriting the entire file

Don't care, as long as it takes care of it automatically.

> - metadata operations such as listing directories have poor performance due to network latency

Only very occasionally would we have to keep this in mind. (Large directory listings or an operation done often - **if done *every* time a task runs, could be an issue.**)

> - eventual consistency can temporarily yield stale data

I have to think about this a bit more (and have spent most of the time doing exactly that). That's the final big concern. @richard-jones can you paste the photo of the planned workflow that we drew on your whiteboard into this issue?

> - no atomic renames of files or directories

If we ever implement logic by rename, this might be an issue. (E.g. rely on renaming to result in the old file definitely not being there, or the new file being there immediately.)

> - no coordination between multiple clients mounting the same bucket

Hmmmm. Sounds like writing to the same file would be an issue (conflict) - not clear what the resolution strategy is. Of course we don't plan to do this.

There might be other effects, but we can only know if we try, I've not been able to find further info on the exact possible effects, short of reading the s3fuse source code.

> - no hard links

Fine.

richard-jones:
![2016-11-16 09 12 55](https://cloud.githubusercontent.com/assets/856138/20341167/e7eeee42-abdc-11e6-99cd-b6688b4f16b2.jpg)

richard-jones:
I'm not sure why the orientation is wrong, sorry!

richard-jones:
Also, that's creepy that the photo on the wall behind is just visible over the top!

emanuil-tolev:
Thanks Richard.

A little further info from the s3fs man page:

* * *

Due to S3's "eventual consistency" limitations, file creation can and will occasionally fail. Even after a successful create, subsequent reads can fail for an indeterminate
       time, even after one or more successful reads. Create and read enough files and you will eventually encounter this failure. This is not a flaw in s3fs and it is  not  some‐
[...]

PR #1161, opened 2017-01-09
[hotfix] set up S3FS

emanuil-tolev:
# hotfix, merge it on your command line

#1129 

After this is merged you'll need to run a few things on your dev environments. This time I wrote the documentation for this functionality directly on the wiki, as I should... https://github.com/DOAJ/doaj/wiki/Using-S3FS

EDIT: One more thing - the reason this is a hotfix is because:

- it's a bit more of a faff to only roll it out to test, since the scripts being modified are actually shared between test and production environments. And at the moment there's nothing specific to only one environment in either of them, so I didn't want to introduce that.
- there's no reason to wait - this will not affect live operations. At most it will fail to mount S3FS on production, in which case I'll have to fix it.
- I have already created the production-specific bits and prepared the config on the production app server.

Nonetheless, it'd be better to deploy this to test first, see if it works, fix it (heh), and only then deploy it live.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,8 @@ history/article/*
 history/journal/*
 !history/journal/.gitkeep
 
+.s3fs-credentials
+
 ### OSX template
 .DS_Store
 .AppleDouble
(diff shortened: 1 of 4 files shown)


richard-jones:
I don't have sufficient knowledge of s3fs to really know if this works, but it looks fine on superficial inspection.  Let's just get it out to test on the test machine and find out.

Do we need this as a hotfix, rather than as part of the background jobs project?

emanuil-tolev:
> Do we need this as a hotfix, rather than as part of the background jobs project?

It is proposed as a hotfix at the moment. The reasons are at the bottom of the first post.

PR #1164, opened 2017-01-10
Feature/1154 set in doaj

richard-jones:
This generalises the SetInDAOJBackgroundTask to support multiple journal ids, and adds back-end support for triggering the job by query (e.g. from the facetview).

This does not integrate with the bulk edit interface, that needs to come next.


richard-jones:
In order to attach to the bulk editing interface, #1157 needs to be merged first, then this can be integrated

richard-jones:
No need to try to merge this yet, I'm going to rebase with develop and do the integration with the bulk first.

portality/tasks/journal_in_out_doaj.py
```diff
@@ -24,23 +34,23 @@ def run(self):
         job = self.background_job
         params = job.params
 
-        journal_id = params.get("set_in_doaj__journal_id")
+        journal_ids = params.get("set_in_doaj__journal_ids")
```
emanuil-tolev: Use cls.get_param('journal_ids')

portality/tasks/journal_in_out_doaj.py
```diff
@@ -24,23 +34,23 @@ def run(self):
...
         in_doaj = params.get("set_in_doaj__in_doaj")
 
-        if journal_id is None or in_doaj is None:
+        if journal_ids is None or len(journal_ids) == 0 or in_doaj is None:
```
emanuil-tolev: If journal_ids is actually none as opposed to [], this line will crash with `TypeError: object of type 'NoneType' has no len()`. Perhaps simply "if journal_ids or in_doaj is None"? (you can't do the same to in_doaj since it could legitimately be False, but a list should only ever be [] or have some items)

portality/tasks/journal_in_out_doaj.py
```diff
@@ -50,10 +60,10 @@ def cleanup(self):
         # remove the lock on the journal
         job = self.background_job
         params = job.params
-        journal_id = params.get("set_in_doaj__journal_id")
+        journal_ids = params.get("set_in_doaj__journal_ids")
```
emanuil-tolev: cls.get_param(params, 'journal_ids'). Maybe with a default of [], not sure if get_param supports defaults, but can be easily put in.

portality/tasks/journal_in_out_doaj.py
```diff
@@ -70,20 +80,23 @@ def prepare(cls, username, **kwargs):
         job.user = username
         job.action = cls.__action__
 
-        journal_id = kwargs.get("journal_id")
+        journal_ids = kwargs.get("journal_ids")
 
-        params ={}
-        params["set_in_doaj__journal_id"] = journal_id
+        params = {}
+        params["set_in_doaj__journal_ids"] = journal_ids
```
emanuil-tolev: use cls.set_param

portality/tasks/journal_in_out_doaj.py
```diff
@@ -70,20 +80,23 @@ def prepare(cls, username, **kwargs):
...
         params["set_in_doaj__in_doaj"] = kwargs.get("in_doaj")
 
-        if params["set_in_doaj__journal_id"] is None or params["set_in_doaj__in_doaj"] is None:
+        if params["set_in_doaj__journal_ids"] is None or params["set_in_doaj__in_doaj"] is None:
```
emanuil-tolev: params.get_param

portality/tasks/journal_in_out_doaj.py
```diff
@@ -70,20 +80,23 @@ def prepare(cls, username, **kwargs):
...
             raise RuntimeError(u"SetInDOAJBackgroundTask.prepare run without sufficient parameters")
 
         job.params = params
 
+        if "selection_query" in kwargs:
```
emanuil-tolev: if? Are you allowed to not set this? kwargs['selection_query']

portality/tasks/journal_in_out_doaj.py
emanuil-tolev: also, Bulk Edit classes have an example of validating incoming params and making some required, then you just call `_job_parameter_check` on this line. It encapsulates the param check so that prepare and run (which both need to check the params) use the exact same piece of code.

portality/tasks/journal_in_out_doaj.py
emanuil-tolev: This is not a problem, ignore.

Issue #1126, opened 2016-11-07
Install and "hello world" Huey


emanuil-tolev:
- basic config
- basic task file
- write a simple task, like journal activate + deactivate (take in/out of DOAJ)
- ensure we know how to make huey consume the task and that it does it when the task is triggered

PR #1174, opened 2017-01-16
Feature/1126 redis docker compose setup with persistence

emanuil-tolev:
#1126 

This is a production-ready dockerised Redis with a convenient developer interface as well (just `docker-compose up` when you want to start Redis, as the docs say).


emanuil-tolev:
Sorry, wrong branch name. I decided to create a new issue to track this better. #1178 is the new PR.

Issue #1177, opened 2017-01-16
Redis


emanuil-tolev:
- [x] set up conveniently for devs
- [x] docker must be used
- [x] enable persistence and test it
- [x] don't throw away records (don't use Redis as a cache)
- [x] set maxmemory
- [x] spin up new server (?) or run Redis on application server in production. Either way, set maxmemory appropriately - we don't intend to use much more than 10-20 MB, so perhaps set to 100MB and be away with it without creating another machine for now.
    The deliverable here is a fully running, configured and automatically deployed (i.e. configuration Redis changes in DOAJ repo are reflected on deploy) Redis instance using Docker.
- [x] make sure it's secure - docker daemon bindings do not respect ufw rules
- [ ] make sure it starts on reboot and its logs are accessible. A page on the wiki on how to view its logs would be a good start. Perhaps look into running docker-compose under supervisor for the "on boot" and gaining "retry on failure" functionality, but if this looks complex, just make sure it runs on machine boot.

PR #1178, opened 2017-01-16
Feature/1177 redis docker compose setup with persistence

emanuil-tolev:
#1177 

This is a production-ready dockerised Redis with a convenient developer interface as well (just `docker-compose up` when you want to start Redis, as the docs say).


emanuil-tolev:
Once merged, if you have defined HUEY_REDIS_PORT in your dev.cfg you'll have to move this to your app.cfg. This was done because docker-compose uses `.env` in the current working dir as its default environment file, and I couldn't symlink it to 3 different files depending on context (dev.cfg, test.cfg and production.cfg). I see that Richard's move to make octopus read from `local.cfg` is a wise one indeed.

emanuil-tolev:
Example - that's my app.cfg:

```bash
DOAJENV='dev'
HUEY_REDIS_PORT=6400
```

The point is it has to be bash-compatible, so no spaces around equal signs!

Issue #985, opened 2016-02-10
New article search facet: From a Journal with the DOAJ Seal


remocrevo:
I would like to be able to search for articles only within journals that have the DOAJ Seal. This would provide quick access to full-text searching of the best of DOAJ.  Currently DOAJ provides a journal search facet to limit by the DOAJ Seal, resulting in a URL like this:

https://doaj.org/search?source={"query":{"filtered":{"filter":{"bool":{"must":[{"term":{"_type":"journal"}},{"term":{"index.has_seal.exact":"Yes"}}]}},"query":{"match_all":{}}}},"from":0,"size":10}

But it doesn't yet support applying the same index to an article search.  Replacing the "_type" with "article" (or using the interface to choose "Articles" and then "DOAJ Seal: Yes") should produce a list of articles published in journals that have received the DOAJ Seal.

emanuil-tolev:
Articles currently hold a limited set of metadata regarding their parent journals, and the seal is not of one the items unfortunately (which is why your search does not work). However, it is a good idea.

@greboun @dommitchell this will involve a few hours to get a simple migration to add this piece of information to all articles so we could probably add it to TnM if you think it worthwhile. (Unless @richard-jones or @Steven-Eardley think it'll be more complex, this is just a finger in the air estimate.)

richard-jones:
Yes, it would involve adding the seal information to the journal info synchronising that @Steven-Eardley is working on.

The question is for the DOAJ admins to decide if this is functionality that should be provided, and balance that with the user experience, given there are already a lot of facets for articles (remember we kept the list of facets as short as possible to minimise information overload on users).

The other thing to consider is whether this information would be useful in the API, even if it is not in the UI.

@remocrevo - would you be likely to use the DOAJ API directly, or are you only interested in the search UI?

remocrevo:
@richard-jones - I could make use of the API if that were the only option, but the search UI would be much more useful. I'm imagining telling students "Click on this link to DOAJ (which limits to journals with the seal) and search for articles for your research project".

As for the concern of overloading users with too many facets, I currently count 10 facets if I select "Journals", and only 6 if I select "Articles". Isn't there room to add the DOAJ Seal to the list of article facets? If there is still concern, I would be happy with a "More" link that reveals advanced or less common facets (as others have suggested).

richard-jones:
Assigned to @greboun for review.  Can be included in tnm if required.

dommitchell:
[...]

PR #1184, opened 2017-01-20
Added the DOAJ seal facet to articles, and updated the sync function …

Steven-Eardley:
…to retrieve it from the Journal. Doc and model updates for the new field. Updates to test and a new test for this functionality. Version bump for JS changes. Removed empty file. #985


Steven-Eardley:
For this to show in the facets, ```add_journal_metadata()``` and ```prep()``` needs to be run on all articles. The ```article_cleanup_sync``` script is currently out of action, however.

PR #1192, opened 2017-01-24
normalise usage of backgrount tasks

richard-jones:


emanuil-tolev:
Branch had 1 conflict - it had introduced the article cleanup script background task back into the huey schedule. I've removed it again (by leaving only a commented out entry). As far as I understood after you two are done with the script this Friday you will reintroduce it into the huey schedule.

emanuil-tolev:
However, I see that actually the schedule is now used in the huey article cleanup script scheduled task. Are we happy for this task to run on the test server?

portality/tasks/journal_in_out_doaj.py
```diff
@@ -83,16 +85,18 @@ def prepare(cls, username, **kwargs):
         journal_ids = kwargs.get("journal_ids")
 
         params = {}
-        params["set_in_doaj__journal_ids"] = journal_ids
-        params["set_in_doaj__in_doaj"] = kwargs.get("in_doaj")
+        cls.set_param(params, "journal_ids", journal_ids)
+        cls.set_param(params, "in_doaj", kwargs.get("in_doaj"))
 
-        if params["set_in_doaj__journal_ids"] is None or params["set_in_doaj__in_doaj"] is None:
+        if journal_ids is None or kwargs.get("in_doaj") is None:
```
emanuil-tolev: journal_ids could be `[]` as well, which would not be valid. Perhaps best to just ask `if journal_ids or kwargs.get("in_doaj") is None`.

portality/tasks/journal_in_out_doaj.py
```diff
@@ -34,8 +36,8 @@ def run(self):
         job = self.background_job
         params = job.params
 
-        journal_ids = params.get("set_in_doaj__journal_ids")
-        in_doaj = params.get("set_in_doaj__in_doaj")
+        journal_ids = self.get_param(params, "journal_ids")
+        in_doaj = self.get_param(params, "in_doaj")
 
         if journal_ids is None or len(journal_ids) == 0 or in_doaj is None:
```
emanuil-tolev: Might be better to factor this out to a `_job_parameter_check` classmethod, see portality.tasks.journal_bulk_edit.JournalBulkEditBackgroundTask for an example definition and use.

The exact same logic needs to be executed both in prepare() and in run() at the beginning of the methods. Putting the definition in one place should make it much harder to modify the condition only in one place, forgetting about the other one. (In fact I've suggested a change to this if condition in the prepare() method below.)

emanuil-tolev:
I've restored article_cleanup_sync to the schedule Richard originally put up with this PR. Just need to know if it's OK to merge into develop (pending the other comment on the PR), since I was under the impression we don't want to run that script on the test server yet.

portality/tasks/journal_in_out_doaj.py
richard-jones: I'm against checks for falsity in this way, but yes we can definitely extend the check to cover the empty list

richard-jones:
We can leave it commented out, I think I just uncommented it to check it all looked ok.

I'll do another revision of the code shortly, with the requested changes.

portality/tasks/journal_in_out_doaj.py
emanuil-tolev: I wouldn't propose it on a boolean either (e.g. the in_doaj param)

emanuil-tolev:
> We can leave it commented out, I think I just uncommented it to check it all looked ok.

Could you comment it back out in that case? Including the little scheduled_article_sync function that refers to the schedule from settings.py. If only the settings.py line is commented out, the build fails.

portality/tasks/journal_in_out_doaj.py
richard-jones: After giving this a bit of thought and trying it out on one class, I've decided not to do it.  The checks on prepare and the checks on run are not necessarily the same, and the kwargs to prepare and the params for run are not necessarily the same or consistently named.  After fiddling with it the _job_parameter_check seemed to be a bit more complex than was necessary and also made it less clear what prepare and run were doing in my opinion.

emanuil-tolev: Ah, ok. Perhaps the kwargs were named the same only on the jobs I did.

Issue #1175, opened 2017-01-16
Refactor delete articles and delete journals buttons into background jobs


emanuil-tolev:

richard-jones:
Done

PR #1194, opened 2017-01-25
Feature/1175 delete bulk - make background job. Journals only, and not tied to UI yet.

emanuil-tolev:
#1175 partial implementation. Journals only, see last comment for what remains.


emanuil-tolev:
Our mappings are in serious need of sorting out. I'm trying to write tests for the bulk delete operation on journals (and their assoc articles) and have already had to fudge a query by using consistent_order=False so I could avoid the sort it does - or the entire thing fails with a mapping error.

I'm now stuck on one that I can't quite fix:
```
[2017-01-25 17:54:33,712][DEBUG][action.search.type       ] [Possessor] [doajtest][1], node[A2l_VL45TESEYeg3Nw77_g], [P], s[STARTED]: Failed to execute [org.elasticsearch.action.search.SearchRequest@4c98ec1a]
org.elasticsearch.search.SearchParseException: [doajtest][1]: from[-1],size[-1]: Parse Failure [Failed to parse source [{"sort": [{"last_updated": {"order": "desc"}}], "query": {"bool": {"must": [{"term": {"about.exact": "journalid4"}}, {"term": {"type.exact": "journal"}}]}}}]]
        at org.elasticsearch.search.SearchService.parseSource(SearchService.java:687)
        at org.elasticsearch.search.SearchService.createContext(SearchService.java:543)
        at org.elasticsearch.search.SearchService.createAndPutContext(SearchService.java:515)
        at org.elasticsearch.search.SearchService.executeQueryPhase(SearchService.java:277)
        at org.elasticsearch.search.action.SearchServiceTransportAction$5.call(SearchServiceTransportAction.java:231)
        at org.elasticsearch.search.action.SearchServiceTransportAction$5.call(SearchServiceTransportAction.java:228)
        at org.elasticsearch.search.action.SearchServiceTransportAction$23.run(SearchServiceTransportAction.java:559)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: org.elasticsearch.search.SearchParseException: [doajtest][1]: from[-1],size[-1]: Parse Failure [No mapping found for [last_updated] in order to sort on]
        at org.elasticsearch.search.sort.SortParseElement.addSortField(SortParseElement.java:210)
        at org.elasticsearch.search.sort.SortParseElement.addCompoundSortField(SortParseElement.java:184)
        at org.elasticsearch.search.sort.SortParseElement.parse(SortParseElement.java:86)
        at org.elasticsearch.search.SearchService.parseSource(SearchService.java:671)
```

I have no idea what is even issuing this query - I didn't write it during bulk deleting. 

It may have to do with snapshots..? I think that's what the "about.exact" is for. As you will be able to see, in the new test I am creating at least one article and one journal snapshot before anything else happens, just to make sure these objects exist in the index alongside articles and journals, so that ES recognises their "last_updated". Unfortunately to no avail.

I am using the exact same code that the old Delete buttons in the admin area used, so I suspect this will work fine on test and live. The problem is that if we had tests for those delete buttons (do we - I can't find them?) then those tests would have failed as well due to this problem.

richard-jones:
Probably this is just because there's no data in the index type.  This mapping failure is true of all our types, and usually it doesn't matter as there's data in them, so the mappings have been auto-created.  This is partly why I'm motivated to get mappings created from dataobj structs.  I do have some other tricks that I use in Octopus, too, such as pushing an example document and then deleting it at init time, which sets up the mappings.

emanuil-tolev:
But there are documents in the type! I create a journal and an article snapshot in the test setup for that exact reason.

emanuil-tolev:
So there are articles, journals, 1 article snapshot and one journal snapshot when the test runs.

richard-jones:
That's pretty weird, since last_updated is applied by all our DAOs to the objects in the index.

You said you weren't sure where the query was being generated, have you pinned that down yet?

Also, didn't we do away with the snapshot index?  If that's being queried, perhaps we have some legacy code that can just be removed.

I am around tomorrow to pair on it if you want; Steve and I are also talking about the article sync script, so will be on a DOAJ call on and off during the day.

emanuil-tolev:
> That's pretty weird, since last_updated is applied by all our DAOs to the objects in the index.
> 
> You said you weren't sure where the query was being generated, have you pinned that down yet?

Yes, it was locks. The lock type must not exist in the index for some reason. I've changed the code issuing the lock query so it deals with the situation, though I think overall our handling of such situations can be improved a lot further than I've gone. If the lock type does not exist, then obviously no locks exist, so I let the function return as if no locks were found (which is true). It would be worth having your thoughts on how it's now being handled in the DAO, I think it can probably be done more elegantly.

I did improve the ES logging further. The source of me being unable to understand the exact problem was again some data object code outside of dao.py - the Lock model was interpreting a raw ES response. I've moved it to a dao.py method that will give a good error, though I've shied away from modify `DomainObject.query` since that could break parts of the app that are currently trundling along issuing queries that error out.

> Also, didn't we do away with the snapshot index? If that's being queried, perhaps we have some legacy code that can just be removed.

Turned out it wasn't snapshot: "about" is a term used in locking.

The now reasonable test passes for journal + associated article deletion. Remaining are:

- bulk article deletion - this is entirely a subset of journal + associated article deletion so I expect little trouble
- the UI needs to be modified so that the bulk "Submit" button is only enabled if an option has been selected on the Journals vs. Articles facet. I think we do similar restrictions elsewhere so this shouldn't be too bad.
- the UI needs to send the bulk request to /admin/journals/bulk/delete or /admin/articles/bulk/delete appropriately
- the code that the old delete buttons used must be deleted.

richard-jones:
Ok, this looks fine to me.

We're paying a bit of a price for the very old data access layer in DOAJ.  I'd like to switch it out for esprit at some point, where we can focus on more general solutions to ES communication/query issues.  That said, esprit also needs an overhaul due to patchy changes due to ES version increments.

emanuil-tolev:
Ok, thanks. I'm getting the rest of the changes detailed in my last comment up on another PR.

PR #1195, opened 2017-01-27
Feature/1183 article cleanup enhance

richard-jones:
This introduces a bunch of performance enhancements and feature improvements to the article cleanup script.  While it will max the CPU as much as it is allowed, it should use less than 150Mb of memory now (possibly as little as 100Mb).  Estimated run time for the entire index is around 10 hours.

In order to allow this to run alongside shorter background tasks without interfering with them, I've also created a new queue I've called "long_running", where we could put other non-trivial tasks to be executed.  You now need to start these with

   deploy/huey_main_queue.sh
   deploy/huey_long_running.sh

### DIFF
--- a/deploy/huey.sh
+++ b/deploy/huey.sh
@@ -1,2 +1,0 @@
-#!/usr/bin/env bash
-huey_consumer.py portality.tasks.consumer.main_queue
--- a/deploy/huey_long_running.sh
+++ b/deploy/huey_long_running.sh
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+huey_consumer.py portality.tasks.consumer_long_running.long_running
--- a/deploy/huey_long_running.sh
+++ b/deploy/huey_long_running.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-huey_consumer.py portality.tasks.consumer_long_running.long_running
+huey_consumer.py portality.tasks.consumer_long_running.long_running >> ~/huey_long_running.log 2>&1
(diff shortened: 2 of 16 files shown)


emanuil-tolev:
3 failing tests, all related to TestArticleCleanupSync. Did you introduce new env variables or anything that might not be on codeship, but is on your local machine? (Assuming these tests don't fail locally - if they do, that's easy.)

https://app.codeship.com/projects/49473/builds/22106599

Steven-Eardley:
They do fail for me locally now.

richard-jones:
@Steven-Eardley - are you looking at this, or should I?

This is weird, because one of those errors I saw before when running the tests with nose, but not when running the tests with PyCharm.

richard-jones:
ok, this was because I'd added a new queue called long_running and I hadn't set always_eager on it during testing.  All the tests are now passing in PyCharm, but I'm interested to see if this one rogue test which failed in nose also fails on codeship.

Steven-Eardley:
@richard-jones that's passing now, both locally for me and on codeship.

emanuil-tolev:
Let me just test this quickly on the branch before merging. I'm going to run article cleanup sync, but put a thousand year sleep in it. Then I'll watch the regular tasks (CSV, sitemap, whatever) and my manually requested bulk actions get executed alongside.

Steven-Eardley:
Confirmed tasks run alongside a long running wait on the second queue. Merging and we'll see what happens on test. Unless we run it manually before, that'll be the 2nd day of next month.

Issue #1196, opened 2017-01-30
Wrong index.publisher field being generated for some records


emanuil-tolev:
The suggestion with id `399405c701a14278b54c94529c7144e2` in live has the following field in its `index` section:

```json
"publisher": "[u'SN Education Society']"
```

That's weird, and it makes searches that have specified "search only in the publisher field" fail.

Rikard has reported it for some journals as well, although a short while later, when I viewed the search link, the journal was showing up correctly. I don't know whether this application will also fix itself, but it shouldn't be broken in the first place, obviously. It seems like the result of a deterministic problem, like it would either work correctly or not.

Here's the search link by Rikard: http://bit.ly/2jJzFHN . At the time of writing this shows "no results". Change the field-to-search-in to "search all" instead of "Publisher" and you will see the records appear. That's probably because "Publisher" translates to "index.publisher.exact".

I don't know how high priority this is, but I suppose not being able to discover records in the obvious ways is not something we want to keep around for too long.

emanuil-tolev:
This could be caused by earlier code that is now fixed in production, by the way - something to keep in mind while trying to reproduce locally. So locally, if you put `[u'SN Education Society']` into index.publisher of an application, and then .save() it, that might well fix it. If so the obvious overall fix would be a simple migration.

richard-jones:
I've looked at suggestion 399405c701a14278b54c94529c7144e2 and it looks fine now.

Inspecting the code, it looks like the struct is wrong, though.  It says that publisher is a single value unicode field, but it's actually a list of unicodes.

So, no risk of data loss here, but I'll straighten out the struct and provide a hotfix shortly.

ClaraDOAJ:
Looks like it is fixed! Many thanks

ClaraDOAJ:
We are still seeing the problem in the Admin - Applications
![image](https://cloud.githubusercontent.com/assets/22593616/23204381/b194ae52-f8e6-11e6-8267-037f8394ab70.png)

richard-jones:
Apologies, this is my fault, I didn't update the application model definition to store the index.publisher correctly there - only for the journals themselves.

I have submitted a PR for this: #1230

richard-jones:
Fixed on live

PR #1197, opened 2017-01-31
HOTFIX to ensure publisher index field is correctly structured

richard-jones:
HOTFIX, see issue #1196 

This changes the struct for the journals and applications, so that the publisher index field is now a list not a unicode string.  This should line up with the index mapping.

There is a migration that should be run which will make sure that the index is updated, though it is likely that items that are incorrect will correct themselves in time anyway.

### DIFF
--- a/portality/migrate/1196_publisher_struct/README.md
+++ b/portality/migrate/1196_publisher_struct/README.md
@@ -1,0 +1,11 @@
+# 1196_publisher_struct
+
+RJ 31-01-2017
+
+Migration script to be run to convert incorrect index.publisher fields to correct form, caused
+by incorrect struct.  See issue 1196.
+
+Run
+
+    python portality/upgrade.py -u portality/migrate/1196_publisher_struct/publisher_struct.json
+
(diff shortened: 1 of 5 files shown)


richard-jones:
Add some more mechanics in the model objects to try to avoid encapsulation violation of the dataobj.  The index section of the object should be set via the struct.

emanuil-tolev:
> Add some more mechanics in the model objects to try to avoid encapsulation violation of the dataobj. The index section of the object should be set via the struct.

@richard-jones that's notes from a tech meeting a few weeks ago I think. Do you need me or @Steven-Eardley to review anything further here?

richard-jones:
ok, can we get this merged and deployed?  Does it need to go out to the test server first?

Steven-Eardley:
@richard-jones Did this comment of yours
>Add some more mechanics in the model objects to try to avoid encapsulation violation of the dataobj. The index section of the object should be set via the struct.

mean there's more work to do on this branch before merge? We probably will run on test first.

richard-jones:
hm, I would like to do that change, but I don't want to hold this fix up given that it's affecting publicly visible data.  Let me think about it today and make a decision this afternoon.

richard-jones:
ok, this wasn't a huge job to do, so I've done it

emanuil-tolev:
Deploying now. Should this impact any tests? (As in, does the current passing build imply we need to write some new tests, or is the code covered just fine and it passes because it works?)

PR #1203, opened 2017-02-02
bulk article delete; no UI yet

emanuil-tolev:
Fully working, tested. Back-end only.

Remaining:

- on the UI, disable the bulk job Submit button if the Delete option is selected and if the user has not yet selected an option in the _type facet.
- make sure the UI takes the option selected in the _type facet into account when submitting bulk delete jobs. There are 2 URLs in admin.py, not one.

### DIFF
--- a/portality/tasks/article_bulk_delete.py
+++ b/portality/tasks/article_bulk_delete.py
@@ -1,0 +1,128 @@
+from copy import deepcopy
+import json
+
+from flask_login import current_user
+
+from portality import models
+
+from portality.tasks.redis_huey import main_queue
+from portality.decorators import write_required
+
+from portality.background import AdminBackgroundTask, BackgroundApi, BackgroundException
+
+
+def article_bulk_delete_manage(selection_query, dry_run=True):
+
+    if dry_run:
+        ArticleBulkDeleteBackgroundTask.check_admin_privilege(current_user.id)
+        return ArticleBulkDeleteBackgroundTask.estimate_delete_counts(selection_query)
+
+    ids = ArticleBulkDeleteBackgroundTask.resolve_selection_query(selection_query)
+    job = ArticleBulkDeleteBackgroundTask.prepare(
+        current_user.id,
+        selection_query=selection_query,
+        ids=ids
+    )
+    ArticleBulkDeleteBackgroundTask.submit(job)
+
+    return True
+
+
+class ArticleBulkDeleteBackgroundTask(AdminBackgroundTask):
+
+    __action__ = "article_bulk_delete"
+
+    @classmethod
+    def _job_parameter_check(cls, params):
+        # we definitely need "ids" defined
+        return bool(cls.get_param(params, 'ids'))
+
+    def run(self):
+        """
+        Execute the task as specified by the background_job
+        :return:
+        """
+        job = self.background_job
+        params = job.params
+
+        ids = self.get_param(params, 'ids')
+
+        if not self._job_parameter_check(params):
+            raise BackgroundException(u"{}.run run without sufficient parameters".format(self.__class__.__name__))
+
+        estimate = self.estimate_delete_counts(json.loads(job.reference['selection_query']))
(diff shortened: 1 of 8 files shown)


portality/tasks/article_bulk_delete.py
```diff
@@ -0,0 +1,128 @@
...
+
+    ids = ArticleBulkDeleteBackgroundTask.resolve_selection_query(selection_query)
+    job = ArticleBulkDeleteBackgroundTask.prepare(
+        current_user.id,
+        selection_query=selection_query,
+        ids=ids
+    )
+    ArticleBulkDeleteBackgroundTask.submit(job)
+
+    return True
```
richard-jones: Shouldn't this also return the delete count estimate, so there's a consistent return type?  Does the UI also need the delete count to be returned once the job has been successfully delivered?

portality/tasks/article_bulk_delete.py
```diff
@@ -0,0 +1,128 @@
...
+        job = self.background_job
+        params = job.params
+
+        ids = self.get_param(params, 'ids')
+
+        if not self._job_parameter_check(params):
+            raise BackgroundException(u"{}.run run without sufficient parameters".format(self.__class__.__name__))
+
+        estimate = self.estimate_delete_counts(json.loads(job.reference['selection_query']))
+        job.add_audit_message(u"About to delete an estimated {} articles".format(estimate))
```
richard-jones: Don't we know the number of articles by the list of ids that are passed in?

portality/tasks/article_bulk_delete.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # TODO this could be an issue with lots of article IDs (the should_terms being so big)
+        # Break up in batches if a problem occurs during manual testing.
+        # Might need a unit or integration test to test batching up works.
+        article_delete_q_by_ids = models.Article.make_query(should_terms={'_id': ids}, consistent_order=False)
```
richard-jones: Since we're operating asynchronously, I'd be inclined to do this in batches anyway.  It wouldn't be too difficult just to break down the ids into groups of, say, 1000 and churn through them.  A "terms" query takes a max of 1024, so it wouldn't be surprising if a should query has similar limits.

portality/tasks/article_bulk_delete.py
```diff
@@ -0,0 +1,128 @@
...
+        :return: a BackgroundJob instance representing this task
+        """
+
+        super(ArticleBulkDeleteBackgroundTask, cls).prepare(username, **kwargs)
+
+        # first prepare a job record
+        job = models.BackgroundJob()
+        job.user = username
+        job.action = cls.__action__
+        job.reference = {'selection_query': json.dumps(kwargs['selection_query'])}
```
richard-jones: Need to namespace the field.  There's a similar utility method to your set_param method for references which does this.

portality/view/admin.py
```diff
@@ -545,3 +547,19 @@ def bulk_journals_delete():
...
+    validate_json(payload, fields_must_be_present=['selection_query'], error_to_raise=BulkAdminEndpointException)
+
+    q = get_query_from_request(payload)
+
+    r = article_bulk_delete.article_bulk_delete_manage(
+        selection_query=q,
+        dry_run=payload.get('dry_run', True)
+    )
+
+    return make_json_resp(r, status_code=200)
```
richard-jones: If this isn't a dry_run, then r will be "True".

portality/tasks/article_bulk_delete.py
emanuil-tolev: Yes, this was likely from a previous local iteration of the code where I was wondering when to convert the query into actual records - during the background job run, or during submission. We decided it should be during submission so yes, we do know how many IDs.

emanuil-tolev:
Ready for re-review @richard-jones

richard-jones:
Is this PR now superseded by your other one?

Steven-Eardley:
It's a prerequisite

PR #1209, opened 2017-02-08
Revert "Merge branch 'develop' into hotfix/2017-01-30_iss1051_url_sho…

Steven-Eardley:
…rtner_tocs"

This reverts commit 78b08cbd125a83c45676c8a626f2ced13d609fa4, reversing
changes made to 4039d91963167bd2208390dc9ff351be44406ac4.


Steven-Eardley:
The escaped changes were reverted using ```git revert``` which means a new commit was added un-doing the merge. To restore the changes, this revert needs to be reverted. Or just replace production with develop, I suppose. That might be cleaner.

emanuil-tolev:
> To restore the changes, this revert needs to be reverted.

Yep, exactly.



> Or just replace production with develop, I suppose. That might be cleaner.

This will cause conflicts for anyone who pulls locally. If they resolve them with "merge" it'll be a nice mess :). We could do it since we're a small team atm, and this conflict would only arise if the person has pulled since .. last Thursday, around 1-2 Feb.

emanuil-tolev:
> This will cause conflicts for anyone who pulls locally. If they resolve them with "merge" it'll be a nice mess :). We could do it since we're a small team atm, and this conflict would only arise if the person has pulled since .. last Thursday, around 1-2 Feb.

I would do this in a meeting together btw, if we do it. "revert revert" is standard, I'd be fine with that personally.

Issue #1094, opened 2016-10-10
Group name sort for Editorial Groups doesn't work


dommitchell:
https://doaj.org/admin/editor_groups?source=%7B%22query%22%3A%7B%22match_all%22%3A%7B%7D%7D%2C%22sort%22%3A%5B%7B%22name%22%3A%7B%22order%22%3A%22desc%22%7D%7D%5D%2C%22from%22%3A0%2C%22size%22%3A44%7D

richard-jones:
@Steven-Eardley could you take a look at this one when you have a chance

Steven-Eardley:
Need to add a ```.exact``` to https://github.com/DOAJ/doaj/blob/develop/portality/static/doaj/js/available_facetviews/admin.editorgroups.facetview.js#L24

Steven-Eardley:
Released to live - the editor groups can now be sorted A-Z or Z-A.

ClaraDOAJ:
Many thanks!

PR #1210, opened 2017-02-08
Added .exact to fix editor group search sort. 

Steven-Eardley:
Issue #1094.

### DIFF
--- a/portality/static/doaj/js/available_facetviews/admin.editorgroups.facetview.js
+++ b/portality/static/doaj/js/available_facetviews/admin.editorgroups.facetview.js
@@ -21,7 +21,7 @@ jQuery(document).ready(function($) {
             {'display':'Created Date','field':'created_date'},
             {'display':'Last Modified Date','field':'last_updated'},
             {'display':'Editor ID','field':'editor'},
-            {'display':'Group Name','field':'name'}
+            {'display':'Group Name','field':'name.exact'}
         ],
 
         searchbox_fieldselect: [


portality/static/doaj/js/available_facetviews/admin.editorgroups.facetview.js
```diff
@@ -21,7 +21,7 @@ jQuery(document).ready(function($) {
             {'display':'Created Date','field':'created_date'},
             {'display':'Last Modified Date','field':'last_updated'},
             {'display':'Editor ID','field':'editor'},
```
richard-jones: While we're here, this one probably should be changed too

Steven-Eardley: Editor account names tend to be one word, so I'm seeing it ordering as we expect on the live site. I can change it for completeness, if you like.

emanuil-tolev: ah but there's more to it than that: I remember a problem where you look for account 01234567X and it doesn't turn up, despite the account existing (same string for the username). At the time it was to do with the fact the field (without .exact) was analysed, so ES was looking for a lowercase letter 'x' or something of the sort. We should decide if it should be an analysed field or not and go with one, but there are consequences to doing one or the other (e.g. with .exact it'd really need to match the query exactly).

richard-jones: That would be to do with search rather than sort, but if we change it here we should change it there, too, since you'd probably be looking for the exact editor name string.

Steven-Eardley: @richard-jones I don't like it. Adding ```.exact``` to the search fields means you must match the letter case when you are searching, e.g. to get "West Africa" you can't type 'west' or 'west africa'.

Also, if I add the ```.exact``` to the editor ID sort field, the one editor who's name begins with an upper-case character is at the top of the list, before 'a'. Without the ```.exact``` they are in the correct place.

richard-jones: ok, since there's no specific question over this field, then, let's leave it.

PR #1219, opened 2017-02-14
Feature/1175 Bulk delete UI

emanuil-tolev:
#1175 

I should note this also includes work from #1203 , so #1203 would be easier to review first. Then this diff will shrink considerably to only include my UI work.

### DIFF
--- a/doajtest/unit/test_task_article_bulk_delete.py
+++ b/doajtest/unit/test_task_article_bulk_delete.py
@@ -1,0 +1,55 @@
+from time import sleep
+import json
+
+from doajtest.helpers import DoajTestCase
+
+from portality import models
+from portality.tasks.article_bulk_delete import article_bulk_delete_manage
+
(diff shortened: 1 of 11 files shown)


emanuil-tolev:
I should note this also includes work from #1203 , so #1203 would be easier to review first. Then this diff will shrink considerably to only include my UI work.

emanuil-tolev:
Huey needs to be restarted in its screen on doaj-test after merging this. It would be great if whoever merges it does that or leaves a comment for me to do.

Steven-Eardley:
On test now, and Huey restarted.

emanuil-tolev:
> I think it's worth merging this so we can at least confirm that the functionality works through the UI. If we then go on to re-structure how it is presented, that's another step. I'll give that one some thought.

Yeah this doesn't really pertain to whether we decide to simply remove the option from the menu, or only add it for technical admin staff (CL) or whatever.

PR #1230, opened 2017-02-26
HOTFIX modify struct for publisher in index for applications

richard-jones:
HOTFIX

Modify application struct for index.publisher field, omitted by error in the previous hotfix.

After this has been applied, the original migrate script needs to be run again:

    python portality/upgrade.py -u portality/migrate/1196_publisher_struct/publisher_struct.json


emanuil-tolev:
Can be deployed. I'll do it in a few hours if somebody else doesn't before me.

Steven-Eardley:
Deploying now

Issue #1218, opened 2017-02-14
Intermittent build failure related to article cleanup tests


emanuil-tolev:
```
======================================================================
ERROR: test_03_prepall (test_article_cleanup_sync.TestArticleCleanupSync)
----------------------------------------------------------------------
Traceback (most recent call last):
File "/home/rof/src/github.com/DOAJ/doaj/doajtest/unit/test_article_cleanup_sync.py", line 133, in test_03_prepall
assert a1u.bibjson().data.get("journal") == a1.bibjson().data.get("journal")
AttributeError: 'NoneType' object has no attribute 'bibjson'
-------------------- >> begin captured stdout << ---------------------
account 200
bulk_upload 200
bulk_reapplication 200
background_job 200
provenance 200
lock 200
journal 200
cache 200
upload 200
journal_history 200
article_history 200
lcc 200
suggestion 200
toc 200
article 200
news 200
editor_group 200

--------------------- >> end captured stdout << ----------------------

----------------------------------------------------------------------
Ran 273 tests in 564.473s

FAILED (errors=1)
```

emanuil-tolev:
Further intemittent failure:

```
======================================================================
FAIL: test_02_write (test_article_cleanup_sync.TestArticleCleanupSync)
----------------------------------------------------------------------
Traceback (most recent call last):
File "/home/rof/src/github.com/DOAJ/doaj/doajtest/unit/test_article_cleanup_sync.py", line 64, in test_02_write
assert models.Journal.count() == 1
AssertionError: 
-------------------- >> begin captured stdout << ---------------------
account 200
bulk_upload 200
bulk_reapplication 200
background_job 200
provenance 200
lock 200
journal 200
cache 200
upload 200
journal_history 200
article_history 200
lcc 200
suggestion 200
toc 200
article 200
news 200
editor_group 200

--------------------- >> end captured stdout << ----------------------

----------------------------------------------------------------------
Ran 273 tests in 566.977s

FAILED (failures=1)
```

emanuil-tolev:
This happens in about 70-80% of my builds.

emanuil-tolev:
Another failure in the same file. Probably all those are for the same reason?

```
======================================================================
FAIL: test_03_prepall (test_article_cleanup_sync.TestArticleCleanupSync)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #1232, opened 2017-03-07
Change how the article cleanup sync job is being executed, to avoid t…

Steven-Eardley:
…he delay instroduced by using submit() #1218


Steven-Eardley:
I want this in my new branches, so I'm merging it.

richard-jones:
The way we've done this in other tasks is to put the blocking save in the submit() function instead.  This must have just got overlooked when we finalised this task.  See, for example: https://github.com/DOAJ/doaj/blob/develop/portality/tasks/article_bulk_delete.py#L129

PR #1242, opened 2017-03-21
Release/2017 03 21 background jobs

emanuil-tolev:
Release PR, for my own and your perusal!

I'm going to list what migrations I need to run to make this live in here, and have a final look at code.


emanuil-tolev:
Everything looks good to me. No new migrations to run, just deploy the code, make sure Redis works, S3FS works and huey works. Theoretically the deployment scripts should take care of that, and we created and configured the necessary live machines earlier in the project.

I did catch one problem that would've caused a failure - adding dependencies to setup.py still doesn't install them on production. So I've added rstr to the deployment scripts to install explicitly. It's a known issue with the live environment, but there are (and have been) more important issues to deal with. I'm sure its time will come at some point.

Issue #1245, opened 2017-03-21
Customise background jobs email for non-administrative users


richard-jones:
When background jobs complete they send an email to the user.  If the user is a publisher, they should get a more human-readable message, rather than the one which is currently aimed at administrators.

Steven-Eardley:
To clarify the status of this, the email was removed in the above merged pull request.

Issue #1236, opened 2017-03-10
Retain Article XML files in the event of a failure


richard-jones:
To aid in any future debugging of the article xml upload feature, we should keep copies of any files where an error occurs.

richard-jones:
This has now been deployed.

Issue #1093, opened 2016-10-08
Test harness for XML files


richard-jones:
Would be good if we could develop a script or even UI feature for validating article XML files in a way that gives the user more feedback.  There is a recent request to review a couple of files, so propose that we put such a feature together as a script initially, and then use those files as a test-base.

dommitchell:
This is an excellent idea. Ideally it will be a process that I can use without bothering CL.

dommitchell:
CL will discuss this week,.

richard-jones:
This is live

PR #1254, opened 2017-03-25
Feature/1093 test article xml

richard-jones:
This PR adds significant improvements in the way that errors are recorded and presented during article ingest:

* all exceptions are logged in the background job and in the file upload object.  This is particularly useful when the XML fails schema validation, as it will give the reasons for the validation failure.
* All failed article imports (for valid XML files) have their reasons for failure logged: whether they contain an ISSN which doesn't match a journal, an ISSN which has shared or confused ownership, or an ISSN which is owned by another user account
* All failed imports are kept and put in a directory for inspection by the technical team if necessary
* Flash messages during initial upload are now enhanced with the additional error reporting
* File upload history display in the publisher area is now enhanced with the additional error reporting

This PR also stops sending background job completion emails to publishers, as they are no use for the publisher, as they contain links to pages they cannot access (i.e. the background jobs page).

Affected issues here are: #1236 #1093 and #1245


portality/tasks/ingestarticles.py
```diff
@@ -365,29 +387,38 @@ def _file_upload(cls, username, f, schema, previous):
...
             previous.insert(0, record)
         else:
+            # This code should never get executed
             record.failed("File could not be validated against a known schema")
             record.save()
-            os.remove(xml)
+            try:
+                file_failed(xml)
+            except:
+                pass
```
emanuil-tolev: At least log? Swallowing silently can be really difficult to debug, in that it starts with "put logging in place".

portality/tasks/ingestarticles.py
```diff
@@ -71,7 +77,7 @@ def ftp_callback(chunk):
             file_upload.failed("The file at the URL was too large")
             job.add_audit_message("too large")
             try:
-                os.remove(path)
+                os.remove(path) # don't keep this file around
             except:
                 pass
```
emanuil-tolev: Add a log statement here. (I know it's not what you were modifying in this PR, but it is needed for improving our ability to troubleshoot these jobs.)

Or maybe even record the exception in the job as an audit message? That would need a bit more work to do sensibly, if you still want the job to be "success" on file removal exceptions.

emanuil-tolev:
I'll try to add a couple logging statements, then merge. This week ;).

richard-jones:
Did you do this yet?  I'm eager to get this code tested, as there are a lot of xml related questions coming in, and I would rather we got this out sooner rather than holding it up for minor enhancements that won't make much difference. Can someone merge this as-is today please?

PR #1268, opened 2017-04-05
Feature/notes mapping from struct

Steven-Eardley:
Merging as-is should have no effect on the test server - the mappings already exist and the app is running happily. The changes take effect on next clean app initialisation.

We should test the following on the test server before pushing live:

1. test you can search for notes (put a random string in a journal note that's in_doaj, search for it in public interface)
1. run the app under an alias (requires app restart)
1. put in read only mode
1. run the script to copy the index
1. test you now can't search for notes
1. (later) delete the old index

Related to PR https://github.com/CottageLabs/esprit/pull/8 (merged)


emanuil-tolev:
> test you can search for notes (put a random string in a journal note that's in_doaj, search for it in public interface)

http://bit.ly/2oLypqZ - journal-only search on the public search page of testdoaj for `Thestringmostdelectable` . Unfortunately, it works ;).

PR #1274, opened 2017-04-10
[hotfix] separate supervisor configs per host

emanuil-tolev:
Otherwise huey will run on the production app server, rather than the background apps server. Oops.

EDIT:

1. This puts the supervisor configs in the DOAJ repo into subfolders, one per server role.
2. The other side of things is on each server, where /etc/supervisor should be symlinked to the correct role. (e.g. on production apps, `/etc/supervisor -> /home/cloo/repl/production/supervisor`, but on production background apps, `/etc/supervisor -> /home/cloo/repl/production-background/supervisor`).

Test is slightly more complicated, since it needs to take on both `test` and `test-background` configs. They are kept separately in this DOAJ repository. Then, on the doaj-test server, `/etc/supervisor -> /home/cloo/repl/test/supervisor`. This will immediately catch the supervisor config for the web app itself. 

However, the huey configs are being written to `/home/cloo/repl/test-background/supervisor` (according to this very PR). Solution: in the replication repo, symlinked `/home/cloo/repl/test-background/supervisor -> /home/cloo/repl/test/supervisor`.

Now the live application server only runs the web app. The background jobs server only runs huey. The test server runs both.


Steven-Eardley:
@emanuil-tolev WARNING deploying this will briefly kill the app in production. The target of the supervisor symlink has moved, so it won't be able to restart the app.

emanuil-tolev:
> @emanuil-tolev WARNING deploying this will briefly kill the app in production. The target of the supervisor symlink has moved, so it won't be able to restart the app.

supervisor is asked to reread and update all config files before doing a restart on the app. That's in the deploy script. Would it still be a problem?

Steven-Eardley:
It didn't seem to work for test. It showed a broken link:
```
ls -l /etc/supervisor/conf.d/
total 12
lrwxrwxrwx 1 cloo cloo 67 Apr 11 10:12 doaj-test.conf -> /home/cloo/repl/test/doaj/src/doaj/deploy/supervisor/doaj-test.conf
```

emanuil-tolev:
Ah yes. That's because the deploy script is wrong, it turns out. Opening a new PR.

PR #1278, opened 2017-04-12
tweaks to reindex script

emanuil-tolev:


Steven-Eardley:
@emanuil-tolev I was getting bored with editing the script, so I've added argparse. I don't know how cool it is to read the host but not the index from the config, but it gives more flexibility when running the reindex, e.g. moving from ```doaj_v1``` to ```doaj_v2```

emanuil-tolev:
Yeah, this is better. Merge if it looks OK to you too?

PR #1288, opened 2017-05-09
clear editor when changing editor group.  also adds existing editor g…

### LINKED ISSUE #1282: Bulk change of editor group should unassign editor
See background job 7c4faab25a6d4d35bd79a793e9191928

When editor group is changed, when there is an existing editor, the job fails, as the editor is not necessarily in the new group.

Background task should unassign the editor before assigning the new group

richard-jones:
…roups to tests to ensure this behaviour works

Fixes #1282


richard-jones:
The build seems to have failed randomly on an unrelated test - works fine locally

PR #1290, opened 2017-05-10
Hotfix/2017 04 28 jobs fv dates

Steven-Eardley:
## HOTFIX

A few fixes:
* the DAO now saves created date using ```utcnow()``` rather than ```now()``` which means background jobs no longer display that they were updated before being created.
* changed the journal csv writer to use the unicode writer in clcsv, which is more capable.
* Removed sys unicode hack in journalcsv task which was breaking pycharm test runs and hiding code errors.
* converted the languages_iso639_2 dict into unicode strings, because double-encoding them was causing errors, and string comparisons were failing all over the place (```UnicodeWarning```).
* code style improvements.

Subsequently:
* removed unused imports from all unit tests.


doajtest/unit/test_journalcsv.py
```diff
@@ -1,4 +1,4 @@
-import time
+import time, sys
```
emanuil-tolev: This addition doesn't seem to be used in this test. We can just checkout the old version of this file by the look of the diff.

portality/dao.py
```diff
@@ -96,27 +96,22 @@ def save(self, retries=0, back_off_factor=1, differentiate=False, blocking=False
             app.logger.warn("System is in READ-ONLY mode, save command cannot run")
             return
 
-        if 'id' in self.data:
-            id_ = self.data['id'].strip()
```
emanuil-tolev: Not doing this anymore?

Steven-Eardley: I've replaced these five lines:
```python
        if 'id' in self.data:
            id_ = self.data['id'].strip()
        else:
            id_ = self.makeid()
            self.data['id'] = id_
```

The variable ```_id``` isn't used anywhere else in the function (according to searches and PyCharm), just to set it in the object. My change just removes the extra fluff.

Steven-Eardley:
Fixed conflicts on this branch.

portality/dao.py
emanuil-tolev: :+1:

PR #1297, opened 2017-05-16
Hotfix/2017 05 16 reissue dao utc fixes

Steven-Eardley:
## HOTFIX

Replacement for #1290

A few fixes:
* the DAO now saves created date using ```utcnow()``` rather than ```now()``` which means background jobs no longer display that they were updated before being created.
* changed the journal csv writer to use the unicode writer in clcsv, which is more capable.
* Removed sys unicode hack in journalcsv task which was breaking pycharm test runs and hiding code errors.
* converted the languages_iso639_2 dict into unicode strings, because double-encoding them was causing errors, and string comparisons were failing all over the place (```UnicodeWarning```).
* code style improvements.
* removed unused imports from all unit tests.


portality/journalcsv.py
```diff
@@ -42,7 +41,7 @@ def make_journals_csv(file_object):
     issns = cols.keys()
     issns.sort()
 
-    csvwriter = csv.writer(file_object)
+    csvwriter = clcsv.UnicodeWriter(file_object)
```
emanuil-tolev: make a note to check this still works on live after releasing it

emanuil-tolev: though we should have tests.. but real unicode data is the deal

Steven-Eardley: admittedly ```doajtest/unit/test_journalcsv.py``` doesn't have unicode strings, I'll add those before merging.

doajtest/unit/test_journalcsv.py
```diff
@@ -266,5 +254,148 @@ def test_04_background_task(self):
             assert csv_rows[1] != csv_rows[2]             # ID was changed, rows should differ
 
             # check that the difference between rows is the ISSN
+            assert diff(csv_rows[2], csv_rows[1]) == ["8765-4321"]
 
-            assert diff(csv_rows[2], csv_rows[1]) == ["8765-4321"]
\ No newline at end of file
+    def test_05_mixed_unicode(self):
```
emanuil-tolev: I might be mistaken, but I'm not seeking any actual non-ascii data despite seeing some unicode strings. Those may silently convert (successfully) to normal strings without errors where non-ascii strings will not. Try setting the journal's title to `u"Заглавие на журнала"` then running the CSV export and checking for it.

Steven-Eardley: 👌🏻

Steven-Eardley:
screen shot 2017-05-18 at 15 14 52

This is working fine on the test server - all the entries in the background job search after the oldest have times that make sense. The rest of the UI still works, too.

PR #1298, opened 2017-05-16
Enabled day of week in huey cron by passing all args from the config

Steven-Eardley:
I've yet to see the scheduled tasks running in my local environment, so I'm going to leave this PR until I do.


Steven-Eardley:
Got there in the end. Deploying.

Issue #1303, opened 2017-05-18
Bulk editing not working


ClaraDOAJ:
Hello CL,

An email from my colleague. Please, can you help?

Hi Clara,

I never have any success with the bulk editor! 

I have tried to add a note to a group of journals, before then rejecting them. But it has come back with an error. It looks as if I am not allowed to add a note as I am not the editor. I cannot change the editor group via bulk edit as it won't work if an assistant editor is already selected (that was my previous unsuccessful attempt at bulk editing). So I guess I can't reject them either, as I am not the editor.

I am left with having to do them all individually because the bulk editor is effectively useless.

Frustratedly,

here you have the error

Your background job 5e1eb55629cd4b91b4c580f4a329db90 for "suggestion_bulk_edit" has finished with status "error"

You can see the record of your job and its status in the background jobs interface here:

https://doaj.org/admin/background_jobs?source=%7B%22query%22%3A%7B%22query_string%22%3A%7B%22query%22%3A%225e1eb55629cd4b91b4c580f4a329db90%22%2C%22default_operator%22%3A%22OR%22%7D%7D%7D

richard-jones:
This is an odd one - the error is saying that the validation error is that the editor is not in the editor group, but the action is to add a note.

There is no editor_group set during this add_note operation, which would suggest that the Applications themselves are outdated in the database.  Looking at the code and the logs, I am basically certain the background job is not touching the editor or editor_group fields.

The error is correct that the editor "falk" is not in the group "German1".  Was there a point where editor "falk" was in that group, and has subsequently been removed?  I can't see a record of a user called "falk" in the system, perhaps that account was deleted?

ClaraDOAJ:
Hi
I even gave my colleague the "ultra bulk edit" privileges and this did not help either. I think my colleague tried to bulk change the editor group so she was the editor for that group, but she could not as there was an editor already assigned to that group of journals...

mm.. what do you suggest?

Many many thanks!

Clara

Steven-Eardley:
I think the bulk action interface validates the whole record in the same way as the edit form does, even on fields that aren't specifically being changed. So if the record is broken it won't save properly, and would need individual attention.

ClaraDOAJ:
mmm..and how do we know if a record is broken? and why could a record be broken?
Apologies for silly questions...

emanuil-tolev:
[...]

PR #1309, opened 2017-05-23
add migration to blank out nonexistent editor usernames

emanuil-tolev:
#1303 

Data clean. Can be rerun as necessary. **Should** be rerun once we implement a bit in the "delete user" feature which finds all the users' assigned applications and blanks them out of the editor field.


emanuil-tolev:
@richard-jones we can get this on the test server after review, I'll run it. Then hopefully in the afternoon roll out.

PR #1316, opened 2017-06-06
Feature/1231 more bulk edit features

richard-jones:
This adds an edit form to the journals/articles admin page, which allows the overlay of specific metadata fields onto the selected items as a bulk job.


portality/static/doaj/js/available_facetviews/admin.journalarticle.facetview.js
```diff
@@ -1,248 +1,548 @@
...
+            }
+        },
+        validators : {
+            withdraw : journalSelected,
+            reinstate: journalSelected,
+            delete: anySelected,
+            note : function(context) {
+                var valid = journalSelected();
+                if (!valid.valid) {
+                    return valid;
```
Steven-Eardley: This looks weird, but it will end up returning ```{valid: false}``` like the other case.

portality/static/doaj/js/available_facetviews/admin.journalarticle.facetview.js
```diff
@@ -1,248 +1,548 @@
...
+                    if (val !== "") {
+                        found = true;
+                    }
+                }
+                if (!found) {
+                    return {valid: false};
+                }
+
+                // now check for valid field contents
+                // quick and dirty email check - this will be done properly server-side
```
Steven-Eardley: What's the advantage to having a not-'proper' check? The form already has an email validator on it, wouldn't that also give immediate feedback like this?

richard-jones: Well, it depends what you mean by a proper check.  Email validation is surprisingly hard - this is a basic regex approach for the front-end.

The back-end does also do validation when you submit, but it can't do it in real-time while the user types like this one, so this gives us a quick and dirty validation up-front, and we can leave it to the back-end to do proper validation.

Incidentally, I tested how proper WTForms email validation is, and I don't think it's much stronger than this.

emanuil-tolev: I would generally not attempt anything past name@domain.tld . See https://stackoverflow.com/questions/201323/using-a-regular-expression-to-validate-an-email-address or many similar questions. It's best to not attempt to validate email via regex since you always exclude valid addresses, like `foo@emanuil@domain.com`

portality/tasks/journal_bulk_edit.py
```diff
@@ -76,12 +100,29 @@ def run(self):
...
+
+            if "contact_email" in metadata:
+                fc.form.confirm_contact_email.data = metadata["contact_email"]
+
+            for k, v in metadata.iteritems():
+                job.add_audit_message(u"Setting {f} to {x} for journal {y}".format(f=k, x=str(v), y=journal_id))
+                fc.form[k].data = v
+                updated = True
+
+            """
```
emanuil-tolev: just delete this block

emanuil-tolev:
Nothing blocking, though a few potential improvements. For the most part it would be nice to see this on test. I'm deploying it there atm.

portality/static/doaj/js/available_facetviews/admin.journalarticle.facetview.js
richard-jones: yup, that's all this one does.  Checks for an @ and a .

portality/formcontext/forms.py
```diff
@@ -528,3 +528,32 @@ class ReadOnlyJournalForm(JournalInformation, JournalSubject, JournalLegacy, Not
...
+        choices=[("", "Leave unchanged"), ("True", "Yes"), ("False", "No")],
+    )
+    country = DOAJSelectField('Country',
+        [validators.Optional()],
+        description='Select the country where the publisher carries out its business activities.',
+        choices=Choices.country(),
+    )
+    owner = StringField('Owner',
+        [ReservedUsernames(), validators.Optional()],
+        description='DOAJ account to which the application belongs.'
```
emanuil-tolev: Hmm, I did not see that. @richard-jones can you change it (+ the kwargs thing above), or I can pick it up?

PR #1319, opened 2017-06-09
[hotfix] add useful gitignores

emanuil-tolev:
# hotfix, merge with git flow


Steven-Eardley:
I've merged and released this to production, not that it makes a difference to the running app.

Issue #1313, opened 2017-05-24
IJCH problems with API


ClaraDOAJ:
IJCH is currently running on PJS 3.0.2 and there is an inbuilt plugin of DOAJ with API. I generated a new key for them but they are still having problems.

See error:

Deposit was not successful! The DOAJ API returned an error: '1 - {"status": "unauthorised", "error": "An API Key is required to access this. (ref: 6885413c-3fc1-11e7-aa44-04015bfb3501)"}'.
 
Notification
Deposit was not successful! The DOAJ API returned an error: '1 - 
301 Moved Permanently

nginx/1.4.6 (Ubuntu)

richard-jones:
@ClaraDOAJ Do you know what operation they were trying to do when this happened?

richard-jones:
Here's a few things to check to make sure that they should be able to access the API:

* That they have an API key set in their user account (which you have done), and that they are using the right API key to access the API
* That they have the "api" role set against their user account
* That their system is using the right field when sending the api key: "api_key"

Can you give us the account name, and I'll check their user account setup.

ClaraDOAJ:
09717587

ClaraDOAJ:
More detail;:
![image](https://cloud.githubusercontent.com/assets/22593616/26732379/b331f4f2-47b7-11e7-807f-7dfec87633c0.png)


![image](https://cloud.githubusercontent.com/assets/22593616/26732355/a24d32aa-47b7-11e7-9045-4c91158ee6f2.png)

![image](https://cloud.githubusercontent.com/assets/22593616/26732367/ad39c494-47b7-11e7-8e4a-781f8bbeef24.png)

emanuil-tolev:
I think some of the developers behind a technical library used by the OJS platform have contacted me via github. The original error at the top of this issue might not be relevant anymore - we shall see. I'm just recording it here since it looks like it might be relevant to this issue.

This is the thread (it is very technical): https://github.com/pkp/pkp-lib/issues/2543#issuecomment-304570146

They are currently (at the time of writing) getting an unexplained 502 bad gateway error despite the website being functional - it only happens for their specific API request. I need to investigate that further.

emanuil-tolev:
A potential fix that cuts processing time in half is now rolled out to live. I need to get back to those folks. I'll look at increasing the timeout slightly as well, to make the errors they are experiencing less likely.

emanuil-tolev:
[...]

PR #1320, opened 2017-06-13
[hotfix] Do not rerun article creation twice. Use the results of the dry run.

emanuil-tolev:
For #1313 

We used to do a dry run, then redo the entire article creation process. We thought a 1000 x2 would not be a big problem at the time.

We also only avoided processing duplicates during the real run.

Now:

1. we avoid duplicates immediately. The code has been refactored to be more defensive. This might well need tests to prove it is more defensive against bad data and you should ask for such tests if they are not there.

2. we accumulate the results of the dry run as we go, so we still don't save anything until 100% of the articles have passed validation

3. we save the results of the dry run. This is based on the fact that the dry run currently only consists of:

```python
# ArticlesCrudApi.create , portality/api/v1/crud/articles.py:97
        if not dry_run:
            am.save()
```

There are no other uses of `dry_run` in this method.

Of course, this may change in the future without notice to the Bulk API classes and the new dependency I'm introducing is not obvious. We can leave a comment that there is a dependency, or make the dry run functionality more explicit. For example, create a method on ArticlesCrudApi like `create_not_dry_run` and call that inside `create` as well as outside of it (as I'm doing in this PR). That's quite ugly, but you get the ultimate goal - express the dependency in code and make sure it's easily (or 0 effort) to maintain.

### DIFF
--- a/portality/api/v1/bulk/articles.py
+++ b/portality/api/v1/bulk/articles.py
@@ -29,8 +29,10 @@ class ArticlesBulkApi(Api):
 
     @classmethod
     def create(cls, articles, account):
-        # we run through create twice, once as a dry-run and the second time
-        # as the real deal
+        # We run through the articles once, validating in dry-run mode
+        # and deduplicating as we go. Then we .save() everything once
+        # we know all incoming articles are valid.
+
         new_articles = []
 
         ids = []
--- a/portality/api/v1/bulk/articles.py
+++ b/portality/api/v1/bulk/articles.py
@@ -31,9 +31,7 @@ class ArticlesBulkApi(Api):
     def create(cls, articles, account):
         # we run through create twice, once as a dry-run and the second time
         # as the real deal
-        for a in articles:
-            ArticlesCrudApi.create(a, account, dry_run=True)
-        # no formatting issues, the JSON of each article complies with required format at this point
+        new_articles = []
 
         ids = []
         dois_seen = set()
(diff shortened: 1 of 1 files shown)


portality/api/v1/bulk/articles.py
```diff
@@ -31,39 +31,43 @@ def create_swag(cls):
     def create(cls, articles, account):
         # we run through create twice, once as a dry-run and the second time
```
Steven-Eardley: This comment is now out of date / misleading

portality/api/v1/bulk/articles.py
```diff
@@ -31,39 +31,43 @@ def create_swag(cls):
...
+                        if l['url'] in fulltext_urls_seen:
+                            duplicate = True
+                            break
+                        fulltext_urls_seen.add(l['url'])
 
             if duplicate:
                 continue
 
-            n = ArticlesCrudApi.create(a, account)
+            n = ArticlesCrudApi.create(a, account, dry_run=True)
```
Steven-Eardley: Have you profiled the code again - do you have any numbers to show it's faster when only running the task once?

(It obviously will be faster, I'm just wondering if it'll be enough to consider the issue fixed)

emanuil-tolev: I did locally, though I'm afraid I don't have the setup to easily re-try this at the moment. It was exactly 2x faster (relative to the same machine). That means, I believe, 24s instead of 48s on the machine I tried it on.

emanuil-tolev: So actually the answer is that I don't have numbers :(, I didn't keep any copy of the results!

emanuil-tolev:
The test `test_02_create_duplicate_articles_success` in `doajtest/unit/test_api_bulk_article.py` actually satisfies 

> we avoid duplicates immediately. The code has been refactored to be more defensive. This might well need tests to prove it is more defensive against bad data and you should ask for such tests if they are not there.

emanuil-tolev:
This should be ready to merge now (at least to go to test). I'll merge it as soon as there's a successful build.

emanuil-tolev:
@Steven-Eardley could you merge this? Didn't have access to a laptop so I could use git flow by the time the build passed

Steven-Eardley:
This is on the test server just now. Use ```fab deploy_live:tag=2017-06-13_1313_speed_up_bulk_create_api``` to deploy to live when ready.

PR #1323, opened 2017-06-23
detect character encoding errors in setspecs and report these as bada…

### LINKED ISSUE #684: Better crawl error response for the OAI endpoint
Some errors for OAI produce a 502 error. We should have a better response.

https://www.google.com/webmasters/tools/crawl-errors?hl=en&siteUrl=http%3A%2F%2Fdoaj.org%2F&pli=1&authuser=1

and

https://www.google.com/webmasters/tools/crawl-errors?hl=en&siteUrl=http%3A%2F%2Fdoaj.org%2F&pli=1&authuser=1#t1=0&t2=1

richard-jones:
…rguments

fixes #684

### DIFF
--- a/portality/view/oaipmh.py
+++ b/portality/view/oaipmh.py
@@ -110,7 +110,24 @@ def make_set_spec(setspec):
     return base64.urlsafe_b64encode(setspec).replace("=", "~")
 
 def decode_set_spec(setspec):
-    return base64.urlsafe_b64decode(str(setspec).replace("~", "="))
+    # first, make sure the setspec is a string
+    try:
+        setspec = setspec.encode("utf-8")
+    except:
+        raise SetSpecException()
+
+    # switch the ~ for =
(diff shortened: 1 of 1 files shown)


portality/view/oaipmh.py
```diff
@@ -131,6 +148,9 @@ def make_resumption_token(metadata_prefix=None, from_date=None, until_date=None,
 class ResumptionTokenException(Exception):
     pass
 
+class SetSpecException(Exception):
+    pass
```
Steven-Eardley: Should we at least log these?

richard-jones: I don't see a particular need to - all we'll be learning is that user's have fluffed their inputs, it doesn't really affect us.  They are not errors on our side - I mostly use these for message passing in the stack to affect the behaviour higher up.

Steven-Eardley:
I see, that makes sense.

Issue #1293, opened 2017-05-16
Associate editors privileges


ClaraDOAJ:
Hello CL

We have detected that associate editors are able to change the status of an "accepted" application to "completed". This does not make any sense in terms of workflow but it is also dangerous for our database.

Associate editors should only be able to go from "pending" to "in progress" and "completed".

Thanks!

Steven-Eardley:
@ClaraDOAJ there are few things I'd like to clear up before I submit my fixes.

Firstly, there's already code in place on each role's application edit form which prevents changes when an application is "accepted" - including admin. Can you confirm you're definitely seeing associate editors changing it successfully, and not seeing this message:
```You cannot edit applications which have been accepted into DOAJ.```

Secondly, do you only want to limit changes to the application status, e.g. prevent associate editors from changing 'on hold' to 'in progress', or do you want to prevent them from editing any aspect of the application - for example, preventing associate editors from changing a contact email address on an application that's in status 'ready'. They'd not be able to make any changes, instead see a message like:
```You cannot edit applications which are in status ready.```

Finally, do / should associate editors handle reapplications? At the moment there is a distinction in the admin form between pending applications and pending reapplications, which has been hidden from other users. If I lock down associate editor privileges to the 3 statuses above they will no longer be able to edit pending reapplications. Was this intended?

From last time I changed this, there's a full list of editorial status permissions in issue https://github.com/DOAJ/doaj/issues/801:

**reapplication pending** (visible to admin and publishers)
**reapplication submitted** (visible to admin and publishers)
**pending** (visible to asseds, editors, admin)
**in progress** (visible to asseds, editors, admin)
**completed** (visible to asseds, editors, admin)
**ready** (visible to editors and admin)
**accepted** (visible to admin)
**rejected** (visible to admin)
**on hold** (visible to admin)

ClaraDOAJ:
reapplication pending to be deleted

emanuil-tolev:
> reapplication pending to be deleted

we should probably do that in a separate issue though. But still:

> If I lock down associate editor privileges to the 3 statuses above they will no longer be able to edit pending reapplications. Was this intended?

Basically, yes. It's ok if they can't handle reapp pending because that's no longer used as a status.

greboun:
As discussed in the weekly meeting a write up of all privileges
Privileges as follows
ass editors  can see  reapplication submitted ,pending, in progress, completed   can change reapplication submitted to pending to in progress to completed   CANNOT change completed
[...]

PR #1324, opened 2017-06-28
Hotfix/2017 06 10 iss1293 associate privileges

Steven-Eardley:
* Editor and associate editor forms throw errors when application is trying to be saved in a status outwith their editorial permissions, or backwards in the editorial pipeline.
* Editors and associate editors only see statuses that they can change to on the application edit form. Statuses beyond their permitted are visible but greyed out.

In light of the second point, the FormContextExceptions are a little redundant, because users don't get access to the statuses for the UI, but this is more complete (since the UI can be bypassed). The tests reflect this.

This may need to go on the test server so I can prove that some of my weirder code works as intended.
#1293

### DIFF
--- a/doajtest/unit/test_assed_app_review.py
+++ b/doajtest/unit/test_assed_app_review.py
@@ -140,7 +140,7 @@ APPLICATION_SOURCE = {
         "article_metadata" : True
     },
     "admin" : {
-        "application_status" : "reapplication",
+        "application_status" : "pending",
         "notes" : [
             {"note" : "First Note", "date" : "2014-05-21T14:02:45Z"},
             {"note" : "Second Note", "date" : "2014-05-22T00:00:00Z"}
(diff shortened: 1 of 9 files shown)


emanuil-tolev:
> It should also, as an enhancement to the current functionality, disable the save buttons when the user is not allowed to edit the item

Is this worth a red flash message at the top of the page too?

emanuil-tolev:
> It should also, as an enhancement to the current functionality, disable the save buttons when the user is not allowed to edit the item
> Is this worth a red flash message at the top of the page too?

Oh that is what you are suggesting

> We also need to display some text when the user is not allowed to edit an item: status accepted and rejected.

, ignore me.

Steven-Eardley:
@emanuil-tolev ready for another review - the save buttons are disabled, and a warming is shown when editors can't edit.

Steven-Eardley:
This is merged to production, but not deployed via fabric. That's ```git flow```'s fault.

PR #1331, opened 2017-07-04
[hotfix] remove certain types from remote sync task - they do not exist any longer

emanuil-tolev:
just housekeeping


emanuil-tolev:
yes, we should! feel free to commit to this branch

PR #1337, opened 2017-07-06
[hotfix] optimise sponsor images

emanuil-tolev:
Reduce size of images on homepage by a factor of 3. This is mostly thanks to some truly enormous sponsor logos that had made it in, like one was the size I once used to print a tablecloth.

Original size of images on homepage:

pr-screen-size-orig

New size of images on homepage:

pr-screen-size-optimised

It's probably worth scaling new sponsor images down to a maximum of 300 pixels width or height, whichever is greater. It's also best to let the image processing program (GIMP for me) to scale the other attribute accordingly, preserving perspective.

If a particular file format is quite large, it's worth trying another one. Gifs can get very small but will alter the quality of the logo if it is complex or uses dithering (hey 90s). JPGs can be much smaller than PNGs for certain images and vice versa - it's often fastest to just try.


emanuil-tolev:
One reason for doing this is poor connections. I recently had the chance to open DOAJ on a poor connection, and the usually blazing fast home page made the browser struggle a lot, despite async image loading. 1 MB of pictures is also just a very lot to load up considering where over half our traffic comes from (developing countries with high latency to UK servers).

Issue #1328, opened 2017-07-03
Bulk edit- error 3


ClaraDOAJ:
I wanted to give Hindawi journals the seal but it didn’t work
 
journal_bulk_edit by sonja status: error
Job ID: c31ef35247c6457494085ecdb0de51ec
Job Created: 19 Jun 2017 at 11:14
Job Last Updated: 19 Jun 2017 at 11:14
 
More Information
Parameters:
[...]

Issue #1327, opened 2017-07-03
Test bulk edit - error 2


ClaraDOAJ:
2. Selecting 224 Redalyc journals for the Seal
Your bulk edit request has been submitted and queued for execution.
224 journals have been queued for edit.
You can see your request here in the background jobs interface (opens new tab).
You will get an email when your request has been processed; this could take anything from a few minutes to a few hours.
dismiss
 
journal_bulk_edit by alejandra status: error
Job ID: ed403102101b427a895686c064a6d7e9
Job Created: 19 Jun 2017 at 14:58
Job Last Updated: 19 Jun 2017 at 14:59
 
More Information
Parameters:
[...]

Issue #1326, opened 2017-07-03
Testing Bulk edit - error 1


ClaraDOAJ:
1. Changing Editora da Pontifícia Universidade Católica do Rio Grande do Sul (EDIPUCRS) to  Pontifícia Universidade Católica do Rio Grande do Sul
 
Your bulk edit request has been submitted and queued for execution.
14 journals have been queued for edit.
You can see your request here in the background jobs interface (opens new tab).
You will get an email when your request has been processed; this could take anything from a few minutes to a few hours.
dismiss
 
In background jobs:
 
journal_bulk_edit by alejandra status: error
Job ID: 0ac31ce0bcc04510b40bc09609825da8
Job Created: 19 Jun 2017 at 14:22
Job Last Updated: 19 Jun 2017 at 14:22
 
More Information
Parameters:
journal_bulk_edit__ids -- 1f35bd71581447da995ff1d7c12d0d7d, a28b5f47bc8f4159a6dbbb7d350544f7, 7fbd75aa5dec49348b836d3aafbf0dd1, cc05ae5bb359489aad6cbf6eefd1bd5d, 6f3540f4f2254d7d909ea4b3accbce7c, 0402c3d950174e498b777e88a7c7f52b, 239e9294416745bca042e2eb5f0d5a53, 20df01213f724eaa99b3fba458cc82d2, 6db9ddd154b046c9bbe1a28aad76bdbc, f4e54ac7705b48cd83430171ff172430, 5a2cffb30b47496db7f335ae38687dee, c9b3211beec54e8b92c8ce201ac1892e, 580b1b52c0014fac9d3b5313ae0bc54b, ac84b7487b69464491ce6d973053a5c7
journal_bulk_edit__replacement_metadata -- {"publisher": "Pontif\u00edcia Universidade Cat\u00f3lica do Rio Grande do Sul"}
Reference:
journal_bulk_edit__selection_query -- {"query": {"filtered": {"filter": {"bool": {"must": [{"term": {"index.publisher.exact": "Editora da Pontif\u00edcia Universidade Cat\u00f3lica do Rio Grande do Sul (EDIPUCRS)"}}, {"term": {"_type": "journal"}}]}}, "query": {"query_string": {"query": "Pontif\u00edcia Universidade Cat\u00f3lica", "default_operator": "AND"}}}}}
Audit Messages:
2017-06-19T14:22:12.150301Z -- Job Started
2017-06-19T14:22:12.178299Z -- Error in Job Run
[...]

Issue #1138, opened 2016-11-28
Article upload as asynchronous task


richard-jones:
* Refactor the upload handling code into a background task.  This may be best as a single task or as two distinct tasks, as the jobs it needs to handle are:
    * dereference any external files, provided by reference
    * validate and process uploaded or dereferenced files
* Migrate the existing file upload jobs to background jobs, for historical record, then remove the file uploads index.  Depending on exactly how we construct the jobs (whether they are one or two stage), we may not do this - question is whether the publisher UI can be built on the jobs index
* Refactor the publisher UI to point to the background jobs index, if required (see above)
* Refactor the way that file upload jobs are triggered
* Take into account the S3 file system, and its synchronisation lag/other limitations

richard-jones:
done

emanuil-tolev:
> Take into account the S3 file system, and its synchronisation lag/other limitations

Unfortunately this doesn't seem to work. I hooked testdoaj to my local ES and Redis (via ngrok Pro) and restarted the web app, huey main queue and huey long running. I then proceeded to run the doaj web app locally as well, but *not* running any huey locally. What should have happened:

1. I upload a file to localhost:5004. This creates an ES job and a Huey Redis job.
2. Huey tries to execute the job on doajtest.
3a. It succeeds if the file is already there.
3b. If it fails, it should add an audit message "failed to find blah.xml".
4. Once the retries are exausted, the job status should be "error". Instead, it is "processing". The status of the XML import displayed to the user (on the XML upload page) should also change to some appropriate system error. Instead, it is "pending".

Finally, at step 3 it took course 3b, i.e. it failed. The retries don't have a lot of waiting time (I suppose that's just the default 10s delay without any back-off?). As far as I can see the file did actually appear on test, but I assume it did so a little too slowly. I propose to increase the scheduling delay for this particular task to 60s (so it only starts executing 60s after the user first uploads the file, then on any retry the job is only picked up again after 60s - up to 3 times). I'll make and test that change since @richard-jones is probably busy with other things and I have the setup locally.

```
[2017-03-07 18:43:41,469] INFO:huey.consumer.Worker:Worker-1:Adding queuecmd_ingest_articles: db426589-9ed5-47f1-981a-6db3ca2776ed @2017-03-07 18:43:47 3 retries to schedule
[...]

PR #1340, opened 2017-07-09
Feature/1326 bulk edit review

richard-jones:
This contains some fixes and enhancements associated with the testing for the bulk edit feature.  Issues addressed are:

#1326 #1327 #1328 and #1138 

This should be deployed on the test server again for @ClaraDOAJ and the team to check the changes fix the issues they have, then the bulk editing code should be deployed to live.

### DIFF
--- a/doajtest/unit/test_ingestarticles.py
+++ b/doajtest/unit/test_ingestarticles.py
@@ -103,7 +103,7 @@ class TestArticleUpload(DoajTestCase):
         self.ftp = ftplib.FTP
 
         self.upload_dir = app.config["UPLOAD_DIR"]
-        self.ingest_articles_retries = app.config.get("HUEY_TASKS").get("ingest_articles").get("retries")
+        self.ingest_articles_retries = app.config['HUEY_TASKS']['ingest_articles']['retries']
 
     def tearDown(self):
         super(TestArticleUpload, self).tearDown()
--- a/doajtest/unit/test_ingestarticles.py
+++ b/doajtest/unit/test_ingestarticles.py
@@ -103,6 +103,7 @@ class TestArticleUpload(DoajTestCase):
         self.ftp = ftplib.FTP
 
         self.upload_dir = app.config["UPLOAD_DIR"]
+        self.ingest_articles_retries = app.config.get("HUEY_TASKS").get("ingest_articles").get("retries")
 
     def tearDown(self):
         super(TestArticleUpload, self).tearDown()
@@ -114,6 +115,7 @@ class TestArticleUpload(DoajTestCase):
         ftplib.FTP = self.ftp
 
         app.config["UPLOAD_DIR"] = self.upload_dir
+        app.config["HUEY_TASKS"]["ingest_articles"]["retries"] = self.ingest_articles_retries
 
         for id in self.cleanup_ids:
             path = os.path.join(app.config.get("UPLOAD_DIR", "."), id + ".xml")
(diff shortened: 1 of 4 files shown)


doajtest/unit/test_ingestarticles.py
```diff
@@ -103,6 +103,7 @@ def setUp(self):
         self.ftp = ftplib.FTP
 
         self.upload_dir = app.config["UPLOAD_DIR"]
+        self.ingest_articles_retries = app.config.get("HUEY_TASKS").get("ingest_articles").get("retries")
```
emanuil-tolev: `AttributeError: 'NoneType' object has no attribute 'get'` if HUEY_TASKS is not defined.

Since this is a test, I would suggest just using `app.config['HUEY_TASKS']['ingest_articles']['retries']`. The test doesn't seem to make sense without this value being defined.

doajtest/unit/test_task_journal_bulkedit.py
```diff
@@ -167,7 +169,7 @@ def test_05_edit_metadata(self):
                                  doaj_seal=True,
                                  country="AF",
                                  owner="test1",
-                                 platform="my platform",
+                                 platform=u"my platfo®m",   # stick in a weird character for good measure
```
emanuil-tolev: This should be okay here .. in general characters from other alphabets will provide a higher chance of breaking unicode decoding and encoding.

```
>>> u"®".encode('ascii')
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\xae' in position 0: ordinal not in range(128)
>>> u"®".encode('latin-1')
'\xae'  # <-------- it worked with latin-1, a common web browser default
>>> u"асд".encode('latin-1')
Traceback (most recent call last):  # <------------ no such luck with Cyrillic
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 0-2: ordinal not in range(256)
```

Googling "some cyrillic" and "some chinese" yields quick results. If in doubt, you can always refer to http://unicodesnowmanforyou.com/ , which I've only managed to encode in utf-8 from the common codecs.

portality/tasks/ingestarticles.py
```diff
@@ -240,6 +240,14 @@ def _process(self, file_upload):
 
         if not os.path.exists(path):
             job.add_audit_message(u"File not found at path {} . Retrying job later.".format(path))
+            count = self.get_param(job.params, "attempts")
+            retry_limit = app.config.get("HUEY_TASKS", {}).get("ingest_articles", {}).get("retries")
```
emanuil-tolev: `.get("retries")` => `.get("retries", 0)`?

emanuil-tolev:
should be on test server shortly to try out

richard-jones:
The code governing this process hasn't changed since March 2015, so it's not a problem with the OAI endpoint itself.  You can clearly see, though, that languages are being omitted from records in the oai_doaj format:

https://doaj.org/oai.article?verb=GetRecord&metadataPrefix=oai_dc&identifier=oai:doaj.org/article:f9a2ca95cccd4533b23a5f5089fdd565

https://doaj.org/oai.article?verb=GetRecord&metadataPrefix=oai_doaj&identifier=oai:doaj.org/article:f9a2ca95cccd4533b23a5f5089fdd565

The user thinks that this used to work, but no longer does.  I've looked at the data, and it appears to be correct, and I've reproduced this locally in a dev environment.  That would suggest to me that this may never have worked!  I'll dig deeper, anyway....

richard-jones:
I've tracked this down to an error in the crosswalk from country name to 3 letter code - this only works if there is a country name, and if there's a code instead it omits it.  I can put in a quick fix...

PR #1348, opened 2017-07-20
[hotfix] make sure index is initialised even when running via gunicorn

emanuil-tolev:
#1341 fix

(don't deploy yet, moving buckets on live)

### DIFF
--- a/portality/app.py
+++ b/portality/app.py
@@ -52,6 +52,12 @@ app.register_blueprint(openurl)
 app.register_blueprint(atom)
 app.register_blueprint(doaj)
 
+# initialise the index - don't put into if __name__ == '__main__' block,
+# because that does not run if gunicorn is loading the app, as opposed
+# to the app being run directly by python portality/app.py
+# putting it here ensures it will run under any web server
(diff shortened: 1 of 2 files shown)


Steven-Eardley:
Yeah, well because I merged my associate privileges pull request it'll be a pain release this anyway...

emanuil-tolev:
> Yeah, well because I merged my associate privileges pull request it'll be a pain to release this anyway...

Hah, yes, that's why I was thinking of just merging that to develop.

emanuil-tolev:
Actually, I wonder whether we should revert the associate privileges merge commit from production. It will remain on develop, and you've given instructions to test it. Then, when it's time to deploy it live, we just revert the revert commit and it will be back in again.

That would enable us to deploy this ^ and any other hotfixes necessary in the meantime.

Steven-Eardley:
Okay, it took me a while and learning about ```git gc```, but I managed to merge this and revert my boo-boo on ```production```.

Steven-Eardley:
Deployed to live.

Issue #1296, opened 2017-05-16
Write global article duplication script for on-demand execution


richard-jones:
We occasionally want to look at the state of the DOAJ to ensure that no article duplicates have emerged in the system.  We could do with a script which does the following:

* For each article does a global check to see if there are any other articles which are "duplicate"
* Outputs each "set" of duplicate articles, along with the criteria for which they were determined duplicate
* Further identifies articles which are duplicate within a single user's account
* Identifies articles which appear in more than one "set" - that is, they may be duplicate with one set of articles by DOI, and duplicate with a different set by URL, etc

This may require enhancements to the article duplication check function to return more information about the nature of the check.

dommitchell:
Where did we get with this one? What was the outcome?

Steven-Eardley:
This is still on my to-do list. I started this a while ago and shall prioritise finishing it off and reporting back.

ClaraDOAJ:
Hello Steve,

I am attaching an example of this again. Unfortunately you will not be able to access the duplication now as I have deleted one of the copies.

![image](https://user-images.githubusercontent.com/22593616/32104553-3eddab00-bb25-11e7-9cbe-7e3ecbfba6ee.png)

dommitchell:
What's the status of this one @Steven-Eardley ?

Steven-Eardley:
The script is currently running on the test server (report only), and I believe it's producing good results. Since the data is shared, we can see some of the duplicates in the live system:

Correct duplicate match:
```
e3d97accbd084418b354909f366aae37
	6 DOI duplicates: 444bf649032547b2b2eae62c22538cb2, 4b5f13de046a48cfbaad620f543fb2c9, 6807a1d8c55c4c3ab953e41cfd2adefc, 72d55082bd5e47fa859369e91c22311d, a43e1ef27fc547e4904d3711ef20d4af, e2102ab58a2749ab8feef9d8a62375e8
	6 fulltext duplicates: 444bf649032547b2b2eae62c22538cb2, 4b5f13de046a48cfbaad620f543fb2c9, 6807a1d8c55c4c3ab953e41cfd2adefc, 72d55082bd5e47fa859369e91c22311d, a43e1ef27fc547e4904d3711ef20d4af, e2102ab58a2749ab8feef9d8a62375e8
```
[...]

PR #1373, opened 2017-09-06
Hotfix/2017 06 20 iss1296 article dedupe

Steven-Eardley:
This is the first half of the article de-duplication work, covering improvements to the tests and a refactor of fetching duplicates.

This doesn't extend into the script required for #1296. It'll be easier to check that one based on this one.

### DIFF
--- a/portality/scripts/remove_duplicate_articles.py
+++ b/portality/scripts/remove_duplicate_articles.py
@@ -42,10 +42,14 @@ if __name__ == "__main__":
     dry_run = args.dry_run
 
     start = datetime.now()
-    count = 0
-    print 'Starting {0} {snapshot} in 5 seconds, Ctrl+C to exit.'.format(start.isoformat(), snapshot=snapshot_report)
+
+    if dry_run:
+        print 'Starting {0} dry run in 5 seconds, Ctrl+C to exit.'.format(start.isoformat())
+    else:
+        print 'Starting {0} {snapshot} in 5 seconds, Ctrl+C to exit.'.format(start.isoformat(), snapshot=snapshot_report)
     time.sleep(5)
 
+    count = 0
     originals = []  # make sure we only delete FORWARD. So, if articles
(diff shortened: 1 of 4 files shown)


portality/scripts/remove_duplicate_articles.py
```diff
@@ -46,15 +61,19 @@
         if duplicates:
```
richard-jones: This line: XWalk().get_duplicate(a, all_duplicates=True)

all_duplicates is no longer part of the signature for that method, you want XWalk().get_duplicates() instead

portality/scripts/remove_duplicate_articles.py
```diff
@@ -30,13 +39,19 @@
...
+        print 'Starting {0} {snapshot} in 5 seconds, Ctrl+C to exit.'.format(start.isoformat(), snapshot=snapshot_report)
     time.sleep(5)
 
+    count = 0
     originals = []  # make sure we only delete FORWARD. So, if articles
-    # A, B and C are duplicates, we stumble upon A first, and delete B and C.
+    # A, B and C are duplicates, we stumble upon A first, and delete B and C.cat
     # We must then ensure we do not delete A if we chance upon B or C
     # (because they are still in Python memory even though they are not in ES).
     article_iterator = models.Article.iterall(page_size=5000)
```
richard-jones: I seem to remember something about the order that things get considered.  I presume we want A to be the actual one to keep (e.g. the most recent, or the oldest, whatever the rule is).  iterall doesn't inherently order anything, I don't think, or if it does it is by ID.  So your ordering here is not guaranteed.

Steven-Eardley: This script has been completely rewritten in the part 2 pull request #1387 - it doesn't use an iterator any more, but a scroll. I should've really made this PR just to cover the changes in ```article.py```

Steven-Eardley:
Since the run script is almost unrecognisable in the next PR, I'm going to count your review of this one as OK!

PR #1375, opened 2017-09-06
WIP on script to manage ES better

emanuil-tolev:
# DO NOT MERGE, WORK IN PROGRESS


richard-jones:
I don't fully understand the details, but the code looks good.

One question - this looks like something that you have to run explicitly to do the job, right?  Could it run continually, or every 15 minutes or so, and restart nodes if needed?

emanuil-tolev:
Not quite finished unfortunately! And yes it polls.

PR #1380, opened 2017-09-12
HOTFIX fix bug with role detection in status reports, causing some users not…

richard-jones:
… to be counted

### DIFF
--- a/doajtest/unit/test_reporting.py
+++ b/doajtest/unit/test_reporting.py
@@ -182,3 +182,64 @@ class TestReporting(DoajTestCase):
         assert len(prov_outfiles) == 4
         assert len(cont_outfiles) == 1
 
+    def test_05_status_role(self):
+        counter = reporting.StatusCounter("month")
+
+        roles = ["admin"]
+        best = counter._get_best_role(roles)
+        assert best == "admin"
+
+        roles = ["associate_editor", "editor", "admin"]
+        best = counter._get_best_role(roles)
+        assert best == "admin"
+
+        roles = ["api"]
+        best = counter._get_best_role(roles)
+        assert best is None
+
+        roles = ["api", "whatever", "another"]
+        best = counter._get_best_role(roles)
+        assert best is None
+
+        roles = ["associate_editor", "api", "editor"]
+        best = counter._get_best_role(roles)
+        assert best == "editor"
+
+        roles = ["api", "associate_editor", "editor"]
+        best = counter._get_best_role(roles)
+        assert best == "editor"
+
+    def test_06_status_is_countable(self):
+        counter = reporting.StatusCounter("month")
+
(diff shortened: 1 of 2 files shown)


doajtest/unit/test_reporting.py
```diff
@@ -182,3 +182,64 @@ def test_04_background(self):
...
+        assert best is None
+
+        roles = ["associate_editor", "api", "editor"]
+        best = counter._get_best_role(roles)
+        assert best == "editor"
+
+        roles = ["api", "associate_editor", "editor"]
+        best = counter._get_best_role(roles)
+        assert best == "editor"
+
```
Steven-Eardley: what if ```roles = []```?

portality/tasks/reporting.py
```diff
@@ -173,24 +173,8 @@ def count(self, prov):
...
-
-        countable = False
-        if best_role == "admin" and (prov.action == "status:accepted" or prov.action == "status:rejected"):
-            countable = True
-        elif best_role == "editor" and prov.action == "status:ready":
-            countable = True
-        elif best_role == "associate_editor" and prov.action == "status:completed":
-            countable = True
+        best_role = self._get_best_role(prov.roles)
+        countable = self._is_countable(prov, best_role)
```
Steven-Eardley: you'd be sending ```None``` through ```_is_countable()``` if the user doesn't have a best role. But a user should have a best role, right, since they're editing things. I guess this is a niche case.

portality/tasks/reporting.py
```diff
@@ -209,6 +193,35 @@ def count(self, prov):
...
+        countable = False
+
+        if role == "admin" and (prov.action == "status:accepted" or prov.action == "status:rejected"):
+            countable = True
+        elif role == "editor" and prov.action == "status:ready":
+            countable = True
+        elif role == "associate_editor" and prov.action == "status:completed":
+            countable = True
+
+        return countable
```
Steven-Eardley: Could be simplified as

```
if ...
   return True
elif ...
   return True
else:
    return False
```

but that's inconsequential, really.

doajtest/unit/test_reporting.py
Steven-Eardley: ```best``` is ```None```, I suppose. No surprises there. Probably a needless test case.

portality/tasks/reporting.py
emanuil-tolev: Well, that's a bit of code style I'd say. What you're suggesting does fit with general python recommended style. Makes code easier to read (when applied 100 or 1000 times), so it matters slightly.

There are some style guides which might be useful to consider for the refactor actually - provided they fit us as a team.

Steven-Eardley:
Merged and deployed

PR #1387, opened 2017-09-19
Hotfix/2017 08 30 iss1296 article dedupe report

Steven-Eardley:
After quite a lot of faff. I recommend reviewing #1373 first to break up the changes and make this less of an ordeal.


portality/scripts/article_duplicates_report_remove.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def duplicates_per_account(connection, delete, snapshot):
+    dupcount = 0
+    delcount = 0
+
+    for acc in esprit.tasks.scroll(connection, 'account'):
+        account = Account(**acc)
+
+        js = Journal.all(terms={"admin.owner": [account.id]})
```
richard-jones: Probably needs to be admin.owner.exact.  Might also be worth pushing these behaviours into the Journal object.  See Journal.issns_by_owner for a possible template for this.

portality/scripts/article_duplicates_report_remove.py
```diff
@@ -0,0 +1,139 @@
...
+                                    "must": [
+                                        {"terms": {"index.issn.exact": issns}}
+                                     ]
+                                }
+                            },
+                            "query": {"match_all": {}}
+                        }
+                    },
+                    "sort": [{"last_updated": {"order": "desc"}}]
+                }
```
richard-jones: In fact, I think to get the issns you are after, Journal.issns_by_owner is the function you need here.

portality/scripts/article_duplicates_report_remove.py
```diff
@@ -0,0 +1,139 @@
...
+
+            for j in js:
+                issns = j.bibjson().issns()
+                scroll_query = {
+                    "query": {
+                        "filtered": {
+                            "filter": {
+                                "bool": {
+                                    "must": [
+                                        {"terms": {"index.issn.exact": issns}}
```
richard-jones: One thing to remember here is that terms queries have a finite number of terms you can put in there.  It might be configurable, but I remember it being up around 1024, and we've been caught by this before in deduplication.  You'll need to check the number of issns in your list, and maybe break this into several queries.

portality/scripts/article_duplicates_report_remove.py
Steven-Eardley: That's right, if I'm just using the ISSNs to seed my query, it's fewer loops to get them per account rather than per journal.

Steven-Eardley:
Thanks @richard-jones, changes added.

PR #1389, opened 2017-09-20
added a couple more dates.today()s to defaults

Steven-Eardley:
Found a couple more defaults that should be ```dates.today()``` rather than ```dates.now()``` to keep pesky symbols out of filenames, because not everyone uses a superior operating system 🐧

## HOTFIX


emanuil-tolev:
We should add docs/code/python-style.md and add this in.

Steven-Eardley:
released, added the code style file too.

Issue #1376, opened 2017-09-07
Reapplications: write and run script


dommitchell:
Before we can publicise the end of the processing of the reapplications, we need to do some housekeeping. To help us, we would like you to create the following:

1) a list of journals still In DOAJ which have a rejected reapplication.
2) a list of journals In DOAJ with no reapplication record and no Tick
3) a list of journals In DOAJ that have the note: 'Continuation automatically extracted from journal {x} during migration'
[NOT NEEDED 4) a list of journals In DOAJ without The Tick]

Each report should be importable in Excel etc and include: Title, ISSN and internal DOAJ ID.

I expect that there are some stragglers in DOAJ that have the same title but different ISSNs (due to being continuations) so the comparisons for (1) and (2) should be run against ISSN and title.

I expect the 4th one of these will consist largely of journals already covered by other lists but we want to catch all eventualities.

We would like this to be given highest priority so that we can announce the end of the Reapps process.

Steven-Eardley:
The approach here will be to scroll through all of the journals with ```in_doaj = true``` once, with the 4 report generators gathering information from each journal. They will create 1 CSV file each, with the columns as above except that ISSN will be split into two columns for PISSN and EISSN.

> 1. a list of journals still In DOAJ which have a rejected reapplication.

Given the journal, check ```admin.current_application``` to get the reapplication record ID, verify that it's a reapplication (```index.application_type```) then add to report if it's in ```rejected``` status.

> 2. a list of journals In DOAJ with no reapplication record
Given the journal, check ```admin.current_application``` to get the reapplication record ID. If there isn't one, add it to the report.

> 3. a list of journals In DOAJ that have the note: 'Continuation automatically extracted from journal {x} during migration'
Given the journal, perform a text (regex) search on the notes field. Add to report if present.

> 4. a list of journals In DOAJ without The Tick
Given the journal, add to report if ```journal.is_ticked()``` is False.

The rows in the CSV files will be initially ordered by DOAJ ID, but your spreadsheet program can sort by title or by ISSN to deal with continuations.

dommitchell:
> Given the journal, check admin.current_application to get the reapplication record ID, verify that it's a reapplication (index.application_type) then add to report if it's in rejected status.
> 
> a list of journals In DOAJ with no reapplication record

Steve, sorry. I need to modify this one:

- a list of journals In DOAJ with no reapplication record and no Tick.

Steven-Eardley:
[...]

PR #1394, opened 2017-09-22
Hotfix/2017 09 20 1376 reapp housekeeping

Steven-Eardley:
Mostly a copy of ```reports.py```, but with scheduling removed and the reports changed. Run with:

    python portality/scripts/reapplication_reports.py -o <path_to_empty_dir> -e

and it will email the reports as well as store them in the directory specified. Multiple runs on the same day and target will overwrite the files. Example files as emailed are here (generated from my dev env):
[reapplication_reports_2017-09-22.zip](https://github.com/DOAJ/doaj/files/1325111/reapplication_reports_2017-09-22.zip)

For issue #1376 

## HOTFIX

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ src*
 .Python
 build/*
 dist/*
+tmp/*
 
 # PyCharm and other IntelliJ-based IDE-s and editors' folder:
 .idea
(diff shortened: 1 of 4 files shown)


ClaraDOAJ:
Can I download the reports from your .zip?

Steven-Eardley:
@ClaraDOAJ No, that’s been generated as an example using my incomplete and out of date copy of DOAJ data.

When we run them for real they’ll be emailed to feedback@doaj.org.

ClaraDOAJ:
Ah! thanks!

Issue #1406, opened 2017-10-05
Editors should be able to revert to "in progress"


Steven-Eardley:
See https://github.com/DOAJ/doaj/issues/1293#issuecomment-330530959 - Currently, changes in status 'backwards' in the editorial pipeline are disallowed for associates and editors. Changing to "in progress" from "completed" should be possible for editors.

Steven-Eardley:
Released now, editors should be able to change back to in progress now.

dommitchell:
@ClaraDOAJ this has been added to the training manual for asseds

dommitchell:
@Steven-Eardley @ClaraDOAJ don't we need a notification for this event too? Associate Editors get a notification when an application has been assigned to them. Can we re-use that for this instance too?

ClaraDOAJ:
Thanks @dommitchell! 

I think this might have broken something. Now the feedback account gets an email every time an editor sends an application back to an associate editor.

![image](https://user-images.githubusercontent.com/22593616/32770102-a4e4e9f2-c91e-11e7-8a55-cbc3de913a71.png) Why could this be?

Steven-Eardley:
>  don't we need a notification for this event too? Associate Editors get a notification when an application has been assigned to them. Can we re-use that for this instance too?

By 'we' do you mean a notification for managing editors? If so, is the email below enough?

> I think this might have broken something. Now the feedback account gets an email every time an editor sends an application back to an associate editor.

The feedback address is CC'd by default into all emails that the system sends.

ClaraDOAJ:
@Steven-Eardley but I only got these notifications lately. Don't worry, I will create a filter to skip inbox.

dommitchell:
So it looks as if notifications are already set up? Let's see what Steve says.

dommitchell:
Notifications as being sent out.

PR #1407, opened 2017-10-05
Allow an exception to the formcontext editorial pipeline restrictions…

Steven-Eardley:
… for editors changing 'completed' back to 'in progress'. #1406 
## HOTFIX

### DIFF
--- a/portality/formcontext/formcontext.py
+++ b/portality/formcontext/formcontext.py
@@ -4,7 +4,7 @@ from flask.ext.login import current_user
 import json, uuid
 from datetime import datetime
 
-from portality.formcontext import forms, xwalk, render, choices, emails
+from portality.formcontext import forms, xwalk, render, choices, emails, excepts
 from portality.lcc import lcc_jstree
 from portality import models, app_email, util
 from portality.core import app
@@ -12,10 +12,6 @@ from portality.core import app
 ACC_MSG = 'Please note you <span class="red">cannot edit</span> this application as it has been accepted into the DOAJ.'
 SCOPE_MSG = 'Please note you <span class="red">cannot edit</span> this application as you don\'t have the necessary ' \
             'account permissions to edit applications which are {0}.'
-
-
-class FormContextException(Exception):
-    pass
 
 
 class FormContext(object):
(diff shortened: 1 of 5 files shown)


portality/formcontext/formcontext.py
```diff
@@ -797,8 +797,8 @@ def finalise(self):
             raise FormContextException(
                 "You don't have permission to edit applications which are in status {0}.".format(src_status))
 
-        # Don't permit changes to status in reverse of the editorial process
-        if editor_choices.index(target_status) < editor_choices.index(src_status):
+        # Don't permit changes to status in reverse of the editorial process for statuses other than 'in progress'
+        if target_status != 'in progress' and editor_choices.index(target_status) < editor_choices.index(src_status):
```
richard-jones: I have to confess I'm not a huge fan of this "forward only, except in these conditions".  It would be nicer to have a function which would give you your allowed statuses given your current status, which could be done with the Choices object, perhaps.  Something like:

editors_choices = choices.Choices.application_status("editor", mycurrentstatus)

I realise that this is just one example, and this file is probably full of situations where we could better encapsulate the workflow.

portality/formcontext/formcontext.py
```diff
@@ -871,8 +871,16 @@ def _set_choices(self):
                 # Get the full tuple of application status for the current source
                 [full_current_status] = filter(lambda x: self.source.application_status in x, editor_choices)
 
-                # Only show options forward in the editorial pipeline
-                self.form.application_status.choices = editor_choices[editor_choices.index(full_current_status):]
+                # Show options forward in the editorial pipeline
+                forward_choices = editor_choices[editor_choices.index(full_current_status):]
+
+                # But allow an exception: reverts to 'in progress' from 'completed'
+                if self.source.application_status == 'completed':
```
richard-jones: Yeah, same comment as above - let's encapsulate the choices the user had in the Choices module as much as possible.

Steven-Eardley:
@richard-jones Ready for a new review. The functions have been moved to ```choices.py```, but aren't really much cleaner.

Issue #1404, opened 2017-10-03
Journal application submitted without answers to compulsory questions


dommitchell:
Can you tell me how this was submitted?
https://doaj.org/admin/suggestion/5e2d3a5ba1544f7c8ff9f35dc4bedb17

It was filled in online, manually and the 'bug' happened twice. (https://doaj.org/admin/suggestion/239ff6a2d95549358d5f0b1af0094da0 - but we asked the submitter to fill it in again.)

Steven-Eardley:
Interesting. I'm having trouble replicating this. When I try to submit the form after copying all of the fields, it fails validation as expected:

<img width="594" alt="screen shot 2017-10-04 at 10 13 06" src="https://user-images.githubusercontent.com/1446355/31168380-042a34a0-a8ed-11e7-94c9-11bc93d005ce.png">

But I can confirm these fields are missing from the saved application record.

dommitchell:
Another one here: https://doaj.org/admin/suggestion/a29fc2ff235c48298b3a5ba7ddf859f9

greboun:
this really should not be possible  content  licensing info is missing
except for deposit policy question
Are there still more of these?
The application was not triaged yet, so Bella would go and reject it



Dr. Tom Olijhoek
Editor-in-Chief  DOAJ
Support us: www.doaj.org/support

dommitchell:
The last time this happened, [I think] it was tied to an Internal Server Error where something had failed exactly as someone was making a submission. CL will be able to confirm. Bella will alert us if there are more.

richard-jones:
@richard-jones @Steven-Eardley and @emanuil-tolev to review possibilities for reproducing this

@dommitchell to review the extent of the problem.

richard-jones:
We have had a look at this in our tech meeting today, and now know the answer.  There is a bug in the form crosswalk which emerges under the following situation:

* A user selects "Other" on question 47
* A user provides the text "None" next to the "Other" selection on question 47.

This will pass form validation, and then the crosswalk will erroneously interpret the text "None" as a python-language NoneType, and ignore that entire section of the form, resulting in the behaviour described in this issue.

The issue could also be reproduced with the following values in Other: "none", "False", "false".

I will look into a fix for this today, and provide a PR later.

In the mean time, the extent of the problem is as follows:

* 12 items are affected
* Affected items range in date from 16th April to 3rd October

The ids of affected items are:

5e2d3a5ba1544f7c8ff9f35dc4bedb17
[...]

PR #1412, opened 2017-10-10
add test for broken form crosswalk, and fix in crosswalk

richard-jones:
HOTFIX

see #1404


richard-jones:
Incidentally, the tests all passed locally

Steven-Eardley:
Managed to restart the tests enough to get it to pass; deploying.

Issue #1413, opened 2017-10-10
Two identical journal records created upon acceptance


dommitchell:
http://bit.ly/2kC6mZU

They are identical and both were created at exactly the same time. Any idea why?

dommitchell:
And another: http://bit.ly/2xCFEBS

richard-jones:
That's very weird!  Did you just discover these in the dataset, or did you see them being created.  Can you tell us what you were doing when you created them, and anything "interesting" about these records that sets them apart from normal?

dommitchell:
Have connected you to Rikard via email. He spotted these.

ClaraDOAJ:
hi! Apparently the man editor computer has been registering double clicks when he single clicks and this caused duplicate records in some instances...

Can we just delete one of the records?

ClaraDOAJ:
And another one

http://bit.ly/2z9iDsD

Please can you help?

richard-jones:
Hi @ClaraDOAJ - it's fine just to delete one of the records.  I believe @emanuil-tolev is working on a fix to prevent this from happening when the button is erroneously clicked twice.

ClaraDOAJ:
Thanks! I have withdrawn the journals with this problem from DOAJ and added a note so publishers can continue to work with them. I haven´t deleted them completely. @dommitchell I think we should, but I wanted your opinion.

dommitchell:
Yes, we should just delete them as they will just clutter up the db. Thanks!

ClaraDOAJ:
@dommitchell have you just done it?

dommitchell:
Nope!

ClaraDOAJ:
Weird, most of them are not showing in the search as duplicate (with a yes-no) anymore. I have deleted two extra that Rikard had on a separate list too.

richard-jones:
@ClaraDOAJ - do you have a list of the ones that are duplicated, or do we need to try and make one?  We'd basically need to run a report over the database that identified duplicated ISSNs, and then you would review it to see if there is any action that needs taking.  Let me know if you want us to look at that.

ClaraDOAJ:
Hi @richard-jones, in principle it was a silly problem of a failing mouse that kept clicking twice. The man editor made a list of 5 and I have taken action in all of them. If I encounter any more instances I will ask you to run the report, unless it is very easy.

richard-jones:
[...]

PR #1423, opened 2017-11-03
introduce a short piece of javascript to disable the submit button on…

richard-jones:
… journal and application form pages when it is pressed.  This prevents second accidental clicks, which may result in duplicating behaviour.

I have applied this to all the forms, not just the managing editor form, where the problem noted in #1413 first materialised.

Since a submit of the form is always followed by a page reload, there's no need to have a mechanism to re-enable the button - the page reload does this.

### DIFF
--- a/portality/static/doaj/js/suggestions_and_journals.js
+++ b/portality/static/doaj/js/suggestions_and_journals.js
@@ -154,7 +154,11 @@ jQuery(document).ready(function($) {
     $("#editor_group").change(function(event) {
         event.preventDefault();
         $("#editor").html("<option val='' selected='selected'></option>")
-    })
+    });
+
+    $(".application_journal_form").bind("submit", function(event) {
+        $(".save-record").attr("disabled", "disabled");
+    });
 });
 
 function setup_subject_tree() {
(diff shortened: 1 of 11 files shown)


portality/static/doaj/js/suggestions_and_journals.js
```diff
@@ -154,7 +154,11 @@ jQuery(document).ready(function($) {
     $("#editor_group").change(function(event) {
         event.preventDefault();
         $("#editor").html("<option val='' selected='selected'></option>")
-    })
+    });
+
+    $(".application_journal_form").bind("submit", function(event) {
```
Steven-Eardley: ```.bind()``` should be ```.on()```. https://www.w3schools.com/jquery/event_bind.asp

Steven-Eardley: When I declare this event bind in the JS console on the application page it works as intended, but I see no effect when running the DOAJ. Perhaps something is superseding it in the imports?

Steven-Eardley:
Oh, I may have misinterpreted this. Since there's a reload does that mean I won't be seeing the disabled button?

richard-jones:
You should only see the disabled button very very briefly when you hid the SAVE button.  Once the page has been submitted, and reloaded, the button will be available again.  The only intention of this PR is to prevent you from hitting the save button twice in very quick succession.  If you try that, you'll find your second click falls on a disabled button.

richard-jones:
also replaced bind with on :)

PR #1427, opened 2017-11-05
add 1px margin to freetext search box for alignment in chrome

richard-jones:
This does not appear to break firefox.  I have not done extensive cross browser testing, though.


Steven-Eardley:
I've had a quick look, and am not seeing the need for it:

Firefox, Chrome, Safari, local (with your changes) on left, remote on right.

screen shot 2017-11-06 at 10 34 13

richard-jones:
That's so weird - I did wonder whether it was just me.  Maybe a Chrome version?  Here it is without the fix on my machine (local and live look the same).

screenshot from 2017-11-06 12 36 06

richard-jones:
I think probably don't merge this, though.  Lets see if anyone else has the problem, and we can revisit if it is more widespread.

Steven-Eardley:
@dommitchell @emanuil-tolev @ClaraDOAJ Do you see wonky search bars like Richard does?

dommitchell:
Nope. Not seeing this.

Issue #1390, opened 2017-09-21
2405-6561 XML with software exception error- possible BUG.


ClaraDOAJ:
Dear CL,

The attached xml ended up with this error. The publisher is experienced and I do not think it is an erroneus xml. Could you please have a look?
[PETLM_3_3.xml.zip](https://github.com/DOAJ/doaj/files/1320311/PETLM_3_3.xml.zip)

![image](https://user-images.githubusercontent.com/22593616/30684003-d42027f2-9eaf-11e7-86d8-aea1e19ef719.png)

Many thanks!

richard-jones:
I can't see a background job that failed around this time, so I guess this happened when the file was uploaded rather than when it was processed.  A quick skim of the file, and it looks sensible, though I have not validated it.

The journal for which this was intended is not on the test server, so I can't immediately try to replicate the issue.

@Steven-Eardley @emanuil-tolev - could one of you either or both: refresh the test server with up-to-date data so I can try this out, and check the logs for this timestamp to see if there is any more detail there.

For reference, this is the journal whose ISSN appears in the first article: https://doaj.org/toc/2405-5816

ClaraDOAJ:
Also, they have tried to upload the file several times, different days. So it was not when our system was sluggish. 

@richard-jones the journal is "Petroleum"- correct.

richard-jones:
I've just updated my dev environment for parity with live, and run through the process of uploading and ingesting this file, and it worked fine.  This would suggest that the file, upload process and background task are all behaving correctly in principle.  Could this be something in the deployment?

richard-jones:
That's on a reduced set of articles - there could be something going on, say, if a duplicate is found.  I may replicate the entire DOAJ to my machine and re-try.

richard-jones:
Though note that re-uploading the file again and re-processing it straight away works fine, and correctly detects all the duplicates.

richard-jones:
I tried several ways to reproduce this in a dev environment yesterday, and didn't encounter any problems.

@emanuil-tolev @Steven-Eardley - I think this must be a deployment environment thing, but I'm not sure what, can one of you take a look?

Steven-Eardley:
First I'll check the logs for the trace.

richard-jones:
[...]

PR #1429, opened 2017-11-09
Hotfix/2017 11 07 1390 lcc regen subjects

Steven-Eardley:
The migration for #1390 - fixing spelling errors in Applications, Journals and Articles' LCC subject categories in the bibjson. Expected output:

```
Running in dev
Loaded final config from /Users/steve/code/cl/doaj/src/doaj/dev.cfg
Starting 2017-11-09 17:20:54.987763.
Upgrading suggestion
writing  500 to suggestion
writing  500 to suggestion
0c1e4e44533e405db9a4a4fdddb7c864 Changed.
old: [{'code': u'H', 'term': u'WRONG SPELLING', 'scheme': u'LCC'}, {'term': u'Carry over', 'scheme': u'DOAJ'}]
new: [{'code': u'H', 'term': u'Social Sciences', 'scheme': u'LCC'}, {'term': u'Carry over', 'scheme': u'DOAJ'}]
...
```

To test this (on a disposable index), the following edit to your index might be useful. You may need to change the ID to a different application:

```
curl -X POST \
  http://localhost:9200/doaj/suggestion/0c1e4e44533e405db9a4a4fdddb7c864/_update \
  -H 'cache-control: no-cache' \
  -d '{
    "doc": {
    	"bibjson": {
        	"subject": [
                {
                    "code": "H",
                    "term": "WRONG SPELLING",
                    "scheme": "LCC"
                },
                {
                	"term": "Carry over",
                    "scheme": "DOAJ"
                }
            ]
    	},
    	"index": {
    		"classification_paths": [
    			"WRONG SPELLING"
    		],
    		"classification": [
    			"WRONG SPELLING"
    		]
    	}
    }
}'
```

The same cURL body can be used for a Journal and Article, too. The ```DOAJ``` scheme code should be carried over to the new record along with a corrected spelling of the ```LCC``` category, in this case, **Social Sciences**. The changes to the index are overwritten by the ```prep()``` method when the model is saved in the upgrade script. 

The script is run as stated in the readme thusly:

    python portality/upgrade.py -u portality/migrate/1390_lcc_regen_subjects/regen_subjects.json

The script is expected to run for a long time, since it runs on all articles.

### DIFF
--- a/portality/migrate/1390_lcc_regen_subjects/operations.py
+++ b/portality/migrate/1390_lcc_regen_subjects/operations.py
@@ -1,0 +1,15 @@
+from portality import models
+
+
+def refresh_subjects(record):
+    """ Using the GenericBibjson we can fix the subjects on journals, applications and articles with this function """
+    bj = record.bibjson()
+    assert isinstance(bj, models.GenericBibJSON)
+    old_subjects = bj.subjects()
+    bj.remove_subjects()
+
+    new_subjects = []
+    for s in old_subjects:
+        # Do stuff
+        new_subjects.append(s)
+    return record
--- a/portality/migrate/1390_lcc_regen_subjects/operations.py
+++ b/portality/migrate/1390_lcc_regen_subjects/operations.py
@@ -14,10 +14,13 @@ def refresh_subjects(record):
             sobj = {"scheme": u'LCC', "term": lcc.lookup_code(s['code']), "code": s['code']}
             new_subjects.append(sobj)
         except KeyError:
-            # Subject has no code, that's fine, let's have a look at it
-            print "Missing code:", s
+            # Carry over the DOAJ schema subjects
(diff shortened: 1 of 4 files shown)


Steven-Eardley:
@richard-jones right now I think it won't, and that it's a bug. See https://github.com/DOAJ/doaj/blob/develop/portality/models/article.py#L253 in ```add_journal_metadata```:

```python
        for s in jbib.subjects():
            if s not in bibjson.subjects():
                bibjson.add_subject(s.get("scheme"), s.get("term"), code=s.get("code"))
                trip = True
            rbj.add_subject(s.get("scheme"), s.get("term"), code=s.get("code"))
```

We only add subjects to an article, rather than replacing what's broken in the article's bibjson. This also means that if a journal's subject is removed we won't remove it from the article so it's not a proper sync.

Issue #1439, opened 2017-11-17
Applications, provenance and journals in inconsistent state


richard-jones:
In some cases there are applications whose status and last updated date do not match the provenance records surrounding the changes made to them, or the created_dates of the journals made from them.

dommitchell:
CL's work here is done. Reporting will run again in Jan to check for inconsistencies. 

DOAJ to update the records highlighted in RJ's spreadsheet.

Can be assigned to DM?

dommitchell:
New comment sent today indicating that there may still be a problem

"I made some editing and accepted an application yesterday afternoon,
and now I find it back as ready without my comment and edits.

"The application is the following:
https://doaj.org/admin/suggestion/876d65d2c9d64108bd436f140c988d7a

"I edited field 10-11 and added a comment (at least that's what I recall)"

dommitchell:
From @richard-jones 
I've checked this one in particular, and it all seems fine.  There's an update recorded from this morning, and it's in status "accepted" - it's difficult to tell whether this was a real problem or not, as the admin says they have subsequently fixed it, and it's not showing up in our analysis.  I don't see a provenance record suggesting that it was edited earlier on, so perhaps this was a problem submitting the data initially, rather than a recurrence of the inconsistencies issue.

We have re-run the analysis today, anyway, and mostly everything looks fine, though there is one thing to check:

This Journal: 294f885a02c6435092f8c4b0380a766b says that the owner is account 13139703, but there's no record of that user account in the system.  That journal was accepted on 2017-12-19.  I've added it to the Data Restore tab of the spreadsheet for review.

Dom: Thanks for checking that one for us.

Strange re the missing account (I created one manually). It makes me
think that the account problem may not be entirely related to the data
inconsistencies problem. Perhaps this account problem has been going
on for longer and is related to something else? Indeed, there were
much more missing accounts that any other type of error.

Thoughts?

dommitchell:
Worth noting here that Notes have been lost and are non-retrievable.

dommitchell:
A possible, new example: 498f1e35571647f6a5db8e78b186edcc - changes made to the application were unsaved. Can you check? @richard-jones

richard-jones:
Hi Dom,

I've checked the provenance on this one, and there are no provenance
records of acceptance before the most recent one saying the item was
accepted.  I can see an "edit" event on the day that the item had a note
added to accept it, and then no activity until 8th January.  I can also see
a single journal created on 23rd January.

[...]

PR #1442, opened 2017-11-17
Improved error logging in DAO save method.

Steven-Eardley:
#1439

### DIFF
--- a/portality/dao.py
+++ b/portality/dao.py
@@ -134,7 +134,7 @@ class DomainObject(UserDict.IterableUserDict, object):
                 try:
                     error_details = r.json()
                 except (ValueError, AttributeError):
-                    error_details = None
+                    error_details = r.text
 
                 # Retries depend on which end the error lies.
                 if 400 <= r.status_code < 500:
--- a/portality/dao.py
+++ b/portality/dao.py
@@ -3,8 +3,6 @@ from copy import deepcopy
 from datetime import datetime, timedelta
 import time
 import re
-# debugging
-import traceback, sys
 
 from portality.core import app
 import urllib2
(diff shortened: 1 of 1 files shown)


portality/dao.py
```diff
@@ -120,19 +122,29 @@ def save(self, retries=0, back_off_factor=1, differentiate=False, blocking=False
...
-                if 400 <= r.status_code < 500:
-                    # bad request, no retry
-                    print "bad request", r.json()
-                    traceback.print_stack(file=sys.stdout)
-                    break
-                elif r.status_code >= 500:
-                    print "server error", r.json()
-                    attempt += 1
+                if r.status_code > 400:
+                    raise ElasticSearchWriteException(u"Error on ES save. Response code {0}".format(r.status_code))
```
emanuil-tolev: worth getting the body as well? Not r.json(), but (I assume) r.body or r.text, or whatever the raw response is

emanuil-tolev: Oh, my bad, this is handled below

portality/dao.py
```diff
@@ -120,19 +122,29 @@ def save(self, retries=0, back_off_factor=1, differentiate=False, blocking=False
...
                     break  # everything is OK, so r should now be assigned to the result
 
             except requests.exceptions.ConnectionError:
+                app.logger.exception(u"Failed to connect to ES")
                 attempt += 1
+            except ElasticSearchWriteException:
+                try:
+                    error_details = r.json()
+                except (ValueError, AttributeError):
+                    error_details = None
```
emanuil-tolev: In this case maybe use `r.text`? http://docs.python-requests.org/en/master/user/quickstart/#response-content

PR #1456, opened 2017-11-30
Feature/1296 cautious deletes

Steven-Eardley:
#1296 - Only auto-delete duplicate articles when they are both DOI and full text duplicates.
Make the duplicate detection function a little more discerning when it comes to matching DOIs: they should actually look like DOIs.

### DIFF
--- a/portality/article.py
+++ b/portality/article.py
@@ -156,10 +156,12 @@ class XWalk(object):
         if len(dois) > 0:
             # there should only be the one
             doi = dois[0]
-            articles = models.Article.duplicates(issns=issns, doi=doi)
-            possible_articles['doi'] = [a for a in articles if a.id != article.id]
-            if possible_articles['doi']:
-                found = True
+            # Only perform de-duplication on genuine DOIs - todo: do we need a full regex check?
+            if isinstance(doi, str) and doi.startswith('10.'):
+                articles = models.Article.duplicates(issns=issns, doi=doi)
+                possible_articles['doi'] = [a for a in articles if a.id != article.id]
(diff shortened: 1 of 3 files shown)


portality/article.py
```diff
@@ -156,10 +156,12 @@ def discover_duplicates(article, owner=None):
         if len(dois) > 0:
             # there should only be the one
             doi = dois[0]
-            articles = models.Article.duplicates(issns=issns, doi=doi)
-            possible_articles['doi'] = [a for a in articles if a.id != article.id]
-            if possible_articles['doi']:
-                found = True
+            # Only perform de-duplication on genuine DOIs - todo: do we need a full regex check?
```
richard-jones: Probably not - DOIs are relatively free-form after the prefix.  I think you have to have one or more digits following the ".", but after that it's up to you what you do with it.  So, starting with 10. is sufficient, I'd say.

portality/article.py
```diff
@@ -156,10 +156,12 @@ def discover_duplicates(article, owner=None):
...
+            if isinstance(doi, str) and doi.startswith('10.'):
+                articles = models.Article.duplicates(issns=issns, doi=doi)
+                possible_articles['doi'] = [a for a in articles if a.id != article.id]
+                if possible_articles['doi']:
+                    found = True
```
richard-jones: I'm generally against testing for truthiness.  I'd do if len(possible_articles['doi']) > 0: found  = True

Steven-Eardley:
I have fixed it, happy to have PR changes requested for code style changes! I'd also missed a few other things from the tests.

PR #1460, opened 2017-12-01
add script to flip application status, so we can change the status of…

richard-jones:
… items that are accepted to something that can be edited

When we apply this we just need a list of ids in a text file, and the status we want to convert them to, then we can run

python change_application_status.py -c [ids] -s [new status]


richard-jones:
I pondered it too.  I guess it's probably better to do that than not.  Leave it to me, and I'll push something shortly.

richard-jones:
ok, done

PR #1487, opened 2018-01-11
Used supplied svg logo for PLOS, and converted CL to svg. 

Steven-Eardley:
Leading by example!


Steven-Eardley:
I used our shared folder version of the CL logo, which included text. If you'd prefer for me to vectorise the sans-text logo, request a change on this PR.

This is how the new svgs look (PLOS & CL):

screen shot 2018-01-11 at 17 48 56

It's like getting new glasses!

Steven-Eardley:
I've tested on all the browsers available to me (latest chrome, safari, firefox, edge, ie11), but I think our browserstack has expired. I think any would be with old versions of internet explorer.

Steven-Eardley:
Per our previous discussion, here's what it looks like on IE 8 without fallback:

screen shot 2018-01-29 at 14 23 12

As @richard-jones said, users who are on end-of-life browsers are perhaps less interested in the colourful logos on the website, so we may not want to have a fallback at all.

Also worth noting, facetview doesn't work when I tested it on IE 8.

dommitchell:
I am happy to just show the link text for these use cases.

Issue #1462, opened 2017-12-08
Fix and extend Google Analytics reporting for non-UI endpoints


richard-jones:
The GA recording has recently ceased working, and we need to diagnose and fix.

The Universal Analytics python module is a couple of years old, perhaps the API has changed, and we need an update.  This project has not changed on github for a couple of years, so it's possible it no longer provides the support we need.

See also some advice by Google on how to record analytics:

https://cloud.google.com/appengine/docs/flexible/python/integrating-with-analytics

We would basically be integrating with their Measurement Protocol, which is documented here:

https://developers.google.com/analytics/devguides/collection/protocol/v1/

I'd like to see us put together an analytics module which also wraps up some of the shared secrets from the analytics code.  For example, instead of specify the string "API Hit", this should be a reference to a constant.

We should have both a decorator and a direct function for recording events.

Currently the analytics is only applied to the API, and we should extend this to cover all the other non-UI endpoints:

* OAI-PMH
* Atom
* OpenURL
* Journal CSV download
* any others?

richard-jones:
Remember that once we've done this, this could open up better ways of tracking widget usage.

Steven-Eardley:
@dommitchell Can you clarify, has Google Analytics completely stopped working in production? I.e. the dashboard shows no users or sessions whatsoever.

I've tried adding GA to the test server without changing the code, and it works for the main pages.

dommitchell:
Nope, that's the wrong view. Use this one:
https://analytics.google.com/analytics/web/?authuser=1#embed/report-home/a46560124w77600930p80218834/

Steven-Eardley:
API data is missing from behaviour / events (the historical data)

Steven-Eardley:
Okay @dommitchell, I reckon I've fixed events so that they appear in the behaviour reports (on the test server currently). Next I'd like to know a little more about what you'd like to gather from the non-UI endpoints.

### OAI-PMH
At the moment we have the following actions:
* Identify
* ListMetadataFormats
* GetRecord
* ListSets    
* ListRecords
* ListIdentifiers
And two types, article and journal. I'd propose an event like this:
```JavaScript
{
    event_category: "OAI-PMH"
    event_action: "article|journal:{identify|listmetadataformats|getrecord|...}"
    event_label: "<record ID>"
}
```
I.e. For each of the 6 actions we preface their type for 12 total different actions:
```article:identify, article:getrecord, journal:identify, journal:getrecord etc...```

[...]

PR #1492, opened 2018-01-22
Feature/1462 improved google analytics

Steven-Eardley:
#1462 

Analytics have been factored into lib, and added to atom, journalcsv, openurl, and oai-pmh.


portality/view/api_v1.py
```diff
@@ -55,7 +55,7 @@ def docs():
 @swag(swag_summary='Search your applications <span class="red">[Authenticated, not public]</span>', swag_spec=DiscoveryApi.get_application_swag())  # must be applied after @api_key_(optional|required) decorators. They don't preserve func attributes.
 @blueprint.route("/search/applications/<path:search_query>")
 @api_key_required
-@track_analytics('API Hit', 'Search applications', record_value_of_which_arg='search_query')
+@analytics.sends_ga_event('API Hit', 'Search applications', record_value_of_which_arg='search_query')
```
richard-jones: I'm quite keen to get rid of "magic strings" in our code, so it would be good to place "API Hit" and "Search applications" into a constants file that we just reference.  I've done this in the Update Requests code, too, if you want to see an example of one approach.  For the moment, just having these events listed somewhere and referenced would be great.  That way at least we guarantee consistent spelling and capitalisation, with IDE-level checking that we're using the right values, and runtime guarantee that the values are set.

portality/view/oaipmh.py
```diff
@@ -4,32 +4,44 @@
...
-## Web API endpoints
+# Web API endpoints
 #####################################################################
 
+
 @blueprint.route("/oai", methods=["GET", "POST"])
 @blueprint.route("/oai.<specified>", methods=["GET", "POST"])
 def oaipmh(specified=None):
+    # Google Analytics event
+    event = {'category': 'OAI-PMH', 'action': None, 'label': None, 'fieldsobject': None}
```
richard-jones: I think there's too much knowledge about the internal shape of an Event here.  It would be better if the analytics code presented an Event class which was populated through an interface, then implementers don't need to remember what all the fields are called, or risk making a spelling error, or something like that.

Steven-Eardley:
@richard-jones @emanuil-tolev this is ready for another review - logging added and RJ's changed made.

Steven-Eardley:
I've now merged the widget URLs code here, so it can go to test with the analytics.

Related issues:
https://github.com/DOAJ/doajPM/issues/1564
https://github.com/DOAJ/doajPM/issues/974

portality/app.py
```diff
@@ -113,6 +113,14 @@
 SPONSORS = {k: OrderedDict(sorted(v.items(), key=lambda t: t[0])) for k, v in SPONSORS.items()}
 
 
+# Configure the Google Analytics tracker
+from portality.lib import analytics
+try:
+    analytics.create_tracker(app.config['GOOGLE_ANALYTICS_ID'], app.config['BASE_DOMAIN'])
+except KeyError:
+    app.logger.warn("No Google Analytics credentials found. Required: 'GOOGLE_ANALYTICS_ID' and 'BASE_DOMAIN'.")
```
emanuil-tolev: This feels too important in production for a mere warning. Is this because there's no sensible values to set these to in development?

portality/settings.py
```diff
@@ -584,14 +584,53 @@
...
+GA_DIMENSIONS = {
+    'oai_res_id': 'dimension1',                                                                    # In GA as OAI:Record
+}
 
-# =========================================
-# scheduled reports configuration
-REPORTS_BASE_DIR = "/home/cloo/reports/"
-REPORTS_EMAIL_TO = ["feedback@doaj.org"]
+# GA for OAI-PMH
+OAI_GA_CATEGORY = 'OAI-PMH'
```
emanuil-tolev: I'd have personally named these GA_CATEGORY_* rather than *_GA_CATEGORY. I'm totally ambivalent about this though, so feel free to leave as is, just providing a different point of view.

portality/app.py
Steven-Eardley: Well GA is really entirely optional, so I didn't think it was an error to omit the details. Do you suggest I make it an error with a check for env=production? That'd be even less bothersome for developers I suppose.

portality/settings.py
Steven-Eardley: Evidently not completely ambivalent :)

I've often gone back and forth on this - do I want the differentiated text to be first (better for autofill) or last, which I guess is better for categorisation and understanding.

if you have ```dog_collie``` and ```dog_boxer``` is that more understandable than ```collie_dog``` or ```boxer_dog```? Easier to search for, since you can pick your choice of dog when you've half typed it? hmmmmmm

Steven-Eardley:
I'm happy to make the suggested changes then merge to develop. If @richard-jones puts up a PR for his work then we can have both on the test server for a while.

portality/settings.py
emanuil-tolev: Ambivalent in the sense that I mentioned it to have another point of view but you've obviously thought about it :)

emanuil-tolev:
Isn't Richard's branch ... well, published? If it's on test? You could make a PR from your branch to his branch maybe, then I could review and merge?

Steven-Eardley:
It’s my understanding that Richard’s feature branch is deployed to the test server, rather than develop. I don’t remember reviewing the code at any rate.

Steven-Eardley:
To clarify, I meant for the develop branch to return to the test server, with both feature branches merged. It'll be more current that merging mine to his, but with the same effect.

PR #1504, opened 2018-02-17
link DOIs against preferred resolver

katrinleinweber:
Hello!

The DOI foundation recommends a [new, secure resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8).

Cheers :-)

### DIFF
--- a/portality/app.py
+++ b/portality/app.py
@@ -209,7 +209,7 @@ def doi_url(doi):
     :return: the HTML link
     """
     tendot = doi[doi.find('10.'):]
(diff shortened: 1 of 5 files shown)


Steven-Eardley:
Really useful, thanks. I've also searched for other instances of the outdated ```dx.doi``` and found another regex [here](https://github.com/DOAJ/doaj/blob/develop/portality/view/forms.py#L24). I believe you've already provided a solution in https://github.com/richard-jones/magnificent-octopus/pull/12!

PR #1505, opened 2018-02-19
Broaden reg-ex to preferred DOI variants

katrinleinweber:
Same as https://github.com/richard-jones/magnificent-octopus/pull/12 ;-)


Steven-Eardley:
I'm afraid this is superseded by #1506, which I based on your work and further simplified, catching a couple more errors in our original.

katrinleinweber:
There is always a better reg-ex :-)

PR #1506, opened 2018-02-19
Simplified and updated the DOI regex

Steven-Eardley:
based on contributions by @katrinleinweber 

I've made this a string literal to improve readability (doing away with the ```\\```), and removed duplicated protocol portions. Finally, escaped an extra ```.``` which allowed us to accept e.g. ```dx.doiZorg``` since it's a wildcard.

### DIFF
--- a/doajtest/unit/test_regexes.py
+++ b/doajtest/unit/test_regexes.py
@@ -1,8 +1,8 @@
 """ Gather and test DOAJ regexes here """
 
 from doajtest.helpers import DoajTestCase
-from portality.view.forms import DOI_REGEX
-from doajtest.fixtures import dois
+from portality.view.forms import DOI_REGEX, ISSN_REGEX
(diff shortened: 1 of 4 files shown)


doajtest/unit/test_regexes.py
```diff
@@ -0,0 +1,34 @@
...
+from portality.view.forms import DOI_REGEX
+from doajtest.fixtures import dois
+
+import re
+
+
+class TestRegexes(DoajTestCase):
+
+    def setUp(self):
+        super(TestRegexes, self).setUp()
```
emanuil-tolev: These can just be `pass` I think - do you need the index set up and torn down? (Adds 2s per new test.)

Steven-Eardley:
@emanuil-tolev this is ready for another review. See https://github.com/DOAJ/doajPM/issues/1604#issuecomment-371479365 for details about the handle identifiers.

PR #1508, opened 2018-03-02
Hotfix/issue 1296 further expansions to dedupe

Steven-Eardley:
More duplication criteria, from running the report this week. We now consider a fuzzy match on 2 other features of the article as well as DOI and fulltext URL.

https://github.com/DOAJ/doajPM/issues/1296

### DIFF
--- a/portality/scripts/article_duplicates_report_remove.py
+++ b/portality/scripts/article_duplicates_report_remove.py
@@ -9,6 +9,7 @@ from portality.models import Article, Journal, Account
 from portality.article import XWalk
 from portality.core import app
 from datetime import datetime
+from itertools import combinations
 import time
 import esprit
 
@@ -46,8 +47,10 @@ def duplicates_per_article(connection, delete, snapshot, owner=None, query_overr
             if fuzzy_dups:
                 print "\t{0} fuzzy duplicates: {1}".format(len(fuzzy_dups), ", ".join(fuzzy_dups))
 
-            # Only consider duplicates that appear in at least 2 lists (we should be cautious with deletes) TODO: 3 - way comparison
-            set_of_duplicates = set(doi_dups).intersection(set(fulltext_dups))
+            # Only consider duplicates that appear in at least 2 lists (we should be cautious with deletes)
+            set_of_duplicates = reduce(set.union,
+                                       map(lambda x: set(x[0]).intersection(set(x[1])),
+                                           combinations([doi_dups, fulltext_dups, fuzzy_dups], 2)))
             dupcount += len(set_of_duplicates) + 1                     # The detected duplicates plus the article itself
 
             if set_of_duplicates:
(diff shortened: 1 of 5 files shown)


portality/scripts/article_duplicates_report_remove.py
```diff
@@ -36,15 +37,20 @@ def duplicates_per_article(connection, delete, snapshot, owner=None, query_overr
...
                 print "\t{0} DOI duplicates: {1}".format(len(doi_dups), ", ".join(doi_dups))
             if fulltext_dups:
                 print "\t{0} fulltext duplicates: {1}".format(len(fulltext_dups), ", ".join(fulltext_dups))
+            if fuzzy_dups:
+                print "\t{0} fuzzy duplicates: {1}".format(len(fuzzy_dups), ", ".join(fuzzy_dups))
 
-            # Only consider duplicates that appear in both lists (we should be cautious with deletes)
-            set_of_duplicates = set(doi_dups).intersection(set(fulltext_dups))
+            # Only consider duplicates that appear in at least 2 lists (we should be cautious with deletes)
+            set_of_duplicates = reduce(set.union,
```
emanuil-tolev: Sorry, I have a little trouble following this through. Could you do me a favour and write out here what it's doing as a loop (instead of the `map`) and summary operation (instead of the `reduce`)?

Steven-Eardley: I'll just go a step further and explain it:

From inner to outer, we have:
```python
combinations([doi_dups, fulltext_dups, fuzzy_dups], 2)
```
Which gives us pairs of duplicate lists to compare. This ensures we only accept duplicates that appear in 2 of the lists. These are passed into:

```
map(lambda x: set(x[0]).intersection(set(x[1])), ...)
```
Which takes the paired duplicate lists, converts to set and calculates the intersection. Therefore we have a list of sets of the items that appear in 2 duplicate lists.

Finally, we want to get the full list of items that appear in 2 of the duplicate lists, so we combine them all with ```reduce(set.union, ...)``` using the union operation.

emanuil-tolev: So both the DOI and fulltext URL have to be the same for it to be considered a duplicate. Or fuzzy criteria + DOI or fulltext.

Could we put something to that effect in the function's description comment? (if I'm understanding it right)

The reason I asked for it to be written out more imperatively was partly to understand it more easily, but also partly to implicitly compare the readability of both approaches ;).

emanuil-tolev:
I'm sure you've seen but there seems to be a related rather than random test failure: https://app.codeship.com/projects/49473/builds/32888802?pipeline=dc78edd1-08c7-4128-b95d-f482fce3b963

```
File "/home/rof/src/github.com/DOAJ/doaj/doajtest/unit/test_article_match.py", line 397, in test_11_article_replace_weak_criteria
assert dups['doi'] is None
TypeError: 'NoneType' object has no attribute '__getitem__'
```

Steven-Eardley:
Yup, those tests prove the code is broken! Closing this PR.

PR #1509, opened 2018-03-06
Release for GA improvements for widgets.

Steven-Eardley:


Steven-Eardley:
Holding on to this as I've found a bug with searching via the index page, causing an exception.

PR #1510, opened 2018-03-06
simple search widget referrer had broken the homepage search box. 

Steven-Eardley:
Added referrer for homepage to fix.

### DIFF
--- a/portality/view/doaj.py
+++ b/portality/view/doaj.py
@@ -74,8 +74,9 @@ def search():
 
 @blueprint.route("/search", methods=['POST'])
 def search_post():
+    """ Redirect a query from the box on the index page to the search page. """
     if request.form.get('origin') != 'ui':
-        abort(501)  # not implemented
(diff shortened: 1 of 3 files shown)


portality/view/doaj.py
```diff
@@ -74,8 +74,9 @@ def search():
 
 @blueprint.route("/search", methods=['POST'])
 def search_post():
+    """ Redirect a query from the box on the index page to the search page. """
     if request.form.get('origin') != 'ui':
-        abort(501)  # not implemented
+        abort(501)                                          # not implemented - we must receive searches from our own UI
```
emanuil-tolev: Sorry to be a pain, I don't know why I didn't see this before. I'd probably use a 400 here? (bad request) The problem is the client data being sent over is not "right".

emanuil-tolev:
hah what happened?

Steven-Eardley:
fixed, pushed, feature finished! Might've been a bit over-keen...

PR #1513, opened 2018-03-20
Hotfix/1625/last update on toc

richard-jones:
Don't merge just yet, waiting for confirmation that this is the right two additions


richard-jones:
ready to merge

PR #1514, opened 2018-04-02
[hotfix] add and fix marvel pruning script

emanuil-tolev:
https://github.com/DOAJ/doajpm/issues/1525

### DIFF
--- a/doajtest/unit/test_prune_marvel.py
+++ b/doajtest/unit/test_prune_marvel.py
@@ -6,7 +6,7 @@ from doajtest.helpers import DoajTestCase
 from portality.scripts.prune_marvel import generate_delete_pattern
 
 
-class TestFeed(DoajTestCase):
+class TestPruneMarvel(DoajTestCase):
 
     @classmethod
     def setUpClass(cls):
--- a/doajtest/unit/test_prune_marvel.py
+++ b/doajtest/unit/test_prune_marvel.py
@@ -1,0 +1,38 @@
+import datetime, calendar
+from dateutil.relativedelta import relativedelta
+from freezegun import freeze_time
+
+from doajtest.helpers import DoajTestCase
+from portality.scripts.prune_marvel import generate_delete_pattern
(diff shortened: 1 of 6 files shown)


doajtest/unit/test_prune_marvel.py
```diff
@@ -0,0 +1,38 @@
+import datetime, calendar
+from dateutil.relativedelta import relativedelta
+from freezegun import freeze_time
+
+from doajtest.helpers import DoajTestCase
+from portality.scripts.prune_marvel import generate_delete_pattern
+
+
+class TestFeed(DoajTestCase):
```
Steven-Eardley: Is this classname as intended?

portality/scripts/prune_marvel.py
```diff
@@ -0,0 +1,30 @@
...
+    delete_date = now - relativedelta(months=3)
+
+    return ".marvel-{0}*".format(delete_date.strftime('%Y.%m'))
+
+
+def main():
+    delete_pattern = generate_delete_pattern()
+    print "To delete: {0}".format(delete_pattern)
+
+    if len(delete_pattern) < 8:
```
Steven-Eardley: Does this mean that the delete pattern is shorter than the string ```.marvel-*```? That's 9 characters. Or have I missed the point of this?

emanuil-tolev: That is the idea, yes. I think this kind of check should be done in the tests, now that we have tests to verify the behaviour of the code. In fact we know what the length of the output must be: 16. I've documented that with a new unit test assertion.

emanuil-tolev: (and removed this runtime exception)

doajtest/unit/test_prune_marvel.py
emanuil-tolev: Yes, I have decided to name all my tests that!
(fixed ta)

PR #1518, opened 2018-04-12
Added logrotate config for the analytics log, set up the symlinks for…

Steven-Eardley:
… existing logrotate configs.

https://github.com/DOAJ/doajPM/issues/1623


Steven-Eardley:
@emanuil-tolev changes made!

PR #1522, opened 2018-04-20
Added script to generate CSV of publisher email addresses, as request…

Steven-Eardley:
…ed in issue https://github.com/DOAJ/doajPM/issues/1645

### DIFF
--- a/portality/scripts/publisher_email_in_doaj.py
+++ b/portality/scripts/publisher_email_in_doaj.py
@@ -1,0 +1,94 @@
+""" Creates a CSV of publisher email addresses """
+
+import os, codecs
+import esprit
+from portality.core import app
+from portality.models import Account, Journal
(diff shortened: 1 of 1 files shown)


portality/scripts/publisher_email_in_doaj.py
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+
+def publishers_with_journals():
+    """ Get accounts for all publishers with journals in the DOAJ """
+    for acc in esprit.tasks.scroll(conn, 'account', q=publisher_query):
+        account = Account(**acc)
+        journal_ids = account.journal
+        if journal_ids is not None:
```
emanuil-tolev: Surprised to see the underlying method return a list of values when there are journals and `None` rather than `[]` when there are none to return. Don't have to fix it in this PR.

PR #1525, opened 2018-04-30
Attempts to fix the deployment issue - upgrade to flask 1.0.1

Steven-Eardley:
This required some changes to imports, and a few methods have become properties.
A workaround for the email log tests has been removed.

I really can't see why setup.py isn't being read per the issue below, so I've just removed the bodge to see what happens.

https://github.com/DOAJ/doajPM/issues/1527


deploy/deploy-gateway.sh
```diff
@@ -34,19 +34,7 @@ git submodule update --recursive
...
-pip install flask-cors==2.1.2 # for some reason this is not being picked up from setup.py?!
-# none of these newly introduced requirements are being picked up by pip
-# something's wrong with the virtualenv
-pip install LinkHeader==0.4.3
-pip install universal-analytics-python==0.2.4
-pip install huey==1.7.0
-pip install redis==2.10.5
-pip install rstr==2.2.5
-pip install freezegun==0.3.10
+pip install --force-reinstall -r requirements.txt
```
emanuil-tolev: Is the --force-reinstall necessary if the requirements are non-conflicting?

setup.py
```diff
@@ -3,15 +3,15 @@
...
-        "werkzeug==0.8.3",
-        "Flask==0.9",
-        "Flask-Login==0.1.3",
-        "Flask-WTF==0.8.3",
-        "Flask-Mail==0.9.0",
-        "requests==1.1.0",
+        "werkzeug==0.14.1",
+        "Flask==1.0.1",
+        "Flask-Login==0.4.1",
+        "Flask-WTF==0.14.2",
```
emanuil-tolev: Oh dear, this line means testing the forms somehow

emanuil-tolev:
> I really can't see why setup.py isn't being read per the issue below, so I've just removed the bodge to see what happens.

Yeah, me neither. I think we should wipe the replication repository and stop recreating the virtualenv every time. Probably just need to dig to see why it's not working. Maybe the live background jobs server?

deploy/deploy-gateway.sh
Steven-Eardley: possibly not, and the install should be fresh in a new venv in any case. I just found this is the flag that installs any case, including downgrades.

setup.py
Steven-Eardley: it does. Since I've also changed how the form components and validators are imported, it's critical we test.

Steven-Eardley:
that's a good idea, using background-apps. I was wondering how we'd test this without staging infrastructure.

PR #1530, opened 2018-05-18
change how different applications links are presented depending on th…

richard-jones:
…eir status and type

Don't merge this just yet, just waiting for final confirmation of the terminology from #1704


emanuil-tolev:
Need to review, terminology has been confirmed in 1704

Steven-Eardley:
Checked and released.

PR #1532, opened 2018-05-21
Feature/2018 05 15 remove email from author

Steven-Eardley:
TBD: what happens during the time that the updated code runs and the migration is ongoing. We may want read only mode.

https://github.com/DOAJ/doajPM/issues/1667

### DIFF
--- a/portality/static/doaj/doajArticles.xsd
+++ b/portality/static/doaj/doajArticles.xsd
@@ -97,8 +97,6 @@
        <xs:complexType>
         <xs:sequence>
          <xs:element name="name" type="xs:string" />
-         <xs:element name="email" type="xs:string"
-                     minOccurs="0" />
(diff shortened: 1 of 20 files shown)


Steven-Eardley:
The tests are passing locally, but not on Codeship after a few attempts.

Steven-Eardley:
Ahah! there was a successful attempt!

portality/static/doaj/doajArticles.xsd
```diff
@@ -97,8 +97,6 @@
        <xs:complexType>
         <xs:sequence>
          <xs:element name="name" type="xs:string" />
-         <xs:element name="email" type="xs:string"
```
emanuil-tolev: This may need to be communicated in advance of the release. Potentially a temporally restricted note on the XML upload page too?

PR #1535, opened 2018-06-14
Feature/1656 prune live backups

Steven-Eardley:
The es-backups bucket has been duplicated, we must remember to delete that again once this has been running in production for a while.

We could also stage this on the test server, with a smaller TTL to watch the background task do its job.


deploy/nginx/doaj-production-gate
```diff
@@ -55,7 +55,8 @@ server {
     listen          80;
     server_name     testdoaj.cottagelabs.com;
 
-    # temporarily allow plain HTTP acces to testdoaj
+    auth_basic "Restricted";
```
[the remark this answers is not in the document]
Steven-Eardley: You did! However, we have moved to staging feature branches on the test server, so I needed that commit so I didn't revert the security work you did.

Steven-Eardley:
Tests for this were successful: the background tasks ran and deleted the correct backups.

**check_latest_es_backup** worked - it correctly ran as scheduled and found that the backup was present.

**prune_es_backups** also worked, it deleted all backups until the configured TTL, which was 2 days. The log shows which backups were deleted. It also ran as scheduled but I'd already triggered it using the script.

screen shot 2018-06-27 at 13 31 50

screen shot 2018-06-27 at 13 33 04

This is already configured for live - the TTL and snapshots repository are in ```settings.py```.

Steven-Eardley:
I've also brought the copied bucket of all DOAJ backups up to date, so it's safer to make this live.

portality/lib/es_snapshots.py
```diff
@@ -0,0 +1,84 @@
...
+class SnapshotDeleteException(Exception):
+    pass
+
+
+class ESSnapshot(object):
+    def __init__(self, snapshot_json):
+        self.data = snapshot_json
+        self.name = snapshot_json['snapshot']
+        self.state = snapshot_json['state']
+        self.datetime = datetime.utcfromtimestamp(snapshot_json['start_time_in_millis'] / 1000)
```
richard-jones: For all these where we read a property from the input, would it be safer to use `.get` rather than accessing the property directly.  Then its value can be defaulted, or if it is missing it won't necessarily break the class.  Or do we want it to break the class during construction?

portality/tasks/check_latest_es_backup.py
```diff
@@ -0,0 +1,72 @@
...
+        Submit the specified BackgroundJob to the background queue
+
+        :param background_job: the BackgroundJob instance
+        :return:
+        """
+        background_job.save()
+        check_latest_es_backup.schedule(args=(background_job.id,), delay=10)
+
+
+@main_queue.periodic_task(schedule("check_latest_es_backup"))
```
richard-jones: I see you have this on the main_queue and the other task on the long_running queue.  This seems sensible, and I just wanted to double check that this is correct, and that this checking task should be on the main_queue and not the long_running queue.  (I don't know how long it will take to run, it might be super-quick)

portality/lib/es_snapshots.py
```diff
@@ -0,0 +1,84 @@
...
+    def __str__(self):
+        return str(self.__dict__)
+
+    def __repr__(self):
+        return self.__str__()
+
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__
+
+    def delete(self):
```
richard-jones: Delete is a client operation, rather than a model object operation.  If you're going to separate the client and the models it deals with (which I think is right), then you should put the operational stuff into the client.

Steven-Eardley:
> My only general question would be if we've given any thought to whether any of this should sit inside the BLL or not? I'm not sure it should, it feels orthogonal, but we should at least make sure we've thought about it.

We actually were considering this a possible enhancement for esprit - since snapshots are a built-in function of ES, it would fit nicely there.

portality/lib/es_snapshots.py
Steven-Eardley: It intentionally fails if these aren't present, and there's no sensible default for representing a genuine snapshot record. If these are missing, something isn't right. That's caught on line 60.

portality/tasks/check_latest_es_backup.py
Steven-Eardley: I did intentionally move the deletion to the long running queue, but I think I may want the check to go there too. Right now, it takes 2-5 minutes to complete because there's so much metadata to return. When the prune is ongoing, it'll be quicker. I'll move it though.

portality/lib/es_snapshots.py
Steven-Eardley: Thanks, I'll change that

portality/tasks/check_latest_es_backup.py
Steven-Eardley: Actually, we definitely want to run the check every day, and the main queue is the right place for that. I'll leave it in the main queue.

Steven-Eardley:
@richard-jones I've moved the delete function, how do you feel about moving this functionality to esprit?

richard-jones:
@Steven-Eardley - I'm actually quite in favour of putting this in esprit, yes.  If there are no DOAJ-specific mechanics going on, then it makes a lot of sense.

Steven-Eardley:
I've moved the feature to ESSnapshotsClient to esprit, this PR now consists of the background tasks utilising the lib within the DOAJ.

emanuil-tolev:
I vaguely remember there were some tests with the snapshot client. Did they make it to esprit?

emanuil-tolev:
This PR LGTM

PR #1537, opened 2018-06-26
Feature/1535 anonymise test server data

emanuil-tolev:
**Do not merge before https://github.com/CottageLabs/esprit/pull/13 is merged first, but feel free to review.**

https://github.com/DOAJ/doajpm/issues/1535

Welcome feedback on the tests - they seem a bit silly, but are simply setting expectations on what the feature code should do. Also I forgot to make a few of the anon functions return `record` which broke the anon dump process for types that used those funcs, and the tests do catch that.

Another important thing is whether this is comprehensively anonymous, and I think it is as I reviewed all the data models while writing it, but do have a think while reviewing if I've missed something.


emanuil-tolev:
I've realised the username anonymisation is far too inconvenient and that information is not usually identifiable info. When it is, it's just a name. In terms of convenience even the DOAJ team will have trouble logging in.. so I think that needs to be taken out.

Steven-Eardley:
> I've realised the username anonymisation is far too inconvenient and that information is not usually identifiable info. When it is, it's just a name. In terms of convenience even the DOAJ team will have trouble logging in.. so I think that needs to be taken out.

I'm inclined to agree, when I keep in mind how many other users are on the test server.

portality/lib/anon.py
```diff
@@ -0,0 +1,14 @@
...
+def anon_email(email):
+    return hashlib.sha256(app.config['ANON_SALT'] + email).hexdigest() + '@example.com'
+
+
+def basic_hash(id_):
+    return hashlib.sha256(app.config['ANON_SALT'] + id_).hexdigest()
+
+
+def anon_name(name):
+    return 'Stacy ' + basic_hash(name)
```
richard-jones: Is there a reason for using the name "Stacy" here.  I get that this might make the names more readable up-front, but having them all start with the same name might be confusing.  I'd either just do away with this, or we could randomly generate names from a set of first-name last-name pairs, which is a trick I've used before to make realistic looking data.

portality/scripts/anon_export.py
```diff
@@ -0,0 +1,129 @@
+import esprit
+from portality.core import app
+from portality.lib.anon import basic_hash, anon_name, anon_email
+
+
+def anonymise_email(record):
+    if 'email' in record:
+        record['email'] = anon_email(record['email'])
```
richard-jones: I think overall this file is awkwardly structured and implemented, and comes with some maintenance overheads in the long run.  Because it uses the raw data rather than the model objects, it relies on an innate understanding of the underlying data model, which will break as soon as we change it.  Instead, each anonymising function should take a model object, and should perform the anonymisation by re-writing the properties via the object's API.  This will make this module easier to read and less likely to break or have bugs in it.

So, for example, the above should be something like

```
def anonymise_account_email(account):
    account.email = anon_email(account.email)
```

I realise this makes this function less re-usable for other objects that also have an `email` property, but given that all this does is check there is an email property and then call another function that anonymises the email address, that's not a loss.

In fact, I wouldn't even have this function at all, I'd just put the operational content into the `anonymise_account` function.

emanuil-tolev: Excellent point! I came at it from the perspective of what esprit gives me so didn't think of this. Esprit supplies the raw record to the transform function (which is what all these are). At the moment, it doesn't pass on the type of the record it's dumping to the transform f(), so the tranform wouldn't know which model class to initialise with the incoming esprit record data. I could modify esprit so it passes the type on to the transform f(). Then that will be able to init a model and do what you suggest.

emanuil-tolev:
Significantly simpler code now due to removing all the username stuff.

portality/scripts/anon_export.py
richard-jones: I'm not sure why it's hard to get models, we know what the mapping from a type in ES to a model is.  I think that I'd put this information in your `transform` function, then it's none of Esprit's business what the data actually represents.  So, to anonymise an account:

```python
def anonymise_account(raw):
    account = Account(**raw)
    account.email = anon_emmail(account.email)
    return account.data
```

emanuil-tolev: I see what you mean now, the transforms taking charge without esprit passing `type` around works very well. All done now.

portality/lib/anon.py
emanuil-tolev: Just a way to distinguish the data is a name at all. Replaced with human-like names quickly using Faker, a neat library we could use in lots of places.

portality/scripts/anon_export.py
```diff
@@ -0,0 +1,91 @@
...
+    parser.add_argument("-l", "--limit", type=int, help="Number of records to export from each type. If you specify e.g. 100, then only the first 100 accounts, 100 journals, 100 articles etc. will be exported. The \"first\" 100 will be ordered by whatever the esprit iterate functionality uses as default ordering, usually alphabetically by record id.")
+    args = parser.parse_args()
+    if args.limit > 0:
+        limit = args.limit
+    else:
+        limit = None
+
+    conn = esprit.raw.make_connection(None, app.config["ELASTIC_SEARCH_HOST"], None, app.config["ELASTIC_SEARCH_DB"])
+
+    for type_ in ['account']:# esprit.raw.list_types(connection=conn):
```
Steven-Eardley: is the comment supposed to be reinstated?

emanuil-tolev: Yes, not sure how that snuck in. Fixing + fixing tests.

emanuil-tolev:
@richard-jones @Steven-Eardley ready for review again. Conflicts fixed, tests improved, script improved a little. Running them individually actually gave errors but running all at once worked - to do with Faker library, now everything is fixed and we have a working example of it.

emanuil-tolev:
(please don't delete the feature_1535... branch for now)

PR #1539, opened 2018-06-27
Hotfix/1722 hide not in doaj from publisher

richard-jones:
This hides journals from publishers in the publisher area if the journal is not in doaj, as per #1722


richard-jones:
ok, I will merge and deploy

PR #1540, opened 2018-06-27
Feature/1706 reporting on missing passwords

richard-jones:
This adds a script in case we need it in future to report on missing passwords.

It does not need to be run, as it has already been run remotely for the results.

### DIFF
--- a/portality/dao.py
+++ b/portality/dao.py
@@ -92,6 +92,10 @@ class DomainObject(UserDict.IterableUserDict, object):
     def last_updated(self):
         return self.data.get("last_updated")
 
+    @property
+    def last_updated_timestamp(self):
+        return datetime.strptime(self.last_updated, "%Y-%m-%dT%H:%M:%SZ")
+
     def save(self, retries=0, back_off_factor=1, differentiate=False, blocking=False):
 
         if app.config.get("READ_ONLY_MODE", False) and app.config.get("SCRIPTS_READ_ONLY_MODE", False):
--- a/portality/scripts/accounts_with_missing_passwords.py
+++ b/portality/scripts/accounts_with_missing_passwords.py
@@ -1,0 +1,66 @@
+"""
+This script can be run to generate a CSV output of accounts which do not have their passwords set, along
+with some useful account information, and possible explanations for the lack of password
+
+```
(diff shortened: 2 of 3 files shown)


portality/scripts/accounts_with_missing_passwords.py
```diff
@@ -0,0 +1,66 @@
...
+        writer = UnicodeWriter(f)
+        writer.writerow(["ID", "Name", "Email", "Created", "Last Updated", "Updated Since Create?", "Has Reset Token", "Reset Token Expired?"])
+
+        for a in esprit.tasks.scroll(conn, models.Account.__type__, q=MISSING_PASSWORD, page_size=100, keepalive='1m'):
+            account = models.Account(_source=a)
+
+            has_reset = account.reset_token is not None
+
+            expires = account.reset_expires_timestamp
+            is_expired = expires < datetime.utcnow()
```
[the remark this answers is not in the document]
richard-jones: Yes, very good point.  I've pushed a change to move this.

portality/dao.py
```diff
@@ -92,6 +92,10 @@ def created_timestamp(self):
     def last_updated(self):
         return self.data.get("last_updated")
 
+    @property
+    def last_updated_timestamp(self):
+        return datetime.strptime(self.last_updated, "%Y-%m-%dT%H:%M:%SZ")
```
Steven-Eardley: do we use the ```portality.lib.dates``` module for specifying our preferred date stamp format, or perhaps reading it straight from ```app.config.get("DEFAULT_DATE_FORMAT")```?

I don't know if we've done this before, but we must have quite a few formatter strings dotted around the code that are supposed to be the same.

richard-jones: Yeah, we should do.  Probably a concerted effort to move over to `dates` is required.  Let me ponder it, and maybe I'll make the change here, and perhaps to the rest of this file for this PR.

richard-jones: Ok, I've had a think about this, and I'm not going to change it for the PR, as it's an orthogonal question.  Instead I'm going to open an issue to track the idea that we need to move to more standardisation in how we handle dates.

PR #1548, opened 2018-08-05
Hyperlink DOIs to secure resolver also from widget

katrinleinweber:
Same as a3a8d2d96655890f1e6aec6abe329d369edcd80b & #1504 ;-)


Steven-Eardley:
Thanks @katrinleinweber! Looks like we had an inadvertent regression caused by @richard-jones's commit https://github.com/DOAJ/doaj/commit/8ad4cd3ad79a0d72138efee2bfa1a65e0c3213f1#diff-d5bfc22a9b80bfb06c2d5a5a7b703a79 when we re-generated that widget source file.

Steven-Eardley:
Changes merged with #1555

PR #1551, opened 2018-08-10
FOR REVIEW: add simple per-session rate limiter for query endpoint

richard-jones:
Do not merge unless deemed necessary.


Steven-Eardley:
I think it might be worth having a discussion about this in a tech meeting sometime - there are a few aspects of the session that I don't think I am familiar with. Apart from that, the code looks fine, if a little short on documentation.

PR #1558, opened 2018-08-22
Changed urls in API docs to use HTTPS to avoid unnecessary redirects

Steven-Eardley:
https://github.com/DOAJ/doajPM/issues/1771

### DIFF
--- a/portality/app.py
+++ b/portality/app.py
@@ -307,9 +307,9 @@ if 'api' in app.config['FEATURES']:
                 'api_versions': [
                     {
                         'version': '1.0.0',
-                        'base_url': url_for('api_v1.api_spec', _external=True),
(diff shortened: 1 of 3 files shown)


Steven-Eardley:
Note you may wish to override the preferred scheme in your dev.cfg to http.

Also, I was expecting the config change to be sufficient without specifying the scheme in ```url_for``` per https://stackoverflow.com/questions/14810795/flask-url-for-generating-http-url-instead-of-https and http://flask.pocoo.org/docs/1.0/config/#builtin-configuration-values but it didn't have an effect. We're on Flask 1.0.1

richard-jones:
This looks good.  I have heard talk about using SSL locally in development being a good idea too.  Should we consider that?

Steven-Eardley:
We should! https://github.com/DOAJ/doajPM/issues/1772

PR #1570, opened 2018-10-07
Remove duplicate eissn line

bcody:
I think this was just a duplication so wanted to remove it to avoid any potential confusion.


richard-jones:
@bcody - great, thanks for this

PR #1573, opened 2018-10-19
feature/1749 OAI remove deep paging

anusharanganathan:
PR for issue https://github.com/DOAJ/doajPM/issues/1749

Modified query and resumption token to avoid deep paging

### DIFF
--- a/portality/models/oaipmh.py
+++ b/portality/models/oaipmh.py
@@ -51,7 +51,7 @@ class OAIPMHRecord(object):
 
     set_limit = {"term" : { "index.schema_subject.exact" : "<set name>" }}
     range_limit = { "range" : { "last_updated" : {"gte" : "<from date>", "lte" : "<until date>"} } }
-    created_sort = {"last_updated" : {"order" : "desc"}}
+    created_sort = [{"last_updated" : {"order" : "desc"}}, {"id" : "desc"}]
 
     def earliest_datestamp(self):
         result = self.query(q=self.earliest)
@@ -66,19 +66,21 @@ class OAIPMHRecord(object):
         sets = [t.get("term") for t in result.get("facets", {}).get("sets", {}).get("terms", [])]
         return sets
 
-    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None):
+    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None, start_after=None):
         q = deepcopy(self.records)
-        if from_date is not None or until_date is not None or oai_set is not None:
+        if start_after is not None or from_date is not None or until_date is not None or oai_set is not None:
 
             if oai_set is not None:
                 s = deepcopy(self.set_limit)
                 s["term"]["index.schema_subject.exact"] = oai_set
                 q["query"]["bool"]["must"].append(s)
 
-            if until_date is not None or from_date is not None:
+            if until_date is not None or from_date is not None or start_after is not None:
                 d = deepcopy(self.range_limit)
 
-                if until_date is not None:
+                if start_after is not None:
+                    d["range"]["last_updated"]["lte"] = start_after[0]
+                elif until_date is not None:
                     d["range"]["last_updated"]["lte"] = until_date
                 else:
                     del d["range"]["last_updated"]["lte"]
@@ -93,13 +95,16 @@ class OAIPMHRecord(object):
         if list_size is not None:
             q["size"] = list_size
 
-        if start_number is not None:
-            q["from"] = start_number
+        if start_after is not None:
+            q["from"] = start_after[1]
+        else:
+            q["from"] = 0
 
-        q["sort"] = [deepcopy(self.created_sort)]
+        q["sort"] = deepcopy(self.created_sort)
 
         # do the query
         # print json.dumps(q)
+
         results = self.query(q=q)
 
         total = results.get("hits", {}).get("total", 0)
@@ -107,9 +112,9 @@ class OAIPMHRecord(object):
 
 
 class OAIPMHArticle(OAIPMHRecord, Article):
-    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None):
+    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None, start_after=None):
         total, results = super(OAIPMHArticle, self).list_records(from_date=from_date,
-            until_date=until_date, oai_set=oai_set, list_size=list_size, start_number=start_number)
+            until_date=until_date, oai_set=oai_set, list_size=list_size, start_number=start_number, start_after=start_after)
         return total, [Article(**r) for r in results]
 
     def pull(self, identifier):
@@ -120,9 +125,9 @@ class OAIPMHArticle(OAIPMHRecord, Article):
         return None
 
 class OAIPMHJournal(OAIPMHRecord, Journal):
-    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None):
+    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None, start_after=None):
         total, results = super(OAIPMHJournal, self).list_records(from_date=from_date,
-            until_date=until_date, oai_set=oai_set, list_size=list_size, start_number=start_number)
+            until_date=until_date, oai_set=oai_set, list_size=list_size, start_number=start_number, start_after=start_after)
         return total, [Journal(**r) for r in results]
 
     def pull(self, identifier):
--- a/portality/view/oaipmh.py
+++ b/portality/view/oaipmh.py
@@ -145,12 +145,14 @@ def decode_set_spec(setspec):
     return decoded
 
 
-def get_start_after(docs):
+def get_start_after(docs, current_start_after, list_size):
     last_date = docs[-1].get("last_updated")
     count = 0
     for doc in docs:
         if doc.get("last_updated") == last_date:
             count += 1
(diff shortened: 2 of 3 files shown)


portality/models/oaipmh.py
```diff
@@ -66,19 +66,21 @@ def list_sets(self):
         sets = [t.get("term") for t in result.get("facets", {}).get("sets", {}).get("terms", [])]
         return sets
 
-    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None):
+    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None, start_after=None):
```
richard-jones: It looks like we can remove start_number from the method signature now

portality/models/oaipmh.py
```diff
@@ -93,23 +95,26 @@ def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=
...
         total = results.get("hits", {}).get("total", 0)
         return total, [hit.get("_source") for hit in results.get("hits", {}).get("hits", [])]
 
 
 class OAIPMHArticle(OAIPMHRecord, Article):
-    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None):
+    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None, start_after=None):
         total, results = super(OAIPMHArticle, self).list_records(from_date=from_date,
-            until_date=until_date, oai_set=oai_set, list_size=list_size, start_number=start_number)
+            until_date=until_date, oai_set=oai_set, list_size=list_size, start_number=start_number, start_after=start_after)
```
richard-jones: Same comment as above: start_number can be removed from the method signature, and doesn't need to be passed forward here.

portality/models/oaipmh.py
```diff
@@ -120,14 +125,14 @@ def pull(self, identifier):
         return None
 
 class OAIPMHJournal(OAIPMHRecord, Journal):
-    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None):
+    def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=None, start_number=None, start_after=None):
         total, results = super(OAIPMHJournal, self).list_records(from_date=from_date,
-            until_date=until_date, oai_set=oai_set, list_size=list_size, start_number=start_number)
+            until_date=until_date, oai_set=oai_set, list_size=list_size, start_number=start_number, start_after=start_after)
```
richard-jones: Another place to remove start_number

portality/view/oaipmh.py
```diff
@@ -145,7 +145,15 @@ def decode_set_spec(setspec):
...
 
 
-def make_resumption_token(metadata_prefix=None, from_date=None, until_date=None, oai_set=None, start_number=None):
+def get_start_after(docs):
+    last_date = docs[-1].get("last_updated")
+    count = 0
+    for doc in docs:
+        if doc.get("last_updated") == last_date:
+            count += 1
+    return (last_date, count)
```
richard-jones: I was thinking a bit more about this, and wondering if we might encounter any issues, and one possibility came to mind.

Sometimes we run migrations on the database which touch all of the items.  As a result, their last_updated date gets refreshed, and for a time the entire index will have last_updated dates all within a few seconds of each other. Could this be a problem?  It might cause the same deep paging effect, until the last_updated dates get more spread out by ongoing actions.

We don't do this very often, and given how long migrations can take to run it's unlikely that the entire index would have the same last_updated date, so the problem could never be as bad as it is now, but is there anything else clever we can do to order the items or filter the items to avoid this date-searching issue.  The newer ES version's "search_after" /would/ get us around this issue, so a possible +1 for a back-end upgrade there.

richard-jones: Ah, in fact, I'll go you one further - I can see a bug that this might cause.  If you get to two pages with all the same last_updated dates, doesn't this make it impossible to get past the second page?

So, for example, suppose there are 200 records with the same last_updated date, and we have page sizes of 10.  On the first page there will be 10 results, all with date A.  The start_after value will then be (A, 11).  So the next query will get records 11 - 20.  These will also have date A, so the start_after value will be (A, 11) again, as the date is the same and there are 10 results all with that date.

I have some thoughts on a fix, though not well worked through yet:

* We can get the correct value for `start_after[1]` if we can keep track of how many previous records (including on previous pages) have the same value for last_updated
* We can probably encode this information in the resumption token
* We'll need to check the final record on the current page, and if its last_updated date is the same as the one that we started with, then sum the count of records on the page with the sum of all previous records with that last_updated date, and send that value forwards.  If it is not the same, then reset that cross-page counter back to 0, and then start again from the current page's offset.

portality/view/oaipmh.py
```diff
@@ -157,6 +165,8 @@ def make_resumption_token(metadata_prefix=None, from_date=None, until_date=None,
         d["s"] = oai_set
     if start_number is not None:
```
richard-jones: start_number can be removed

richard-jones: I'll stop writing that now - throughout there are various places where start_number appears and can just be removed.

portality/view/oaipmh.py
```diff
@@ -409,51 +420,57 @@ def _parameterised_list_identifiers(dao, base_url, specified_oai_endpoint, metad
...
+
             # if there are no results, PMH requires us to throw an error
             if len(results) == 0:
                 return NoRecordsMatch(base_url)
-            
+
+            # get the full total
+            full_total = total
+            if start_after is not None:
+                full_total = total + start_number - start_after[1]
```
richard-jones: I didn't understand this bit of code, sorry.  start_number seems to have become redundant elsewhere, so I'm not sure how it is being used here, or how full_total is different to total.  Can you just add some comments to explain?

portality/view/oaipmh.py
richard-jones: On the other point, another thing we could do is increase the last_updated granularity to the millisecond scale, which would minimise the number of records which appeared to have the same date.  Right now it is just stored to the second.

portality/models/oaipmh.py
anusharanganathan: The reason we need to keep the start number is so that the resumption token has the correct cursor and complete_list_size.

portality/view/oaipmh.py
anusharanganathan: Upgrading ES to use search_after will not help in case of migrations, as it uses the sort functionality to figure out the next record. Increasing the granularity of time to milliseconds should help

Ah yes, If the new record_offset equals the number of records, I also need to add the record_offset from the existing search. Thanks for spotting that. I will make that change.

portality/view/oaipmh.py
anusharanganathan: In the xml for the resumption token, it has complete_list_size and cursor

Since we are doing new searches each time, the total will be a reduced total, not the same as the one received in the first search. So I recalculate the total, and return the complete list size.

We would need to hold onto the start number so we can return the current cursor for the resumption token xml

portality/models/oaipmh.py
anusharanganathan: same answer as above :)

anusharanganathan:
@richard-jones I was thinking of how we could solve the problem of deep paging in case of migrations. I now sort with date and id. 

Could the search for the next set of results have both until_date less than x and id less than y, where x and y come from the last record served. if we do this, from will always be 0, we don't have to count similar results and there will be no need to change the granularity of the date.

```
```

portality/view/oaipmh.py
anusharanganathan: @richard-jones comment added to code and code modified to fix bug you spotted

richard-jones:
I had a similar thought about the ordering, and you can certainly use a range search on a string term, which would behave the way you would want.

The problem is that this constraint would apply to the entire result set, so you would only get results which had:

* A last updated date earlier than the requested date
* An ID alphabetically less than the last ID

This would work /within/ the same last_updated date perfectly well, but when we transition to a new date in the result set, the id constraint would still be in effect, and you'd miss all the ids greater than the specified one under the new last_updated date.

I think the solution we have now is probably the right one, and the only thing I think we then need to do is increase the granularity of the dates we store.

If we want to go further, I'd consider replacing the entire OAI interface with a true change feed which is statically served.  This could be something we need to do at some point soon anyway, as then we could drive both OAI and a ResourceSync implementation off the same feed.

richard-jones:
When I say "soon", I mean in the next year or so, not in the timeline for this PR

portality/models/oaipmh.py
richard-jones: ok, I see the point in keeping it across the board.  In this method, though, that argument is never used so isn't it redundant?  And the same as in the other methods in models/oaipmh.py.

portality/view/oaipmh.py
richard-jones: This function looks good to me now

portality/view/oaipmh.py
richard-jones: Ah, got it!  Clever!

portality/models/oaipmh.py
anusharanganathan: Oh yes and Now fixed.

portality/models/oaipmh.py
anusharanganathan: Fixed here too

richard-jones:
I added a couple of lines to the oai client test which keeps track of ids, and checks that we get them all only once.  Everything looks good.

@Steven-Eardley or @markmacgillivray are you planning to do a review of this, or are we good to merge?

PR #1581, opened 2018-10-31
Feature/1751 remove deep paging from discovery

anusharanganathan:
This PR is a work in progress and not ready to be reviewed

### DIFF
--- a/portality/api/v1/discovery.py
+++ b/portality/api/v1/discovery.py
@@ -99,17 +99,17 @@ class DiscoveryApi(Api):
     @staticmethod
     def get_application_swag():
         return deepcopy(DISCOVERY_API_SWAG['application'])
-    
+
     @staticmethod
     def get_journal_swag():
         return deepcopy(DISCOVERY_API_SWAG['journal'])
-    
+
     @staticmethod
     def get_article_swag():
         return deepcopy(DISCOVERY_API_SWAG['article'])
 
     @classmethod
-    def _sanitise(cls, q, page, page_size, sort, search_subs, sort_subs):
+    def _sanitise(cls, q, page, page_size, sort, search_subs, sort_subs, bulk):
         if not allowed(q):
             raise DiscoveryException("Query contains disallowed Lucene features")
 
@@ -128,6 +128,14 @@ class DiscoveryApi(Api):
 
         # calculate the position of the from cursor in the document set
         fro = (page - 1) * page_size
+        # If fro is greater than the max allowed, throw error
+        # using bulk to provide an override when needed
+        max_records = app.config.get("DISCOVERY_MAX_RECORDS_SIZE", 1000)
+        if not bulk and fro >= max_records:
+            message = """Looks like you want to view results beyond {x}. \
+If you would like to see more results, you can download all of our results \
+from link""".format(x=max_records)
+            raise DiscoveryException(message)
 
         # interpret the sort field into the form required by the query
         sortby = None
@@ -154,9 +162,9 @@ class DiscoveryApi(Api):
         return q, page, fro, page_size, sortby, sortdir
 
     @classmethod
-    def _make_search_query(cls, q, page, page_size, sort, search_subs, sort_subs):
+    def _make_search_query(cls, q, page, page_size, sort, search_subs, sort_subs, bulk):
         # sanitise and prep the inputs
-        q, page, fro, page_size, sortby, sortdir = cls._sanitise(q, page, page_size, sort, search_subs, sort_subs)
+        q, page, fro, page_size, sortby, sortdir = cls._sanitise(q, page, page_size, sort, search_subs, sort_subs, bulk)
 
         # assemble the query
         query = SearchQuery(q, fro, page_size, sortby, sortdir)
@@ -190,7 +198,7 @@ class DiscoveryApi(Api):
 
         page_count = ((total - 1) // page_size) + 1
         last_page = FISRT_PAGE + page_count - 1
-        
+
         # Links to previous and next page
         if requested_page > FISRT_PAGE:
             previous_page = requested_page - 1
@@ -205,8 +213,7 @@ class DiscoveryApi(Api):
         return page_count, previous_page, next_page, last_page
 
     @classmethod
-    def _make_response(cls, endpoint, res, q, page, page_size, sort,
-                       obs):
+    def _make_response(cls, endpoint, res, q, page, page_size, sort, obs):
         total = res.get("hits", {}).get("total", 0)
 
         page_count, previous_page, next_page, last_page = cls._calc_pagination(total, page_size, page)
@@ -236,10 +243,10 @@ class DiscoveryApi(Api):
         return SearchResult(result)
 
     @classmethod
-    def search_articles(cls, q, page, page_size, sort=None):
+    def search_articles(cls, q, page, page_size, sort=None, bulk=False):
         search_subs = app.config.get("DISCOVERY_ARTICLE_SEARCH_SUBS", {})
         sort_subs = app.config.get("DISCOVERY_ARTICLE_SORT_SUBS", {})
-        query, page, page_size = cls._make_search_query(q, page, page_size, sort, search_subs, sort_subs)
+        query, page, page_size = cls._make_search_query(q, page, page_size, sort, search_subs, sort_subs, bulk)
 
         # execute the query against the articles
         res = models.Article.query(q=query.query(), consistent_order=False)
@@ -254,10 +261,10 @@ class DiscoveryApi(Api):
         return cls._make_response('search_articles', res, q, page, page_size, sort, obs)
 
     @classmethod
-    def search_journals(cls, q, page, page_size, sort=None):
+    def search_journals(cls, q, page, page_size, sort=None, bulk=False):
         search_subs = app.config.get("DISCOVERY_JOURNAL_SEARCH_SUBS", {})
         sort_subs = app.config.get("DISCOVERY_JOURNAL_SORT_SUBS", {})
-        query, page, page_size = cls._make_search_query(q, page, page_size, sort, search_subs, sort_subs)
+        query, page, page_size = cls._make_search_query(q, page, page_size, sort, search_subs, sort_subs, bulk)
 
         # execute the query against the articles
         res = models.Journal.query(q=query.query(), consistent_order=False)
--- a/portality/api/v1/discovery.py
+++ b/portality/api/v1/discovery.py
@@ -121,11 +121,14 @@ class DiscoveryApi(Api):
         if page < 1:
             page = 1
 
-        if page_size > app.config.get("DISCOVERY_MAX_PAGE_SIZE", 100):
+        if bulk :
+            max_pag_size = app.config.get("DISCOVERY_BULK_PAGE_SIZE", 1000)
+        else:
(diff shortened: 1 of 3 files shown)


anusharanganathan:
* Replace DiscoveryApi search with a scroll for the dump
* Use [query filters](https://github.com/DOAJ/doaj/blob/develop/portality/lib/query_filters.py)
    * Add another query filter to [include source](https://github.com/DOAJ/doaj/blob/develop/portality/api/v1/discovery.py#L362)
    * Replace public_result_filter with a source query filter instead in [settings.py](https://github.com/DOAJ/doaj/blob/develop/portality/settings.py#L342). 
        * The query filter needs to include sealed and ticked from the public_result_filter.
* Scroll gives one record at a time. Just json.dumps each record to file. Keep count and close the file after N number of lines and start a new file.

markmacgillivray:
What does scroll gives one record at a time mean? Can't a scroll return many records?

portality/api/v1/discovery.py
```diff
@@ -121,13 +121,24 @@ def _sanitise(cls, q, page, page_size, sort, search_subs, sort_subs):
...
-
         # calculate the position of the from cursor in the document set
         fro = (page - 1) * page_size
+        # If fro is greater than the max allowed, throw error
+        # using bulk to provide an override when needed
+        max_records = app.config.get("DISCOVERY_MAX_RECORDS_SIZE", 1000)
+        if not bulk and fro >= max_records:
+            message = """Looks like you want to view results beyond {x}. \
+If you would like to see more results, you can download all of our results \
+from link""".format(x=max_records)
```
Steven-Eardley: This is a placeholder for the full data dump URL, but this should be either be meaningful without that or we shouldn't release this until the dump is ready.

portality/api/v1/discovery.py
```diff
@@ -221,25 +231,39 @@ def _make_response(cls, endpoint, res, q, page, page_size, sort,
             "results" : obs
         }
 
+        # FIXME: URLs constructed without helper if exception is thrown
+        #       as application context is not available
+        s = sort
+        if s == None:
```
Steven-Eardley:
```suggestion
        if s is None:
```

portality/api/v1/discovery.py
```diff
@@ -221,25 +231,39 @@ def _make_response(cls, endpoint, res, q, page, page_size, sort,
             "results" : obs
         }
 
+        # FIXME: URLs constructed without helper if exception is thrown
```
Steven-Eardley: Can you explain in which circumstance we don't have an application context?

portality/api/v1/discovery.py
```diff
@@ -221,25 +231,39 @@ def _make_response(cls, endpoint, res, q, page, page_size, sort,
...
+        #       as application context is not available
+        s = sort
+        if s == None:
+            s = ''
         if previous_page is not None:
-            result["prev"] = app.config['BASE_URL'] + url_for(app.config['API_BLUEPRINT_NAME'] + '.' + endpoint, search_query=q, page=previous_page, pageSize=page_size, sort=sort)
+            try:
+                result["prev"] = app.config['BASE_URL'] + url_for(app.config['API_BLUEPRINT_NAME'] + '.' + endpoint, search_query=q, page=previous_page, pageSize=page_size, sort=sort)
+            except:
```
Steven-Eardley: too broad an exception - when and why does the line above fail?

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+def _save_file(typ):
+    # Create a dir for today and save all files in there
+    data = json.dumps(results, cls=ModelJsonEncoder)
+    filename = os.path.join(today, "{typ}_{page}.json".format(typ=typ, page=page))
+    output_file = tmpStore.path(container, filename, create_container=True, must_exist=False)
+    dir = os.path.dirname(output_file)
+    if not os.path.exists(dir):
+        os.makedirs(dir)
+    print("Saving file {filename}".format(filename=filename))
+    out = codecs.open(output_file, "wb", "utf-8")
```
Steven-Eardley: I prefer the ```with``` syntax here

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+    out.write(data)
+    out.close
```
Steven-Eardley: if not above, this should be `out.close()`

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+    return
```
Steven-Eardley: No need for `return` statement; it's implied at the end of a function

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def _get_last_page():
+    url = results.data.get('last', '')
+    last_page = None
+    if url:
```
Steven-Eardley: no case for `not url` - is that correct?

Steven-Eardley: You can remove the `else` so the `return none` covers both cases (if that's intended)

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+        for f in filenames:
+            fp = os.path.join(dirpath, f)
+            total_size += os.path.getsize(fp)
+    return total_size
+
+
+def _prune_container():
+    # Delete all files and dirs in the container that does not contain today's date
+    for dir in mainStore.list(container):
+        if today not in dir:
```
Steven-Eardley:
```suggestion
        if dates.today() not in dir:
```

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+            mainStore.delete(target_name=dir)
+
+
+def _copy_on_complete(path):
+    name = os.path.basename(path)
+    raw_size = _get_dir_size(path)
+    print("Compressing temporary file {x}".format(x=path, y=raw_size))
+    zipped_name = name + ".gz"
+    dir = os.path.dirname(path)
```
Steven-Eardley: this overwrites python's `dir`

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+    if url:
+        last_page = parse_qs(urlparse(url).query).get('page', [])
+    if last_page:
+        return int(last_page[0])
+    else:
+        return None
+
+def _get_dir_size(path):
+    total_size = 0
+    for dirpath, dirnames, filenames in os.walk(start_path):
```
Steven-Eardley:
```suggestion
    for dirpath, dirnames, filenames in os.walk(path):
```
no variable `start_path`

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+    # Delete all files and dirs in the container that does not contain today's date
+    for dir in mainStore.list(container):
+        if today not in dir:
+            mainStore.delete(target_name=dir)
+
+
+def _copy_on_complete(path):
+    name = os.path.basename(path)
+    raw_size = _get_dir_size(path)
+    print("Compressing temporary file {x}".format(x=path, y=raw_size))
```
Steven-Eardley: you're not using value `y` that the poor little computer worked so hard to calculate.
Also, if you're not using meaningful formatting names you may as well use `"{0} and {1}".format(positional, args)`

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+
+    tmpStore = StoreFactory.tmp()
+    mainStore = StoreFactory.get("api_data")
+    container = app.config.get("STORE_ANON_DATA_CONTAINER")
+
+    if args.clean:
+        mainStore.delete(container)
+
+    # create dir with today's date
+    today = datetime.now().strftime("%Y%m%d")
```
Steven-Eardley: we already have his function, `portality.lib.dates.today()` (as above in line 47)
I think the format with dashes should still play nicely with s3, right?

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+
+if __name__ == '__main__':
+    import argparse
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-t", "--type", choices=['articles','journals', 'all'], help="type of data to export. ")
+    parser.add_argument("-c", "--clean", action="store_false", help="Clean any pre-existing output before continuing")
+    parser.add_argument("-p", "--prune", action="store_true", help="Delete previous backups if anyafter running current backup")
+    args = parser.parse_args()
+
+    tmpStore = StoreFactory.tmp()
```
Steven-Eardley: I'm not a fan of global variables in Python

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+        page = 1
+        last_page = 1
+        while page <= last_page:
+            print("journals page {p} of {l}".format(p=page, l=last_page))
+            try:
+                results = DiscoveryApi.search_journals(query, page, page_size, sort, bulk=True)
+            except DiscoveryException as e:
+                raise Api400Error(e.message)
+            _save_file(typ='journals')
+            new_last_page = _get_last_page()
```
Steven-Eardley: as my comment regarding globals above, I'd much prefer `_get_last_page()` to take `results` as an argument

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+            if new_last_page is not None:
+                last_page = new_last_page
+            page = page + 1
```
Steven-Eardley:
```suggestion
            page += 1
```

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+    zipped_path = os.path.join(dir, zipped_name)
+    tar = tarfile.open(zipped_path, "w:gz")
+    tar.add(path)
+    tar.close()
+    zipped_size = os.path.getsize(zipped_path)
+    print("Storing from temporary file {x} ({y} bytes)".format(x=zipped_name, y=zipped_size))
+    mainStore.store(container, zipped_name, source_path=zipped_path)
+    tmpStore.delete(container, name)
+    tmpStore.delete(container, zipped_name)
+    return
```
Steven-Eardley: another unnecessary return statement

portality/scripts/api_export.py
```diff
@@ -0,0 +1,135 @@
...
+            new_last_page = _get_last_page()
+            if new_last_page is not None:
+                last_page = new_last_page
+            page = page + 1
+
+    print(dates.now() + ": done\n")
+
+    # Copy the source directory to main store
+    out_dir = tmpStore.path(container, today)
+    _copy_on_complete(out_dir)
```
Steven-Eardley: do we need a `try:` here? It'd be a shame to discard our work from the temp store if the compression or save fails in `_copy_on_complete()`

anusharanganathan:
@Steven-Eardley this PR is a work in progress. I guess I should have mentioned it. I raised the PR so Richard could easily review what I had done. We did this on late Wednesday evening. I haven't yet gone through all of your comments but will do so as I am doing the work

anusharanganathan:
I have split this issue into 2 parts split, so it's easier to manage the coding and review process. Also, most of the code here will move to the feature relating to removing deep paging. See [#1750](https://github.com/DOAJ/doajPM/issues/1750) and [#1751](https://github.com/DOAJ/doajPM/issues/1751)

PR #1584, opened 2018-11-12
Feature/1693 file duplicate checking

richard-jones:
READY FOR FINAL REVIEW AND MERGE - pending go-ahead for go-live.

This is the PR that covers 1693 and 1698 as well as works towards 1296.

In order to run the tests you will need the latest version of Combinatrix, which is set up in requirements.txt

Note that on deploy to production, a migration must be run:

    python portality/upgrade.py -u portality/migrate/20180808_1693_article_duplication/migrate.json


portality/upgrade.py
```diff
@@ -94,13 +98,24 @@ def do_upgrade(definition, verbose):
...
-                print "writing ", len(batch), "to", tdef.get("type")
+                total += len(batch)
+                print "writing ", len(batch), "to", tdef.get("type"), ";", total, "of", max
                 esprit.raw.bulk(tconn, batch, type_=tdef.get("type"))
                 batch = []
+                # do some timing predictions
+                batch_tick = datetime.now()
+                time_so_far = batch_tick - type_start
+                seconds_so_far = time_so_far.total_seconds()
+                estimated_seconds_remaining = ((seconds_so_far * max) / total) - seconds_so_far
```
Steven-Eardley: why isn't this `((seconds_so_far / total) * max) - seconds_so_far`?

i.e. you know how many you've done, you estimate how long it's taken per item and guess the full batch time from that.

portality/view/publisher.py
```diff
@@ -222,9 +224,12 @@ def metadata():
...
                 return render_template('publisher/metadata.html', form=form, author_error=True)
             else:
-                xwalk = article.FormXWalk()
+                # xwalk = article.FormXWalk()
+                xwalk = ArticleFormXWalk()
                 art = xwalk.crosswalk_form(form)
-                art.save()
+                articleService = DOAJ.articleService()
+                articleService.create_article(art, current_user._get_current_object())
+                # art.save()
```
Steven-Eardley: articleService.create_article does indeed save the article if it's successful, these commented out lines should be removed.

portality/crosswalks/article_doaj_xml.py
```diff
@@ -0,0 +1,260 @@
...
+            raise CrosswalkException(message="Unable to validate document with identified schema", inner_message=msg)
+
+        return doc
+
+    def validate(self, doc):
+        valid = self.schema.validate(doc)
+        if not valid:
+            el = self.schema.error_log.__repr__()
+            # strip the filename, as we don't want to leak the path to the UI
+            rx = "[\da-f]{32}.xml:(.*)"
```
Steven-Eardley: this is a regex - should it be in the regexes file?

portality/bll/exceptions.py
```diff
@@ -42,4 +44,36 @@ class ConfigurationException(Exception):
...
\ No newline at end of file
+    pass
+
+class DuplicateArticleException(Exception):
+    """
+    Exception to raise when a duplicate article is detected, and this is not permitted
+    """
+    pass
+
+class IngestException(Exception):
```
Steven-Eardley: should this be a more generalised traceable exception that we can use elsewhere?

portality/bll/services/article.py
```diff
@@ -0,0 +1,375 @@
...
+        fail = 0
+        update = 0
+        new = 0
+        all_shared = set()
+        all_unowned = set()
+        all_unmatched = set()
+
+        for article in articles:
+            result = self.create_article(article, account, duplicate_check=duplicate_check, merge_duplicate=merge_duplicate, limit_to_account=limit_to_account, dry_run=True)
+            success += result.get("success", 0)
```
Steven-Eardley: since we enforce the structure of `result` below, we shouldn't need to do this careful value extraction. No harm done, though!

portality/bll/services/article.py
```diff
@@ -0,0 +1,375 @@
...
+                # available in the index
+                articles[i].save(blocking=block)
+
+            # return some stats on the import
+            return report
+        else:
+            raise exceptions.IngestException(message=Messages.EXCEPTION_ARTICLE_BATCH_FAIL, result=report)
+
+
+    def _batch_contains_duplicates(self, articles):
```
Steven-Eardley:
```suggestion
    def _batch_contains_duplicates(self, articles):
        dois = [a.get_normalised_doi() for a in articles]
        dup_doi = len(set(dois)) != len(dois)
        
        fts = [a.get_normalised_doi() for a in articles]
        dup_ft = len(set(fts)) != len(fts)
        
        return dup_doi or dup_ft
```

hmm, mine always loops through articles twice, yours could halt early too. You win!

portality/bll/services/article.py
```diff
@@ -0,0 +1,375 @@
...
+            if duplicate is not None:
+                if merge_duplicate:
+                    is_update  = 1
+                    article.merge(duplicate) # merge will take the old id, so this will overwrite
+                else:
+                    raise exceptions.DuplicateArticleException()
+
+        # finally, save the new article
+        if not dry_run:
+            article.save()
```
Steven-Eardley: can the save() method throw up any surprises leading to our success response being a lie?

Steven-Eardley:
From codeship we have:

```
======================================================================
FAIL: test_01_discover_duplicates_45_46 (test_bll_article_discover_duplicates.TestBLLArticleDiscoverDuplicates)
----------------------------------------------------------------------
File "/home/rof/.pyenv/versions/2.7.15/lib/python2.7/site-packages/parameterized/parameterized.py", line 392, in standalone_func
return func(*(a + p.args), **p.kwargs)
File "/home/rof/src/github.com/DOAJ/doaj/doajtest/unit/test_bll_article_discover_duplicates.py", line 154, in test_01_discover_duplicates
svc.discover_duplicates(article, owner_id)
AssertionError: ValueError not raised
```

doajtest/fixtures/article.py
```diff
@@ -78,15 +78,57 @@ def make_article_source(eissn=None, pissn=None, with_id=True, in_doaj=True, with
         for idx in delete:
             del source["bibjson"]["identifier"][idx]
 
+        if doi is not None:
+            set_doi = False
+            for i in range(len(source["bibjson"]["identifier"])):
```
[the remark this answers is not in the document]
richard-jones: Yes, I have no idea why the code is like that.  Probably I copied it from somewhere :)

portality/lib/normalise.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    if url is None:
+        return url
+
+    schemes = ["http", "https", "ftp", "ftps"]
+    url = url.strip()
+    if url.startswith("//"):
+        url = "http:" + url
+
+    if "://" not in url:
```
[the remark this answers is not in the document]
richard-jones: this is intended behaviour, no need for a fix

portality/migrate/20180808_1693_article_duplication/README.md
```diff
@@ -0,0 +1,13 @@
+# 2018-08-08; Issue 1693 - Article Duplication
+
+This migration covers the deployment of code to bring the article objects in the datastore up to standard
+with the structure required for consistend deduplication.
```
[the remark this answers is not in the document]
richard-jones: done

portality/upgrade.py
```diff
@@ -51,7 +51,11 @@ def do_upgrade(definition, verbose):
     for tdef in definition.get("types", []):
         print "Upgrading", tdef.get("type")
         batch = []
-        for result in esprit.tasks.scroll(sconn, tdef.get("type"), keepalive=tdef.get("keepalive", "1m")):
+        total = 0
+        first_page = esprit.raw.search(sconn, tdef.get("type"))
+        max = first_page.json().get("hits", {}).get("total", 0)
```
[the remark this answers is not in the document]
richard-jones: I'm fine with that, won't fix :)

portality/upgrade.py
richard-jones: They are mathematically equivalent :)

portality/view/publisher.py
richard-jones: done

portality/crosswalks/article_doaj_xml.py
richard-jones: hm, I'm not sure.  I had thought the regex module would be for common re-usable regexes, like for DOIs, etc.  This regex is completely specific to this exact point in the code, so I don't think I would move it out.  What do you think?

portality/bll/exceptions.py
richard-jones: as discussed, maybe but not now.

portality/bll/services/article.py
```diff
@@ -0,0 +1,375 @@
...
+            new += result.get("new", 0)
+            all_shared.update(result.get("shared", set()))
+            all_unowned.update(result.get("unowned", set()))
+            all_unmatched.update(result.get("unmatched", set()))
+
+        report = {"success" : success, "fail" : fail, "update" : update, "new" : new, "shared" : all_shared, "unowned" : all_unowned, "unmatched" : all_unmatched}
+
+        # if there were no failures in the batch, then we can do the save
+        if fail == 0:
+            for i in xrange(len(articles)):
```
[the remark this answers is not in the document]
richard-jones: I decided to do it the python3 compatible way in the end.  The performance boost is virtually immesurable here, and this is asynchronous code anyway, so saves us one hassle when upgrading to p3 later.

portality/bll/services/article.py
richard-jones: The new version of the save method does some fairly aggressive error handling and reporting, so if save fails, even after several re-tries, we would expect an exception.  I don't think there's any further handling that we need to do here - the exception would cause the front end to get an error, but that's probably what we want.

portality/crosswalks/article_doaj_xml.py
Steven-Eardley: Yup, keep it here.

portality/bll/services/article.py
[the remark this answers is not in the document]
Steven-Eardley: lovely

richard-jones:
I've fixed up that test, and will check in a new version of this branch as soon as the rest of the tests have run locally.

portality/upgrade.py
```diff
@@ -106,7 +106,7 @@ def do_upgrade(definition, verbose):
             # When we have enough, do some writing
             if len(batch) >= batch_size:
                 total += len(batch)
-                print "writing ", len(batch), "to", tdef.get("type"), ";", total, "of", max
+                print datetime.now(), "writing ", len(batch), "to", tdef.get("type"), ";", total, "of", max
```
Steven-Eardley: wow, what a way to format a string! (it was already like that so I'll let it pass)

PR #1586, opened 2018-11-26
change the email recipients when update requests are rejected or requ…

richard-jones:
…ire revisions.  were going to the journal contact, now go to the owner


portality/formcontext/formcontext.py
```diff
@@ -676,13 +677,14 @@ def finalise(self):
 
         # if this application is being accepted, then do the conversion to a journal
         if self.target.application_status == constants.APPLICATION_STATUS_ACCEPTED:
+            # remember whether this was an update request or not
```
Steven-Eardley: unnecessary duplication

richard-jones:
I didn't make the changes on 510 because I also raised an issue to review all the sending rules.  We'll remove it then, or re-enable it then, depending on what happens I guess

portality/formcontext/formcontext.py
richard-jones: meh, maybe.  is_update_request isn't required in all the if/else statements, so in some code paths this would be unnecessarily executed.  Six of one, half a dozen of the other.  I prefer it this way myself, just for clarity - each if/else branch is self-contained in its behaviour.

Steven-Eardley: Okay, we'll keep it as-is!

Steven-Eardley:
I'll merge and deploy this now.

Steven-Eardley:
Released. I couldn't find an issue to update about it, though.

richard-jones:
This was related to https://github.com/DOAJ/doajPM/issues/1786 - I've updated the issue now

PR #1587, opened 2018-11-29
Feature/1853 refactor discovery search query

anusharanganathan:
Refers to issue #1853

The [failing test](https://github.com/DOAJ/doajPM/issues/1853#issuecomment-442796638) needs to be fixed before this PR can get merged.

### DIFF
--- a/portality/api/v1/discovery.py
+++ b/portality/api/v1/discovery.py
@@ -1,5 +1,4 @@
 from portality.api.v1.common import Api
-from portality import models
 from portality import util
 from portality.core import app
 from datetime import datetime
@@ -7,6 +6,7 @@ import esprit
 import re, json, uuid, os
 from copy import deepcopy
 from flask import url_for
+from portality.bll.doaj import DOAJ
 
 class DiscoveryException(Exception):
     pass
@@ -99,11 +99,11 @@ class DiscoveryApi(Api):
     @staticmethod
     def get_application_swag():
         return deepcopy(DISCOVERY_API_SWAG['application'])
-    
+
     @staticmethod
     def get_journal_swag():
         return deepcopy(DISCOVERY_API_SWAG['journal'])
-    
+
     @staticmethod
     def get_article_swag():
         return deepcopy(DISCOVERY_API_SWAG['article'])
@@ -154,26 +154,17 @@ class DiscoveryApi(Api):
         return q, page, fro, page_size, sortby, sortdir
 
     @classmethod
-    def _make_search_query(cls, q, page, page_size, sort, search_subs, sort_subs):
+    def _make_query(cls, index_type, account, q, page, page_size, sort, search_subs, sort_subs):
         # sanitise and prep the inputs
         q, page, fro, page_size, sortby, sortdir = cls._sanitise(q, page, page_size, sort, search_subs, sort_subs)
 
         # assemble the query
-        query = SearchQuery(q, fro, page_size, sortby, sortdir)
-        # print json.dumps(query.query())
+        search_query = SearchQuery(index_type, account, q, fro, page_size, sortby, sortdir)
+        query, dao_klass = search_query.query()
+        print json.dumps(query)
 
-        return query, page, page_size
+        return dao_klass, query, page, page_size
 
-    @classmethod
-    def _make_application_query(cls, account, q, page, page_size, sort, search_subs, sort_subs):
-        # sanitise and prep the inputs
-        q, page, fro, page_size, sortby, sortdir = cls._sanitise(q, page, page_size, sort, search_subs, sort_subs)
-
-        # assemble the query
-        query = ApplicationQuery(account.id, q, fro, page_size, sortby, sortdir)
-        # print json.dumps(query.query())
-
-        return query, page, page_size
 
     @staticmethod
     def _calc_pagination(total, page_size, requested_page):
@@ -190,7 +181,7 @@ class DiscoveryApi(Api):
 
         page_count = ((total - 1) // page_size) + 1
         last_page = FISRT_PAGE + page_count - 1
-        
+
         # Links to previous and next page
         if requested_page > FISRT_PAGE:
             previous_page = requested_page - 1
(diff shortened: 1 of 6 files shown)


portality/api/v1/discovery.py
```diff
@@ -301,72 +294,17 @@ def __init__(self, qs, fro, psize, sortby=None, sortdir=None):
...
-            "size" : self.psize
-        }
-
         if self.sortby is not None:
             q["sort"] = [{self.sortby : {"order" : self.sortdir, "mode" : "min"}}]
 
-        return q
\ No newline at end of file
+        query_service = DOAJ.queryService()
+        query, dao_klass = query_service.prepare_query('api_query', self.index_type, q, self.account)
```
richard-jones: After pondering it for a bit, I don't think this bit of code should be in this place.  The idea of the Query objects is they are a bit like Prepared Statements in SQL and by adding a second part to the return, and an integration with the queryService, it feels like this class is over-stretching its intention.  Instead I would have this class return the pure search query that it *wants* to run, and then in the caller (`_make_search`) add this call to the query service to constrain the query and get the class.

portality/api/v1/discovery.py
```diff
@@ -239,59 +230,61 @@ def _make_response(cls, endpoint, res, q, page, page_size, sort,
     def search_articles(cls, q, page, page_size, sort=None):
```
richard-jones: All these functions `search_articles` `search_journals` and `search_applications` are now extremely similar.  They pretty much only differ by the config values that they depend on.  So, we can probably refactor them so that there is a single function that does most of the work, and they just pass in the specifics.  That would minimise code duplication.

portality/bll/services/query.py
```diff
@@ -117,6 +126,15 @@ def search(self, domain, index_type, raw_query, account, additional_parameters):
 
         return res
 
+    def prepare_query(self, domain, index_type, raw_query, account):
```
richard-jones: After a bit of reflection, I'm against this approach.  The reason being that it violates the encapsulation of the queryService.  The whole point of the query service is that it does all the work for you, and if we prepare a query and then execute it outside that service, then we risk duplicating code and introducing unexpected bugs.  

In this case, for example, this means that something that uses the `prepare_query` method misses applying the post-search filters that would have been applied if the main `search` method was used.  Right now that wouldn't matter as there aren't any post search filters, but were we to add any in the settings, they would not be applied like they would with all the other query endpoints.

Instead what we should do here is modify the `search` query so that `additional_parameters=None` in the signature, then just use `queryService.search` directly in the API code - I made a note above somewhere where I'd do that.

portality/bll/services/query.py
```diff
@@ -80,34 +80,43 @@ def _post_filter_search_results(self, cfg, res):
...
         # the class that will handle it
         dao_name = cfg.get("dao")
         dao_klass = plugin.load_class(dao_name)
         if dao_klass is None:
             raise exceptions.NoSuchObjectException(dao_name)
+        return dao_klass
 
-        # add any required filters to the query
-        query = self._pre_filter_search_query(cfg, query)
+    def search(self, domain, index_type, raw_query, account, additional_parameters):
```
richard-jones: as per my comment below, I think we should change this to have `additional_parameters=None` so that we can bypass parameter validation if needed (which it would be for the API)

portality/api/v1/discovery.py
```diff
@@ -239,59 +230,61 @@ def _make_response(cls, endpoint, res, q, page, page_size, sort,
...
         search_subs = app.config.get("DISCOVERY_ARTICLE_SEARCH_SUBS", {})
         sort_subs = app.config.get("DISCOVERY_ARTICLE_SORT_SUBS", {})
-        query, page, page_size = cls._make_search_query(q, page, page_size, sort, search_subs, sort_subs)
+        dao_klass, query, page, page_size = cls._make_query('article', None, q, page, page_size, sort, search_subs, sort_subs)
 
         # execute the query against the articles
-        res = models.Article.query(q=query.query(), consistent_order=False)
+        res = dao_klass.query(q=query.as_dict(), consistent_order=False)
```
richard-jones: I think this should actually look more like this, to preserve the encapsulation provided by the queryService:

```
query = self._make_query(....)
res = articleService.search("api_query", "[type]", query, account)
```

portality/api/v1/discovery.py
richard-jones: Although, see my comments about not using `prepare_query` in the first place too :)

portality/api/v1/discovery.py
anusharanganathan: Have replaced with a common search function

portality/bll/services/query.py
anusharanganathan: Function removed

richard-jones:
I'm just having a look at this test now; I think what we need to do is provide a test context with a suitable user for the test to pass.  Just trying to remember how to do that now, and what the best way is when testing the API.

richard-jones:
This error is raising a bit of an issue with how we limit to specific accounts; we have a bit of a dichotomy in how the appropriate account to use is handled.

On the one hand the query filters all use `current_user` which is the user flask has logged in through the front end.  The query_service, meanwhile, uses an explicitly passed `account` field.  

For the most part these are the same thing, but the possibility exists for them to be different (e.g. do a search as if you are user X, while logged in as user Y).

I fixed the issue with the test by setting a request context with a suitably logged in user, but the test failed again further down when trying to search in the same context as another user.

We can easily fix the test by changing the request context, but I want to spend a bit of time thinking about this, as there's a chance we could inadvertently introduce an authNZ bug here in the future, and we should avoid that.

richard-jones:
BTW, this problem existed before, you have just unearthed it!

richard-jones:
I have pushed a change which fixes this test.  The issue about the disconnect between the query filters and the query service still exists, but I don't think that should stop this PR going through.  Instead I'm going to raise an issue for that separately.

PR #1590, opened 2018-12-10
explicitly order notes for the application and journal forms

richard-jones:
This sets the display order of notes on application and journal edit pages to newest first, and makes new notes appear at the top, as required by #1671

Before merging this we need to make sure users are ready for the change to the UI involved.


richard-jones:
I assume you closed this because the hotfix isn't against the right branches?  I'll remake this as a feature branch shortly

PR #1598, opened 2019-01-31
Some significant enhancements to the applications CRUD API route:

richard-jones:
* improve and correct some of the error reporting, so the messages are more usable
* fix issue where applications required both ISSNs to successfully import
* fix issue where only first keyword was being imported


richard-jones:
Just doing a bit more review of this, and fixing a couple of broken tests, not ready for merge yet

richard-jones:
ok this is ready to go now.  All tests are passing locally.

richard-jones:
@Steven-Eardley - added those unit test extensions we talked about

PR #1602, opened 2019-02-20
Hotfix/1857 email recipients revisited

richard-jones:
set journal contact and account holder where present as the recipients for in progress and editor assigned emails

### DIFF
--- a/portality/formcontext/emails.py
+++ b/portality/formcontext/emails.py
@@ -91,32 +91,42 @@ def send_publisher_editor_assigned_email(application):
 
     contact_name = application.get_latest_contact_name()
     contact_email = application.get_latest_contact_email()
-    send_list = [(contact_name, contact_email,
-                  Messages.SENT_JOURNAL_CONTACT_ASSIGNED_EMAIL,
-                  Messages.NOT_SENT_JOURNAL_CONTACT_ASSIGNED_EMAIL)]
+    send_list = [
+        {
+            "name" : contact_name,
+            "email" : contact_email,
+            "sent_alert" : Messages.SENT_JOURNAL_CONTACT_ASSIGNED_EMAIL,
+            "not_sent_alert" : Messages.NOT_SENT_JOURNAL_CONTACT_ASSIGNED_EMAIL
+        }
+    ]
 
     owner = models.Account.pull(application.owner)
     if owner is not None:
-        send_list.append((owner.name, owner.email,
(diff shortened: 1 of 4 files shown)


portality/formcontext/emails.py
```diff
@@ -200,25 +213,36 @@ def send_publisher_inprogress_email(application):
...
+    send_list = [(contact_name, contact_email,
+                  Messages.SENT_JOURNAL_CONTACT_IN_PROGRESS_EMAIL,
+                  Messages.NOT_SENT_JOURNAL_CONTACT_IN_PROGRESS_EMAIL)]
 
-    # This is to the publisher contact on the application
-    publisher_name = owner.name
-    publisher_email = owner.email
+    owner = models.Account.pull(application.owner)
+    if owner is not None:
+        send_list.append((owner.name, owner.email,
```
Steven-Eardley: I find this a bit of a weird data structure to use for this - you have a tuple of name, email, and 2 messages - but these are unlabelled and undocumented.

portality/formcontext/formcontext.py
```diff
@@ -1056,15 +1052,11 @@ def finalise(self):
...
-                except app_email.EmailException:
-                    magic = str(uuid.uuid1())
-                    self.add_alert('Hm, sending the ready status to publisher email didn\'t work. Please quote this magic number when reporting the issue: ' + magic + ' . Thank you!')
-                    app.logger.exception('Error sending ready status email to publisher - ' + magic)
+                alerts = emails.send_publisher_inprogress_email(self.target)
+                for alert in alerts:
+                    self.add_alert(alert)
             else:
-                self.add_alert('Did not send email to Journal Contact about the status change, as publisher emails are disabled.')
+                self.add_alert('Did not send email to Owner or Journal Contact about the status change, as publisher emails are disabled.')
```
Steven-Eardley: should this be a Message?

portality/formcontext/emails.py
richard-jones: Fair enough, switching to using a dict for tracking the values.

portality/formcontext/formcontext.py
richard-jones: Yeah, should be.  There's a lot in that file that should be, but I'll stick to only moving stuff out as we touch it.  So, I've moved this one :)

Steven-Eardley:
This one is ready to go. Can't release until the article duplication migration is complete - in a few hours.

PR #1605, opened 2019-02-26
Hotfix/2019 02 21 iss1905 6 oai errors

Steven-Eardley:
Fixes for bugs reported in https://github.com/DOAJ/doajPM/issues/1905 - and some initial work on OAI-PMH tests.

### DIFF
--- a/portality/view/oaipmh.py
+++ b/portality/view/oaipmh.py
@@ -397,7 +397,7 @@ def list_identifiers(dao, base_url, specified_oai_endpoint, metadata_prefix=None
         return _resume_list_identifiers(dao, base_url, specified_oai_endpoint, resumption_token=resumption_token)
 
 
-def _parameterised_list_identifiers(dao, base_url, specified_oai_endpoint, metadata_prefix=None, from_date=None, until_date=None, oai_set=None, start_number=0, start_after=None):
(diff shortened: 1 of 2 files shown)


Steven-Eardley:
@richard-jones this is ready for your review with an aim to release soon, and I'll work on the parameterised tests subsequently - I think we may as well have the fixes live in the short term.

portality/view/oaipmh.py
```diff
@@ -444,42 +444,66 @@ def _parameterised_list_identifiers(dao, base_url, specified_oai_endpoint, metad
...
+            # Determine where our next starting index will be
+            new_start = start_number + len(results)
+
+            # Work out if we need a resumption token.  It can have one of 3 values:
+            # - None -> do not include the rt in the response if we have a full result set
+            # - the empty string -> include in the response if this is the last set of results from an incomplete list
+            # - some value -> include in the response if there are more values to retrieve
+            if len(results) == full_total:
+                resumption_token = None
+            elif new_start == full_total:
```
richard-jones: I'll be honest, I don't understand this whole method now, with the various total calculations.  If there are tests that cover all the angles, and they work, then I'm happy.

What pulled me up here is that == seems very specific, wouldn't we want to send the empty string resumption token if the new_start was >= full_total.  Maybe that can't ever happen, I can't tell from the code :)

Steven-Eardley: Good point, I'll change it to >=. If it acts in the way I think it does, it'll make no difference.

Steven-Eardley:
Released 2019-03-13

PR #1612, opened 2019-03-27
Feature/1835 article metadata in journal csv

richard-jones:
READY FOR REVIEW AND MERGE
(we should specify a release date before we do the merge)

On the surface of it, this PR provides two new columns to the Journal CSV: "Number of Article Records" and "Most Recent Article Added"

Underneath, we have taken the opportunity to do some refactoring, and the following is now done:

* The CSV generation has been moved to the BLL
* The BackgroundJob has been refactored to use the BLL
* Testing has been migrated to use parameterised tests
* The file is now stored on and served from S3

In terms of migration, all that is required is to deploy the new version, and then generate the Journal CSV via the script (or wait until it is generated automatically next, which is every 30 minutes).  The old journal csv directory should also be removed from the appdata directory on the server, as this is no longer used.

At deployment time a new S3 bucket needs to be set up and configured into the application, this is documented in docs/system/S3.md in the section "Cached Files".  I have already created the bucket "doaj-data-cache", and a user "doaj-data-cache-client" - you may want to deactivate the current security credentials and generate some new ones for this user for absolute certainty over the security.

IMPORTANT NOTE - this PR also changes the default names of the storage buckets on S3, to help avoid test environments accidentally writing to the live buckets.  So, at deployment time the following configuration needs to be set in production config:

```
STORE_ANON_DATA_CONTAINER = "doaj-anon-data"
STORE_CACHE_CONTAINER = "doaj-data-cache"
STORE_PUBLIC_DATA_DUMP_CONTAINER = "doaj-data-dump"
```

Note that in between the deployment and the CSV being generated again, the /csv URL will 404.


dommitchell:
@Steven-Eardley Tuesday (today) works for me.

Steven-Eardley:
Slightly bumpy release (3 mins downtime due to misconfiguring app.cfg) but it's released and working now. Verified the metadata is in the CSV.

PR #1613, opened 2019-03-27
remove constraint on removing owner of application, except when appli…

richard-jones:
…cation is in status accepted

NOT FOR MERGE just yet, just want to confirm we are good to go ahead.


richard-jones:
We have put this aside for the time being, the issue https://github.com/DOAJ/doajPM/issues/1939 has been taken off the project board.  Probably we will revisit it when we do the Editorial Workflow changes.  This PR is likely to be obsolete and will not be resurrected

PR #1615, opened 2019-04-03
remove all references to addthis

richard-jones:


Steven-Eardley:
Does this require a widget refresh?

richard-jones:
> Does this require a widget refresh?

Yes, good point, I'll do that in a second

richard-jones:
> > Does this require a widget refresh?
> 
> Yes, good point, I'll do that in a second

Actually, it turns out no, it doesn't need a rebuild.  The addthis feature is only on the site's facetviews not on the FQW's one.

dommitchell:
HOORAY!

PR #1617, opened 2019-04-11
Hotfix/20190409 1357 issue1952

richard-jones:
Fixes the error message when the publisher submits an ISSN that has not previously been seen.

This also turned up a bug with detecting unmatched ISSNs for publishers that only have one journal in DOAJ, and that has also been fixed.

This patch works by shifting the responsibility for adding journal metadata to article records from the crosswalk to the create_article method in the BLL.  This means that the parameterised tests have also been updated to cover adding journal metadata in both the create_article and batch_create_article methods.

In addition, the error message for an unexpected error has been changed from "file system error" to "unanticipated error", and the documentation updated to match, and some additional advice on how to report the issue added.


portality/bll/services/article.py
```diff
@@ -157,21 +171,21 @@ def is_legitimate_owner(self, article, owner):
...
             journals = models.Journal.find_by_issn(issn)
             if journals is not None and len(journals) > 0:
                 for j in journals:
                     owners.append(j.owner)
-                    if j.owner not in seen_issns:
-                        seen_issns[j.owner] = []
-                    seen_issns[j.owner] += j.bibjson().issns()
+                    if j.owner not in seen_journal_issns:
+                        seen_journal_issns[j.owner] = []
+                    seen_journal_issns[j.owner] += j.bibjson().issns()
```
Steven-Eardley: Much clearer!

portality/bll/services/article.py
```diff
@@ -184,15 +198,17 @@ def is_legitimate_owner(self, article, owner):
...
         # single owner must still know of all supplied issns
-        compare = list(set(seen_issns[owners[0]]))
-        if len(compare) == 2:   # we only want to check issn parity for journals where there is more than one issn available.
-            for issn in issns:
-                if issn not in compare:
-                    return False
-
-        # true if the found owner is the same as the desired owner, otherwise false
-        return owners[0] == owner
+        journal_issns = set(seen_journal_issns[owners[0]])
```
Steven-Eardley: we've already established that `owners[0] == owner`, so they're equivalent. I'd prefer to use the singular `owner` here, but it doesn't really matter.

portality/templates/doaj/publishers.html
```diff
@@ -567,11 +567,19 @@ <h4><a name="Error occurred ingesting the records in the document"></a>Error occ
...
-<p><strong>Resolution</strong>: Please <a href="{{ url_for('doaj.contact') }}">contact us</a> immediately with the relevant details.</p>
+<p><strong>Resolution</strong>: Please <a href="{{ url_for('doaj.contact') }}">contact us</a> immediately with the details:</p>
+
+<ol>
+    <li>Your user account</li>
+    <li>When the incident occurred</li>
+    <li>Information about your environment, such as which browser you were using</li>
+</ol>
+
+<p>We may also ask you to supply the file that you tried to upload, so please keep a copy of it until the issue is resolved.</p>
```
Steven-Eardley: Good plan.

PR #1618, opened 2019-04-11
add one year timeout to consent cookie

richard-jones:
Sets cookie consent cookie for one year

### DIFF
--- a/portality/view/doaj.py
+++ b/portality/view/doaj.py
@@ -46,7 +46,8 @@ def cookie_consent():
         resp = redirect(cont)
     else:
         resp = make_response()
-    resp.set_cookie(app.config.get("CONSENT_COOKIE_KEY"), Messages.CONSENT_COOKIE_VALUE)
+    # set a cookie that lasts for one year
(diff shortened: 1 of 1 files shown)


portality/view/doaj.py
```diff
@@ -46,7 +46,8 @@ def cookie_consent():
         resp = redirect(cont)
     else:
         resp = make_response()
-    resp.set_cookie(app.config.get("CONSENT_COOKIE_KEY"), Messages.CONSENT_COOKIE_VALUE)
+    # set a cookie that lasts for one year
+    resp.set_cookie(app.config.get("CONSENT_COOKIE_KEY"), Messages.CONSENT_COOKIE_VALUE, max_age=31536000)
```
Steven-Eardley: Will this fix having to re-consent when we deploy? I haven't established what breaks it, whether it's coincidence that I have to agree every time I update the system because that's when I check it, or whether our cookie is different each time.

richard-jones: I haven't had to reconsent at all so far, so it must be a coincidence.  If we are also having to reconsent after deploy we'll have to look at that again.

PR #1619, opened 2019-04-12
Feature/fix codeship tests

richard-jones:
This came about as I was getting frustrated with the codeship tests randomly failing.

Almost always, it's just an initialisation/data loading delay with ES, so I have been running the tests on codeship over and over and seeing what breaks, then putting in code to handle that.

This PR is by no means all of the fixes, and I'll just keep adding stuff to this (or a future branch once this one is merged) every time I see a codeship build fail, until we've got them all.


richard-jones:
sounds good to me

PR #1622, opened 2019-04-30
Pass account model objects to the BLL from the bulk API rather than L…

Steven-Eardley:
…ocalProxy from current_user

https://github.com/DOAJ/doajPM/issues/1955

Fix only, tests later.

I've looked at the other uses of current user in similar contexts, and they look OK - it's just when we pass from current_user to the BLL which has argvalidate.


Steven-Eardley:
merging since we discussed this in-meeting.

PR #1624, opened 2019-05-01
fix bug with store deleting all files when it should only delete one …

richard-jones:
…from S3

This also removes a bunch of logging as we don't really need all of it, but keeps some back for future reference.

### DIFF
--- a/portality/store.py
+++ b/portality/store.py
@@ -50,8 +50,12 @@ class Store(object):
     def url(self, container_id, target_name):
         pass
 
-    def delete(self, container_id, target_name=None):
+    def delete_container(self, container_id):
         pass
+
+    def delete_file(self, container_id, target_name):
+        pass
+
 
 class StoreS3(Store):
     """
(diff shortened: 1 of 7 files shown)


portality/store.py
```diff
@@ -112,6 +112,11 @@ def delete(self, container_id, target_name=None):
         # we are not allowed to delete the bucket, so we just delete the contents
         keys = self.list(container_id)
 
+        if target_name is not None:
```
Steven-Eardley: Ahah, I now see where we were deleting the entire contents of the bucket!

I don't really like this delete function yet, though - it makes sense to check whether the delete target object is in the bucket before we issue a delete request via the client, but I think the default behaviour of `target_name=None` is potentially dangerous since it will still clear the bucket. If that's the intention, it should at least be noted in a function docstring!

richard-jones: Yes, that is the intention - you can use this method to delete a specific file, or the entire bucket.  What would you do instead?  The obvious alternative is to separate delete of a target from delete of a bucket at the method level.  If we want to make that change, now is a good time!

So, for example

```
store.delete_container(container_id)
store.delete_file(container_id, target_name)
```

richard-jones: I have gone ahead and done this, just doing some more testing to make sure everything works as expected, will update this PR shortly.

Steven-Eardley: Fantastic, that's pretty much what I had in mind - I think it's sensible to keep that clear and separate. Documenting the behaviour would've been enough, though!

richard-jones:
This change is now pushed.  I have updated the tests and all that is passing.  That only tells us that the StoreLocal and all the calling code is working as expected.  I have also manually tested the S3 connector against a test S3 bucket, and was able to confirm that I can add multiple files, delete them individually, and delete them all together through expected use of the StoreS3 connector.

PR #1626, opened 2019-05-01
fix some test race conditions, so that they don't fail on codeship

richard-jones:


richard-jones:
Note that the build is failing at the moment because the develop branch tests are broken until https://github.com/DOAJ/doaj/pull/1624 is merged

PR #1634, opened 2019-05-17
Feature/1977 duplicate articles by edit

richard-jones:
Closes the loophole for duplicate articles to be created by edit as per https://github.com/DOAJ/doajPM/issues/1977


portality/bll/services/article.py
```diff
@@ -50,12 +50,16 @@ def batch_create_articles(self, articles, account, duplicate_check=True, merge_d
...
-                                         add_journal_info=add_journal_info,
-                                         dry_run=True)
+            try:
+                result = self.create_article(article, account,
+                                             duplicate_check=duplicate_check,
+                                             merge_duplicate=merge_duplicate,
+                                             limit_to_account=limit_to_account,
+                                             add_journal_info=add_journal_info,
+                                             dry_run=True)
+            except exceptions.ArticleMergeConflict:
```
Steven-Eardley: Do we also want to handle the `exceptions.DuplicateArticleException()` that `create_article()` can also throw when `merge_duplicate` is False?

richard-jones: There's no current situation where the DuplicateArticleException is raised in a context where it matters.  The API and the IngestArticlesBackgroundTask both specify (by omission) to merge duplicates, so there's no need to catch the exception there.

The only other place this is used is in the tests, and they don't need the error caught.

This option exists in case we want to be able to run these kinds of operations from scripts or from the command line, and in that case I think it's OK for the DuplicateArticleException to bubble up and cease execution.

So, I don't think I would change this code, no.

PR #1635, opened 2019-05-22
fix bug in anon export to match up with updated store api

richard-jones:


richard-jones:
Test failures are unrelated to the change, just another thing to fix in the codeship build

PR #1638, opened 2019-06-20
Feature/1296 article deduplicate scripting

richard-jones:
The PR is for some minor enhancement to the script, but once it is merged we need to run the penultimate data cleanup for this task.

Please run

```
DOAJENV=production python portality/scripts/articlerm.py -c [path to csv]
```

The csv to use as the source for this operation is here
[actions-2019-06-20.csv.txt](https://github.com/DOAJ/doaj/files/3310624/actions-2019-06-20.csv.txt)


portality/scripts/article_duplicate_analysis.py
```diff
@@ -465,6 +495,87 @@ def _read_match_set(reader, next_row):
...
+    with codecs.open(report_out, "wb", "utf-8") as ro:
+        writer = clcsv.UnicodeWriter(ro)
+        writer.writerow(["account", "articles to delete", "article_details"])
+        for k, v in accounts.iteritems():
+            fn = k + "_articles.csv"
+            with codecs.open(os.path.join(articles_dir, fn), "wb", "utf-8") as a:
+                awriter = clcsv.UnicodeWriter(a)
+                awriter.writerow(["DOI", "Fulltext", "Reason for removal", "Number of duplicated articles"])
+                dedupe = []
+                for article in v:
```
Steven-Eardley: I'm finding this loop a little tricky to review. Can you explain it?

richard-jones: We're writing two related CSVs at the same time:

1. lists the account name, the number of articles being deleted, and the location of the CSV listing those articles
2. lists the identifers, reason for removal, and number of articles sharing those identifiers.  There is one CSV of this type per account.

These CSVs will be used by the DOAJ team to communicate with the publishers listed in spreadsheet (1) with the details in the related second spreadsheet.

Above we've gathered the data on the accounts and the articles that will be deleted.  This is in a datastructure like:

```
accounts = {
    "account_name" : [
        ["doi", "fulltext", "reason"],
        ...
    ]
}
```

All this loop is doing is flattening that structure so that a number of rows can be written to the second sheet, and a single row can be written to the first sheet.  The only additional complexity is that since ids may be repeated in each set of id records, they must first be deduplicated.  So by the end of the loop, you have a variable called `dedupe` which contains:

```
[
    ["unique doi 1", "unique fulltext 1", "reason", "count of records with these identifiers"],
    ["unique doi 2", "unique fulltext 2", "reason", "count of records with these identifiers"]
]
```

These rows are then written to the second spreadsheet.

Finally, the `account_name` referred to as `k`, the number of records `len(v)`, and the filename `fn` of the second spreadsheet are written to the first.  We use the number of records `len(v)` rather than `len(dedupe)` because the former tells us how many actual articles there are.

Does that help?

Steven-Eardley: fantastic, thank you!

richard-jones:
ok, cool, can we merge, deploy and run the script?

richard-jones:
or can I do that?  I don't think that I can deploy yet.

Steven-Eardley:
I'll deploy and run it, just checking the build can pass

Steven-Eardley:
tests are failing consistently - would you like to look into that or shall I?

doajtest/unit/test_bll_article_get_duplicates.py
```diff
@@ -119,3 +119,58 @@ def test_01_get_duplicates(self, name, kwargs):
...
+        import time
+        time.sleep(2)
+
+        # check that either way round we look at it, we now have a duplicate
+        # in the system
+        dups = self.svc.get_duplicates(a1)
+        assert len(dups) == 1
+
+        dups = self.svc.get_duplicates(a2)
+        assert len(dups) == 1
```
Steven-Eardley: Tests fail on this line - `dups` is `[]`

richard-jones: ah, probably it just needs a rebase

richard-jones: I can look at it if you like, I'll try to get to it this afternoon

richard-jones: Ok, this should be fixed.  The test was used to demonstrate an error, and by passing showed the problem.  Now that it is failing, the problem is fixed, and the checks were already incorporated into the main test, so I have just removed the experimental test.

PR #1640, opened 2019-06-26
Update gitignore

sssoz:
✨

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Local env file
 .env
+env/
 
 .brackets.json
 *.pyc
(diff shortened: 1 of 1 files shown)


Steven-Eardley:
Happy to merge this - if you want to keep your virtualenv within the project directory, feel free!

Note, we'd normally expect feature branches to be prefixed with `feature/` and be against `develop`, and a hotfix branch would be prefixed `hotfix/` and would be against `master`. The `git-flow` utility will make that stuff straightforward. It's written up in our wiki: https://github.com/DOAJ/doajPM/wiki/Development-Process-for-Code

I'll treat this as a feature branch and merge immediately, though :)

PR #1644, opened 2019-06-28
Removing duplicated subject

glauberm:
The subjects `[TK9001-9401] Nuclear engineering. Atomic power` and `[TA401-492] Materials of engineering and construction. Mechanics of materials` are duplicated under different parents. I believe they're being override, as only one of each appears to be on use on the website. They also only appear under one parent [on the LOC website](https://www.loc.gov/aba/cataloging/classification/lcco/lcco_t.pdf).

### DIFF
--- a/lccSubjects.xml
+++ b/lccSubjects.xml
@@ -1514,11 +1514,6 @@
     <parent>Chemical technology</parent>
   </subject>
   <subject>
-    <name>Materials of engineering and construction. Mechanics of materials</name>
-    <code>TA401-492</code>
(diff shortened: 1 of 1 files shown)


richard-jones:
Thanks for this, and for drawing our attention to the duplication.  We haven't merged this yet (and may not for a while) because it has some down-stream consequences that we need to take into account, and a little bit more coding/migration planning needs to be done.  We have put it into our project issue tracker, and will leave this PR open until we've got that resolved.

PR #1650, opened 2019-07-18
Link to the full text of article use DOI when URL is not specified.

amdomanska:
Changes are made in the article page and facetviews both for search page and list of articles of a journal.

Fixes: https://github.com/DOAJ/doajPM/issues/1961


portality/app.py
```diff
@@ -237,7 +237,7 @@ def doi_url(doi):
     :return: the HTML link
     """
     tendot = doi[doi.find('10.'):]
-    return "<a href='https://doi.org/{0}'>{0}</a>".format(tendot)
+    return "https://doi.org/{0}".format(tendot)
```
richard-jones: This is the same place where we spotted that issue yesterday around ascii characters.  If we're changing this line now, maybe worth checking for that issue and fixing it at the same time.

amdomanska: Could you please remind me what issue that is about? Do we have an issue open?

portality/templates/doaj/article.html
```diff
@@ -208,9 +208,17 @@ <h1>&lt; Article title missing &gt;</h1>
...
-                {% endfor %}
+                {% if bibjson.get_urls("fulltext") %}
+                    {% for url in bibjson.get_urls("fulltext") %}
+                        | <a href="{{url}}">Full Text</a>
+                    {% endfor %}
+                {% else %}
+                    {% if doi %}
+                        {% set tendot = doi[doi.find('10.'):] %}
+                        {% set href = "https://doi.org/{0}".format(tendot) %}
+                        <a href="{{ href }}"> Full Text </a>
```
Steven-Eardley: This can be simplified - we don't need to set `tendot` here in the template because of the changes you made to the template filter in `app.py`. It can look a lot more like the doi link generation above.

portality/app.py
richard-jones: We talked about this out of band, but for reference it's this issue: https://github.com/DOAJ/doajPM/issues/2023

richard-jones: Actually, on reflection, I think we need to do a bit more here.  As I mentioned in the other PR I just reviewed, there is a function `portality.lib.normalise.normalise_doi` and we should probably use that inside `doi_url`, so that we have a normalised DOI going on to the `https://doi.org/` prefix.

richard-jones:
gah, sorry @Steven-Eardley I accidentally re-requested review from you.  I meant to update my review, not sure what happened!

amdomanska:
@richard-jones Could have a look if these are changes that you meant? Thanks!
@Steven-Eardley I've cleaned it up.

portality/templates/doaj/article.html
```diff
@@ -87,7 +87,7 @@ <h1>&lt; Article title missing &gt;</h1>
 
         {# doi #}
         {% if doi %}
-            DOI {{ doi | doi_url | safe}}
+            DOI <a href="{{ doi | doi_url | safe }}"> {{ doi }}</a>
```
richard-jones: We changed the behaviour below to handle if the DOI is non canonical, and returns an empty string, so we need to do something similar here.

Probably the best thing to do is normalise the DOI once at the top, and then it will be easy just to check for a non empty string wherever you might display the DOI.

amdomanska: the doi_url is using normalized doi already as it is changed in the app.py method. So we can only check if the string is not empty and do not display it if so. Question is if we want to display doi as it is or normalized here (everywhere?)

portality/static/doaj/js/available_facetviews/public.journalarticle.facetview.js
```diff
@@ -473,7 +474,7 @@ jQuery(document).ready(function($) {
 
         // extract the fulltext link if there is one
         var ftl = false;
-        if (resultobj.bibjson && resultobj.bibjson.link) {
+        if (resultobj.bibjson && resultobj.bibjson.link.length != 0) {
```
richard-jones: Hey, sorry, late comment.  I was just porting this over to the new UI code and I spotted a couple of minor issues.  I think this line should actually read:

```
if (resultobj.bibjson && resultobj.bibjson.link && resultobj.bibjson.link.length !== 0) {
```

So, check for `resultobj.bibjson.link` first before looking for the `.length` property, and then the comparison should be `!==` rather than `!=` (due to JS weirdness with the `!=` operator)

portality/templates/doaj/article.html
```diff
@@ -208,9 +212,18 @@ <h1>&lt; Article title missing &gt;</h1>
...
-                {% endfor %}
+                {% if bibjson.get_urls("fulltext") %}
+                    {% for url in bibjson.get_urls("fulltext") %}
+                        | <a href="{{url}}">Full Text</a>
+                    {% endfor %}
+                {% else %}
+                    {% if doi %}
+                        {% set url_text = normalised_doi | doi_url | safe %}
+                        {% if url_text != "" %}
+                            <a href="{{ normalised_doi | doi_url | safe }}"> Full Text</a>
```
richard-jones: For a slight improvement to performance here, this should be:

```
<a href="{{ url_text }}"> Full Text</a>
```

since you created it just 2 lines above.

amdomanska: ah, oversight. Sorry for that, ready in 5 minutes.

PR #1651, opened 2019-07-18
Add DOI metadata where available

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/1579

### DIFF
--- a/portality/templates/doaj/article.html
+++ b/portality/templates/doaj/article.html
@@ -25,6 +25,7 @@
 
 {% if jtitle %}<meta name="citation_journal_title" content="{{ jtitle }}">{% endif %}
 {% if jbib.publisher %}<meta name="citation_publisher" content="{{ jbib.publisher }}">{% endif %}
+{% if doi %}<meta name="citation_doi" content="{{ doi }}">{% endif %}
 
 {% for author in bibjson.author %}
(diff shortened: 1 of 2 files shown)


amdomanska:
It's the issue I did for an interview. Do you need me to change the branch name for the proper feature/1651_.... ?

portality/templates/doaj/article.html
```diff
@@ -25,6 +25,7 @@
 
 {% if jtitle %}<meta name="citation_journal_title" content="{{ jtitle }}">{% endif %}
 {% if jbib.publisher %}<meta name="citation_publisher" content="{{ jbib.publisher }}">{% endif %}
+{% if doi %}<meta name="citation_doi" content="{{ doi }}">{% endif %}
```
richard-jones: Just a thought, we should consider normalising this DOI like we have done in the other PR https://github.com/DOAJ/doaj/pull/1650

We could use the `doi_url` filter here once that PR has been merged, which will give us a normalised DOI url.  The alternative is to use `portality.lib.normalise.normalise_doi` to generate a canonical form of it (which won't be a URL but will be an industry standard DOI).

If we take the latter option note that `normalise_doi` can throw an exception, and jinja2 has no exception framework, so we might need to add a template filter wrapper to it like `doi_url` to catch and manage that exception.

PR #1654, opened 2019-07-22
Encoding error on article landing pages #2023

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2023


Steven-Eardley:
@amdomanska and I have tested this locally for one of the problem articles, so I'll send it live now.

PR #1657, opened 2019-07-31
check if user is not anonymous

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2030


Steven-Eardley:
Released 2019-08-04

PR #1658, opened 2019-08-02
add url validation to file uploader

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2041
connected to: https://github.com/DOAJ/doajPM/issues/2038

Simple url validation. Tries to parse url and checks if all parts are included: 
scheme://netloc/path

Examples:
Valid:
http://www.journals.aoftw.edu.uk/view/1202/article.pdf
http://www.cwi.nl:80/%7Eguido/Python.html
http://doaj.org/index

Invalid:
www.journals.aoftw.edu.uk/view/1202
http://doaj.org
journals.aoftw.edu.uk/view/1202
//journals.aoftw.edu.uk/view/1202

Invalid URL causes processing failure with message "Invalid URL: url"

### DIFF
--- a/portality/crosswalks/article_doaj_xml.py
+++ b/portality/crosswalks/article_doaj_xml.py
@@ -5,6 +5,7 @@ from portality.bll import exceptions
 from portality.crosswalks.exceptions import CrosswalkException
 from portality import models
 from datetime import datetime
+from urlparse import urlparse
 
 class DOAJXWalk(object):
     format_name = "doaj"
@@ -215,7 +216,16 @@ class DOAJXWalk(object):
         if ftel is not None and ftel.text is not None and ftel.text != "":
             ct = ftel.get("format")
             url = ftel.text
-            bibjson.add_url(url, "fulltext", ct)
+
+            try:
+                # url validation
+                result = urlparse(ftel.text)
+                if all([result.scheme, result.netloc, result.path]):
+                    bibjson.add_url(url, "fulltext", ct)
+                else:
+                    raise CrosswalkException(message="Invalid URL: {}".format(ftel.text), inner=e)
+            except Exception as e:
+                raise CrosswalkException(message="Invalid URL: {}".format(ftel.text), inner=e)
 
         # keywords
         keyel = record.find("keywords")


amdomanska:
I've tried examples above but if you could think about any additional examples it would be good to test as many options as we can.

Steven-Eardley:
Your solution is really well written, but in this case I'd recommend we enforce the correct URL syntax in our [XSD Schema](https://github.com/DOAJ/doaj/blob/develop/portality/static/doaj/doajArticles.xsd). Currently, we require this to be `anyURI`:

```xml
   <xs:element name="fullTextUrl">
    <xs:complexType> 
     <xs:simpleContent>
      <xs:extension base="xs:anyURI">
       <xs:attribute name="format" />  
      </xs:extension>
     </xs:simpleContent>
    </xs:complexType>
   </xs:element>
```

Which allows for various other relative identifiers as well as URLs. See http://www.datypic.com/sc/xsd/t-xsd_anyURI.html

We could introduce a regex pattern in the schema to validate this field so we don't have to do it in code and handle the exception. Another advantage is that users can validate the data themselves before sending it to us. There may be limitations to the regular expression syntax, so that should be investigated. The ideal scenario would be to use the same or equivalent regex as the dataobj struct for validating Journals.

richard-jones:
A couple of thoughts, I'm not sure of the best order, so here they are in a random one:

* I'd +1 what @Steven-Eardley suggests, getting this into the XSD would be great.  I'm a bit cautious about having a reliable regex that can hit all URLs, but perhaps it's enough that it starts with http(s)://?

* I don't think that this validation should be in the crosswalk - the crosswalk is just about converting data from one form to another, and if we introduce validation there too it is confusing matters

* I would still like to know why the lower-level object validation isn't triggered here.  Is it just because that validation error on the API is triggered by the API model objects when they validate?  If so, then fair enough we should also just include this validation in the XSD.  I'd love it if there was a single validation route, though, that covered both ways in to the system, to give us consistent data.  I wouldn't mind if it did validation at both ends, for convenience, but to have a single canonical validation approach would be good.  (I know that this might not be possible!)

richard-jones:
I've just part-answered my own last point - yes, it's the API's data objects which are doing the URL validation.  The core model object does not validate.  So, that's fine, let's just get this validation into the XSD or a layer very very close to it (i.e. before the crosswalk runs).

In terms of centralised global validation, let's not try to bite that off here.

richard-jones:
Also, can we close this now, as there's a replacement PR?

PR #1659, opened 2019-08-05
url validation for xml uploader v2

amdomanska:
Fixes: DOAJ/doajPM#2041
connected to: DOAJ/doajPM#2038
connected to: https://github.com/DOAJ/doaj/pull/1658

@Steven-Eardley this the version 2 of above PR with regex validation in xsd as you asked.

Add requirement to DOAJArticles.xsd. Regex check:
https://regex101.com/r/xHzUZC/1


richard-jones:
For this I'd like to add some tests to the unit test suite that tries to submit URLs in correct/incorrect forms.

Also, I'm not 100% sure how `[a-zA-Z]` works but we need to remember that URLs are now internationalised, and we may get URLs with unicode characters in them.  So we should try some of those and make sure that the regex lets them in.

portality/static/doaj/doajArticles.xsd
```diff
@@ -130,15 +130,13 @@
...
-     <xs:simpleContent>
-      <xs:extension base="xs:anyURI">
-       <xs:attribute name="format" />  
-      </xs:extension>
-     </xs:simpleContent>
-    </xs:complexType>
+
+         <xs:element name="fullTextUrl">
+    <xs:simpleType>
+     <xs:restriction base="xs:string">
```
Steven-Eardley: I think we can add the regex to the `xsd:anyURI` type, so we get some extra validation (such as no double `#` characters according to the docs)

So we should keep the type as anyURI rather than changing it to a string.

amdomanska:
it now accept ascii characters

amdomanska:
@richard-jones you have regex quite well explained here: https://regex101.com/r/xHzUZC/1
The only difference is substituting [a-zA-Z] with [\p{L}] that matches any letter from any alphabet (see here: https://www.regular-expressions.info/unicode.html in "Unicode Categories" section)

amdomanska:
I added tests for:
1) correct url starting with http
2) correct url starting with https
3) correct url that has some non-ascii characters (from different alphabets)
4) incorrect url in form: //something.org
5) incorrect url with http missing in form: www.example.com

 They're in the same test suite as the previous ones that check also mistakes in the articles (eg no id).

portality/static/doaj/doajArticles.xsd
```diff
@@ -130,15 +131,13 @@
...
-      <xs:extension base="xs:anyURI">
-       <xs:attribute name="format" />  
-      </xs:extension>
-     </xs:simpleContent>
-    </xs:complexType>
+
+         <xs:element name="fullTextUrl">
+    <xs:simpleType>
+     <xs:restriction base="xs:anyURI">
+      <xs:pattern value="https?://([^/:]+\.[\p{L}]{2,10}|([0-9]{1,3}\.){3}[0-9]{1,3})(:[0-9]+)?(/.*)?"/>
```
Steven-Eardley: This looks like a well designed regex, but we might be being overly prescriptive for this use case.

That said, I notice it's very similar to the URL regex we use in https://github.com/DOAJ/doaj/blob/develop/portality/formcontext/fields.py#L5 - and in the validator here: https://github.com/DOAJ/doaj/blob/develop/portality/formcontext/validate.py#L187

The fulltext field in the article metadata form, however, uses WtForms' `validators.url`:
```python
regex = r'^[a-z]+://(?P<host>[^/:]+)(?P<port>:[0-9]+)?(?P<path>\/.*)?$'
```

Interestingly, none of the existing examples use the syntax to allow unicode in URLs, but I'm not sure if that's taken care of by Python elsewhere.

Steven-Eardley: It's also full of unecessary capture groups `()` - but that does keep it fairly close to our existing examples.

I'm torn between completely stripping this down to something like `^https?://.+` and letting the `anyURI` type handle the rest of the validation, or leaving it as it's written here and trust it to do its job since the tests pass.

amdomanska: The regex was intentionally done (as discussed with @richard-jones ) as an equivalent of the one from https://github.com/DOAJ/doaj/blob/develop/portality/formcontext/fields.py#L5

Urls from the forms are later checked if they start with the http(s) via API - so I added that check to the regex to ensure it works the same as API. The last change is using \p{L] instead of [a-zA-Z] that ensures accepting also non-unicode characters and is xml regex feature as described here:   https://www.regular-expressions.info/unicode.html in "Unicode Categories" section. I can check if python itself allowes unicode in regex as we should probably change the regex from formcontext/fields to one accepting them as well.

richard-jones: I've been bashing at this regex a bit, and it seems to work well, with only two edge cases that we might need to take into account.  Neither of the following validate, but I think that they would be considered valid:

http://example.com?q=500&whatever=true
http://example.com#this

If you insert "/" after ".com" then these do validate.  If you put them in a browser, the browser redirects to the version with the "/" inserted, which would suggest that is the preferred form.  I'm cautious about introducing something that would possibly throw an error on an otherwise usable URL though.

doajtest/unit/test_api_bulk_application.py
```diff
@@ -185,7 +185,7 @@ def test_05_test_via_endpoint(self):
                 first_apl = reply.pop()
                 assert first_apl['status'] == 'created'
                 # Check we actually created new records
-                time.sleep(1)
+                time.sleep(1.5)
```
richard-jones: I guess you did this because you were having problems getting the test to pass, due to the race condition here.  In this instance I think this is fine - the test doesn't give you much alternative, but just to note for information that eventually we want to squeeze all these `time.sleep` calls out.

doajtest/unit/test_models.py
```diff
@@ -1099,7 +1099,7 @@ def test_33_article_stats(self):
             articles.append(article)
 
         for i in range(len(articles)):
-            articles[i].save(blocking=i == len(articles) - 1)
+            articles[i].save(blocking=True)
```
richard-jones: This could make the test quite a bit slower, and to solve this problem we did recently introduce an alternative approach.  Instead you can now do:

`Article.blockall([(a.id, a.last_updated) for a in articles])`

That will make sure that all articles are available in search without having to block each one individually.

doajtest/unit/test_api_bulk_application.py
Steven-Eardley: Yup, that was me, seemed to help codeship. Agreed on the `time.sleep`s

richard-jones:
@Steven-Eardley and I talked about the regex today, and agreed that we **should** use the longer more precise regex because of the parity with the WTForms regex for the same purpose.  We just need to modify the regex to support those two URLs that are currently not validating, which I think just means making it:

```
https?://([^/:]+\.[\p{L}]{2,10}|([0-9]{1,3}\.){3}[0-9]{1,3})(:[0-9]+)?(.*)?
```

That is, removing the final "/" match as a requirement.  I'm not 100% sure this doesn't break it in other ways so we need to check it again with the tests, but if that works then I think we're there.

@Steven-Eardley introduced the test issues, and he's going to push a change to fix them.

amdomanska:
I'll make the change and add tests for these two edge cases as well to be sure that if the tests pass we have all possible urls covered

amdomanska:
I've changed the url so that the http anchor and parameters are accepted without slash in front of them. I added the tests for these 2 cases as well.

doajtest/unit/test_models.py
Steven-Eardley: The problem here is that because they are new articles, there is no `last_updated` set. This causes the following exception:

```python
Error
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/steve/code/cl/doaj/doajtest/unit/test_models.py", line 1101, in test_33_article_stats
    models.Article.blockall([(a.id, a.last_updated) for a in articles])
  File "/home/steve/code/cl/doaj/portality/dao.py", line 731, in blockall
    cls.block(id, lu, sleep=sleep, max_retry_seconds=individual_max_retry_seconds)
  File "/home/steve/code/cl/doaj/portality/dao.py", line 707, in block
    threshold = datetime.strptime(last_updated, "%Y-%m-%dT%H:%M:%SZ")
TypeError: strptime() argument 1 must be string, not None
```

I've solved this by applying a last_updated of now():
```models.Article.blockall([(a.id, datetime.now().strftime('%Y-%m-%dT%H:%M:%SZ')) for a in articles])```

Comment (Steven-Eardley) [reply doajtest/unit/test_models.py]: Unless the missing save is applied. Oops. Incoming.

File: portality/crosswalks/article_doaj_xml.py [hunk 5]
```diff
@@ -215,7 +216,16 @@ def crosswalk_article(self, record, add_journal_info=True):
...
+
+            try:
+                # url validation
+                result = urlparse(ftel.text)
+                if all([result.scheme, result.netloc, result.path]):
+                    bibjson.add_url(url, "fulltext", ct)
+                else:
+                    raise CrosswalkException(message="Invalid URL: {}".format(ftel.text), inner=e)
+            except Exception as e:
+                raise CrosswalkException(message="Invalid URL: {}".format(ftel.text), inner=e)
```
richard-jones: I think we need to revert the changes to this file, so that the crosswalk is not involved in url validation at all

amdomanska: done

PR #1660, opened 2019-08-08
Prevent invalid url from raising DataStructureException

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2038

This is a workaround that prevents invalid url from raising DataStructureException (as option 1. in @richard-jones comment here: https://github.com/DOAJ/doajPM/issues/2038#issuecomment-518259336. I believe we should take that into account only as a temporary solution to let the articles with invalid url being retrieved via api as raising this Exception is expected behavior (thus test_04_coerce from test_api_crud_article is expected to fail - for now it needs to be disabled then). I am also not entirely sure if my approach is best here, maybe there is a better way to ensure retrieving problematic articles correctly - I'd be grateful for any suggestions.

### DIFF
--- a/portality/templates/doaj/publishermembers.html
+++ b/portality/templates/doaj/publishermembers.html
@@ -22,14 +22,13 @@
 <li><a href="http://ojs.diffundit.com/">Diffundit Diseño &amp; Comunicación, S.L.</a></li>
 <li><a href="http://www.edpsciences.org/">EDP Sciences</a></li>
 <li><a href="http://electric.press/">Electric Press</a>, Helen J Burgess and Craig Saper Co-editors</li>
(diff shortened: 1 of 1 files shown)


amdomanska:
As a note:

Looks like ES does not really let us to search all the articles with bibjson.link[0].type["fulltext"] with missing "http(s)". Citation from @markmacgillivray:

> ES makes it very easy to search, not not necessarily so easy to find specific objects by values, especially once those values end up in lists of objects

Thus creating the script that searches all articles resulting with this error (as proposed by @richard-jones in p2 here  DOAJ/doajPM#2038 (comment) is non-trivial if not impossible.

PR #1662, opened 2019-08-25
Prevent invalid url from raising DataStructureException

amdomanska:
Fixes: DOAJ/doajPM#2038
Continuation: https://github.com/DOAJ/doaj/pull/1660

Changes made according to this comment: https://github.com/DOAJ/doaj/pull/1660#pullrequestreview-274292043


amdomanska:
Sorry for the new PR. The previous one got closed automatically by GH when I reverted branch's state to the original.

amdomanska:
@richard-jones I've tried that for a few documents (both with correct and incorrect URLs) and confirmed with the existing set of UT.

richard-jones:
@Steven-Eardley - I think we can queue this one up for deployment

portality/api/v1/data_objects/article.py
```diff
@@ -48,14 +48,6 @@
                     }
                 },
 
-                "link": {
```
Steven-Eardley: I think this should still be in, like:

```
"link": {
    "fields": {
        "type": {"coerce": "link_type"},
        "url": {"coerce": "unicode"},
        "content_type": {"coerce": "link_content_type"}
    }
},
```
Note `url` has been replaced by `unicode`. This is a more minimal change to the struct to allow bad URLs out of the system. I don't want to go too far down the route of making this model permissive.

In general, if we have a big problem with bad data causing 500 errors on outgoing articles, we should catch that exception when we are enforcing the outgoingarticleDO and give a JSON response saying we can't give it out.

richard-jones: Agreed.  I pondered this same point for a while and went the other way, but if @Steven-Eardley thinks it should be in I'd +1 that.

doajtest/unit/test_api_crud_article.py
```diff
@@ -164,7 +164,7 @@ def test_04_coerce(self):
 
         # now test some failures
 
-        # an invalid urls
+        # an invalid urls - WE DON'T WANT THIS EXCEPTION TO BE RAISED ANYMORE
```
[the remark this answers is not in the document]
amdomanska: Yep, sorry, it's an oversight from the previous version that should be deleted earlier

amdomanska:
Not a problem ofc! :)

amdomanska:
This is the API response for retrieving the article that contains invalid URL. 

image

portality/api/v1/crud/articles.py
```diff
@@ -123,6 +123,7 @@ def retrieve_swag(cls):
         template['responses']['200']['schema'] = IncomingArticleDO().struct_to_swag(schema_title='Article schema')
         template['responses']['401'] = cls.R401
         template['responses']['404'] = cls.R404
+        #template['responses']['500'] = cls.R500
```
Steven-Eardley: Is this supposed to be commented out?

Issue #735, opened 2015-05-29
Discovery API - public endpoint


emanuil-tolev:
- [x] define url structure (versioning /api/v1/, paths for public endpoints)
- [x] public search over journals and articles

emanuil-tolev:
https://github.com/CottageLabs/oaspectrum#search-api

emanuil-tolev:
After #783 is merged you can go ahead with this @richard-jones - I've pretty much put in the structure of this route as an example, and even written a custom json encoder so you can concentrate on the DiscoveryApi class returning living DOAJ models - the new jsonify_models encoder will read their .data attribute and convert them to JSON for you, so no conversion stuff for you to do, just return model objects. The swagger yaml I've written is also correct so feel free to build on it.

One note - I've put the search query as a path parameter instead of the usual GET param. so /search/article/computers%20are20good%20for%20you is the way to call it. This is somewhat down to taste, but basically since the URL is already url-encoded I didn't think there was any point in dealing with request params, might as well take it from the path (and it makes error handling a bit easier, e.g. Flask will just issue 404 when the search query is missing because it won't match the route).

You will need to change @blueprint.route('/search/<search_type>/<search_query>') to @blueprint.route('/search/<search_type>/<search_query:path>') in the code I've put down in order to allow forward slashes, but after that you should be away..

emanuil-tolev:
> One note - I've put the search query as a path parameter instead of the usual GET param

but feel free to change it back to a ?q GET param if *you* find something wrong with that

richard-jones:
There are going to be other url parameters (paging and page size) so we don't get away with not having to deal with them.  As you say, a matter of taste, and perhaps it shows that I'm old-fashioned in that I prefer the query args :)

emanuil-tolev:
ah yeah. OK, definitely feel free to change it then.

For swagger docs, instead of ```in:path``` use ```in:body``` to describe your params in that case.

richard-jones:
Actually I spent a bit of time consulting the Gods Of REST (the internet), and concluded that actually, while still a matter of taste ultimately, in the path is the sensible place to put the query!

emanuil-tolev:
and the page size and page number is in the args?

richard-jones:
@emanuil-tolev - assigned to you now this is finished, but maybe we can just close?

Nimphal:
@richard-jones should the API be returning a JSON 404 ? Right now it redirects to a Not found page. And are there any other return codes?

[...]

Issue #516, opened 2015-01-13
Serve static files with timestamp in the query string to prevent caching


emanuil-tolev:
Best way of illustrating what this means:

http://doaj.org/static/doaj/js/customise_facetview.js - will get cached. Then if the file changes it will take a while for changes to show up for all users.

http://doaj.org/static/doaj/js/customise_facetview.js?version=20150113093600 - will get cached. Then if the file changes the automatically generated version parameter will change, and will force everybody to update their JS.

Flask-Assets might be able to do this. We should probably do this as part of the public view of data project and get rid of those pesky Ctrl+Reload moments. /cc @richard-jones

emanuil-tolev:
Initially I had the idea of doing this via an asset manager, but we can just change the links we generate and it will be a very effective cache buster! See 2 example links above.

richard-jones:
Did this as part of PVOD

PR #1663, opened 2019-08-27
Fixed timestamp returned by discovery API to be ISO-8601 compatible

Steven-Eardley:
Replaced timestamp line to use `dates` in `portality.lib`:

```python
def now_with_microseconds():
    return format(datetime.utcnow(), format="%Y-%m-%dT%H:%M:%S.%fZ")
```
Additional code style changes.

Note, this is a **BREAKING CHANGE** to the API since users may have created integrations with the dodgy timestamp format. Is a new version required?

timestamp format bug was reported here:
https://github.com/ropensci/jaod/issues/10

### DIFF
--- a/portality/api/v1/discovery.py
+++ b/portality/api/v1/discovery.py
@@ -1,8 +1,8 @@
 from portality.api.v1.common import Api
 from portality import util
 from portality.core import app
+from portality.lib import dates
 from portality import models
-from datetime import datetime
 import esprit
 import re, json, uuid, os
 from copy import deepcopy
@@ -10,12 +10,15 @@ from flask import url_for
 from portality.ui.messages import Messages
 from portality.bll.doaj import DOAJ
 
+
 class DiscoveryException(Exception):
     pass
+
 
 class SearchResult(object):
     def __init__(self, raw=None):
         self.data = raw if raw is not None else {}
+
 
 def query_substitute(query, substitutions):
     if len(substitutions.keys()) == 0:
@@ -56,6 +59,7 @@ def query_substitute(query, substitutions):
 
     return ":".join(subs)
 
+
 def allowed(query, wildcards=False, fuzzy=False):
     if not wildcards:
         rx = "(.+[^\\\\][\?\*]+.*)"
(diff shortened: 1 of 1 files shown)


katrinleinweber:
Thanks for this fix :-) It doesn't seem urgent, so how about following [semver.org's advice of introducing a deprecation warning in vX.Y.Z and then changing the behaviour in vX+1.0.0](https://semver.org/#how-should-i-handle-deprecating-functionality)?

katrinleinweber:
OTOH, given that [this bug went unreported for 4 years](https://github.com/DOAJ/doaj/blame/8567e803fdd525ae8053aacb7a33cdca4882798d/portality/api/v1/discovery.py#L241-L249), either:

a) nobody relied on that output, and therefore no big fuzz is necessary, or 
b) people did rely on it, invested extra time in a workaround instead of reporting the bug, and therefore deserve no special care now ;->

katrinleinweber:
Since it's also been several years since #516, and neither #735, nor #787 add a second argument for the `timestamp`, could its necessity maybe be re-evaluated in general? Do people still need to be told the time that their query was answered?

richard-jones:
Ah, an old typo, by the looks of it.

I'm in two minds about just deploying this.  it is, as you say, a breaking change, though I would guess that no one has ever parsed that date, otherwise they would have noticed that it was skew.  My inclination would be to deploy it as is, without any warning or deprecations.  We don't make any claims anywhere as to the format of this date, it's just a string, so anyone's date parsing ought to be robust enough to handle it.

sckott:
we don't parse the `timestamp` field in the R client, so not a breaking change for the client at least; I've no idea what our users do though.

katrinleinweber:
Thanks for fixing this. Assuming from a lack of [reactions](https://github.com/search?q=language%3Ar+jaod+-repo%3Acran%2Fjaod+-repo%3Aropensci%2Fjaod&type=Code) [elsewhere](https://github.com/search?q=language%3Ar+doaj+-repo%3Acran%2Fjaod+-repo%3Aropensci%2Fjaod&type=Code) There seem to have been no adverse effects :-)

PR #1664, opened 2019-09-02
Feature/1889 support crossref

amdomanska:


amdomanska:
Test comment

doajtest/fixtures/__init__.py
```diff
@@ -2,7 +2,8 @@
 from applications import ApplicationFixtureFactory
 from editors import EditorGroupFixtureFactory
 from accounts import AccountFixtureFactory
-from article import ArticleFixtureFactory
+from article_doajxml import DoajXmlArticleFixtureFactory
```
richard-jones: I noticed that you changed `ArticleFixtureFactory` for `DoajXmlArticleFixtureFactory`.  I think that makes sense for when you are asking the factory for DOAJXML documents, but many calls, like `make_incoming_api_source` are unrelated to DOAJXML, so I think we should keep the `ArticleFixtureFactory` for those, and only use the `DoajXmlArticleFixtureFactory` for actual XML fixtures.

doajtest/functional/bulk_api.py
```diff
@@ -37,7 +37,7 @@
 # you'll change all of them
 dataset = []
 for i in range(0, 10):
-    dataset.append(deepcopy(ArticleFixtureFactory.make_incoming_api_article()))
+    dataset.append(deepcopy(DoajXmlArticleFixtureFactory.make_incoming_api_article()))
```
richard-jones: So, related to my earlier comment, this method `make_incoming_api_article` returns JSON that's formatted according to the API, so I don't think it should be on `DoajXmlArticleFixtureFactory`.

doajtest/mocks/bll_article.py
```diff
@@ -26,7 +26,7 @@ def discover_duplicates(cls, doi_duplicates=0, fulltext_duplicates=0, overlap=0)
 
         possible_duplicates = {"doi" : [], "fulltext" : []}
         for i, ident in enumerate(idents):
-            source = ArticleFixtureFactory.make_article_source(eissn="1234-5678", pissn="9876-5432", doi=ident["doi"], fulltext=["fulltext"])
+            source = DoajXmlArticleFixtureFactory.make_article_source(eissn="1234-5678", pissn="9876-5432", doi=ident["doi"], fulltext=["fulltext"])
```
richard-jones: Another example, `make_article_source` returns the core article format as a python dictionary, rather than XML.

doajtest/unit/resources/crossref_article_uploads.xml
```diff
@@ -0,0 +1,1414 @@
...
+    <head>
+        <doi_batch_id>1dbb27d1030c6c9d9d-7ff0</doi_batch_id>
+        <timestamp>200504260247</timestamp>
+        <depositor>
+            <depositor_name>your name</depositor_name>
+            <email_address>your@email.com</email_address>
+        </depositor>
+        <registrant>WEB-FORM</registrant>
+    </head>
+    <body>
```
richard-jones: One thing we should do is check exactly what format the XML uploads are going to be.  Is this document taken from the Crossref XML API?  It's possible that users will unwrap the feed XML (maybe up to and including this `body` tag) before they give us the data.  It would be good if we could handle both situations (which your code might do, I haven't worked that out yet ;) )

doajtest/fixtures/__init__.py
amdomanska: I removed `ArticleFixtureFactory` at all and added `DoajXmlArticleFixtureFactory` and `CrossrefArticleFixtureFactory` instead. All tests related to articles are run for both doaj and crossref. It might be useful if we would like to make further changes to other parts of the systems to be sure nothing is broken here. I've found only 3 tests that are totally unrelated to the schema (they do not need schema as a parameter at all) and these are separate.

doajtest/unit/test_tasks_ingestCrossrefArticles.py
```diff
@@ -0,0 +1,1748 @@
...
+from urlparse import urlparse
+from lxml import etree
+
+DEFAULT_MAX_REMOTE_SIZE=262144000
+CHUNK_SIZE=1048576
+
+GET = requests.get
+
+
+class MockFileUpload(object):
```
richard-jones: I know you've just copied this from an existing test, so this is entirely reasonable.  I have recently been trying to refactor the mocks into something like a mock factory, and it would be good if we can go that way with this new test class.  Take a look in something like https://github.com/DOAJ/doaj/blob/develop/doajtest/mocks/bll_article.py for examples

portality/crosswalks/article_crossref_xml.py
```diff
@@ -0,0 +1,304 @@
...
+from lxml import etree
+import re
+from portality.bll import exceptions
+from portality.crosswalks.exceptions import CrosswalkException
+from portality import models
+
+NS = {'x': 'http://www.crossref.org/schema/4.4.2', 'j': 'http://www.ncbi.nlm.nih.gov/JATS1'}
+
+
+class CrossrefXWalk(object):
```
richard-jones: Just to check, was the conversion from Crossref XML to DOAJ article model straightforward and obvious?  Is it worth us documenting it and putting it in front of the DOAJ team to check that they're happy with the conversion?

portality/static/doaj/AccessIndicators.xsd
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
richard-jones: For ease of understanding later, I'd suggest we put all the crossref XSDs in their own directory, like `static/crossref/*.xsd`

portality/crosswalks/article_crossref_xml.py
```diff
@@ -0,0 +1,304 @@
...
+    def __init__(self):
+        self.validation_log = ""
+
+        # load the schema into memory for more efficient usage in repeat calls to the crosswalk
+        if self.schema_path is None:
+            raise exceptions.IngestException(
+                message="Unable to validate for CrossrefXWalk, as schema path is not set in config")
+
+        try:
+            schema_file = open(self.schema_path)
```
richard-jones: Just thinking out-loud, what happens here if the remotely hosted bits of the XSD aren't available for whatever reason (like the mathml.xsd can't be reached because of a network issue that's theirs not ours)?  I wonder if there's a way we can localise or cache those resources so that we don't have to go out and get them every time?

doajtest/unit/resources/crossref_article_uploads.xml
amdomanska: >  It's possible that users will unwrap the feed XML (maybe up to and including this body tag) before they give us the data.
I don't think I understand. Could you explain please? XML File Upload Form requires information which schema the file should be validated against. Crossref schema requires <doi_batch> tag as a file root.

portality/crosswalks/article_crossref_xml.py
amdomanska: I was quite obvious. I may explain it during the tech call maybe?

portality/crosswalks/article_crossref_xml.py
amdomanska: We can think about that but these are quite chained (eg mathml.xsd requires other remote resources). We would have to find a way to periodically ensure we have the newest version of these.

doajtest/unit/resources/crossref_article_uploads.xml
richard-jones: ah perfect, in which case that's all good.   I wasn't sure if the `head` and `body` bits were related to the API format.

portality/crosswalks/article_crossref_xml.py
richard-jones: Yes, so that's an interesting thought - while the Crossref schema remains fixed at a given version, the transient dependencies can change without us directly knowing?  That sounds like a good argument for taking our own copies, so the upload is stable until we decide for ourselves to do an upgrade.  But it also sounds like a pain, so difficult to know what the right answer is.

portality/templates/publisher/uploadmetadata.html
```diff
@@ -82,6 +84,8 @@ <h3>History of uploads (showing last {{previous|length}})</h3>
                     <td>
                         {% if file.schema == "doaj" %}
                             DOAJ Native XML
+                        {%  elif file.schema == "crossref" %}
+                            Crossref XML
```
richard-jones: Propose that we include the Crossref XML version number here

portality/crosswalks/article_crossref_xml.py
```diff
@@ -0,0 +1,304 @@
...
+            # more than 32766 UTF8 characters
+
+        '''
+        # keywords
+        keyel = record.find("keywords")
+        if keyel is not None:
+            for kel in keyel:
+                if kel.text != "":
+                    bibjson.add_keyword(kel.text)
+        '''
```
richard-jones: Are these commented out sections hang-overs from the copy/paste from the DOAJ XML Crosswalk.  If so, can we remove them?  Or do we need to see if we can get this data from elsewhere in the Crossref XML?

amdomanska: oh no, no, sorry, these may be removed. I overlooked that when cleaning.

amdomanska:
If crossref schema cannot be loaded due to unavailability of the remote server the proper error should be logged and the task is supposed to be re-queued.

amdomanska:
I added the raising Retry Exception with a proper message when schema cannot be loaded.

@richard-jones @Steven-Eardley it's ready for review.

amdomanska:
This branch has been migrated to python 3. Continuation: https://github.com/DOAJ/doaj/pull/1705

PR #1671, opened 2019-09-18
Script to find malformed urls

amdomanska:
Connects to: https://github.com/DOAJ/doajPM/issues/2038
Solution for: https://github.com/DOAJ/doajPM/issues/2038#issuecomment-518259336 (p. 2)

In the portality/scripts you can find articles_invalid_url.py script. Running that on our live server will give us info about which articles have invalid URL. In the output file, there is Article id, Article title, URL and info if the article is in DOAJ or not (articles in doaj are listed in the first place).

### DIFF
--- a/portality/scripts/articles_invalid_url.py
+++ b/portality/scripts/articles_invalid_url.py
@@ -38,7 +38,7 @@ if __name__ == "__main__":
 
     with codecs.open(args.out, "wb", "utf-8") as f:
         writer = UnicodeWriter(f)
-        writer.writerow(["ID", "Journal Name", "URL", "In Doaj", "Created Date"])
+        writer.writerow(["ID", "Article Title", "URL", "In Doaj", "Created Date"])
 
         for j in esprit.tasks.scroll(conn, models.Article.__type__, q=INVALID_URLS, page_size=100, keepalive='5m'):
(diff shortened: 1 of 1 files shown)


amdomanska:
Please note that this script is quite slow as most of the job is run by python on the articles, not by ES query.

portality/scripts/articles_invalid_url.py
```diff
@@ -0,0 +1,49 @@
...
+            "filter": {
+                "exists": {"field": "index.fulltext"}
+            },
+            "query": {
+                "match_all": {}
+            }
+        }
+    },
+    "sort": [{
+        "in_doaj": "desc"
```
Steven-Eardley: Does this do the ones that are in_doaj: true first? Do we want to do the ones that aren't in the DOAJ at all?

amdomanska: Yes, it does in_doaj: true first. And to be honest I don't know if we need one these that aren't in doaj at all so I just took all to check easily what is the scope of the issue.

Steven-Eardley:
This is another PR that would have benefited from being a hotfix, but I'm happy to merge it to develop and it'll be ready to run after the next release.

amdomanska:
Thank you!

PR #1676, opened 2019-09-25
Feature/2019 06 22 pycountry language codes

Steven-Eardley:
todo: Requires a bit of code from the crossref branch to fix the tests (loading the schemata locally for tests)


portality/datasets.py
```diff
@@ -1,570 +1,48 @@
-# -*- coding: UTF-8 -*-
-# the comment above is for the Python interpreter, there are Unicode
-# characters written straight into this source file
-
-import json
+import pycountry
```
richard-jones: This file is way nicer than it was, though it still (necessarily) has a lot of the stuff that gets executed on import that I'm not that keen on.  I wonder if we could just tidy it up a little further by pulling together all the properties that you might want to use after import together at the top with a bit of documentation around them.  Even if we then go on to populate those objects further down the file.  Like:

```
## These are the data structures you might want to access...
# country options - [some documentation about that]
country_options = [('', '')]
# currency options - [some documentation about that]
currency_options = [(CURRENCY_DEFAULT, CURRENCY_DEFAULT)]

# ... etc
```

portality/datasets.py
```diff
@@ -1,570 +1,48 @@
...
-    countries = json.loads(f.read())
-countries_dict = OrderedDict(sorted(countries.items(), key=lambda x: x[1]['name']))
-countries = countries_dict.items()
-
-country_options = [('','')]
+# Gather the countries with 2-character codes
+country_options = [('', '')]
 country_options_two_char_code_index = []
 
+for co in sorted(pycountry.countries, key=lambda x: x.name):
```
richard-jones: And actually, lets gather these code blocks into functions at least, and then call them, so it's a bit cleaner.  Like:

```
country_options = gather_country_options()
```

portality/migrate/20190828_regenindex_dataset_change/README.md
```diff
@@ -0,0 +1,8 @@
+*S.E. 2019-08-28*
+
+Save (and re-generate index) for Application (Suggestion) and Journals
+Refresh the journal information stored within each Article, save (and re-generate index) each Article
+
+Run
+
+    python portality/upgrade.py -u portality/migrate/20190828_regenindex_dataset_change/migrate.json
```
richard-jones: Can we also include the issue number in the migration directory name, like `20190828_1234_regen..`

portality/migrate/p1p2/country_cleanup.py
```diff
@@ -4,11 +4,12 @@
 import csv
```
richard-jones: Realistically, I'm not sure that this file would ever run or be run again.  Is there an argument for just deleting these directories when they become obsolete?  They will be in version control if we ever needed them.  Or we could leave them broken deliberately to record what it was that they looked like at the time they were run?

portality/scripts/orphaned_datasets.py
```diff
@@ -0,0 +1,57 @@
+""" Check whether the DOAJ contains records with invalid country, currency or language datasets """
```
richard-jones: Have you run this yet on a test dataset?  It would be interesting to see the result.

portality/static/doaj/iso_639-2b.xsd.old
```diff
@@ -0,0 +1,2279 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
richard-jones: Since this is version 1.0 maybe better to name it according to its version rather than `.old`?

PR #1675, opened 2019-09-25
For update requests alerts go to the account holder

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2052

I am not sure how we could test it.

PR #1686, opened 2019-10-04
Feature/2052 emails

amdomanska:
Continuation: https://github.com/DOAJ/doaj/pull/1675
Fixes: https://github.com/DOAJ/doajPM/issues/2052


amdomanska:
`make_update_request_source` is the same as `make_application_source` but with `template["admin"]["current_journal"]` set. So now fixtures have these two distinguished. I had to change `make_application_source` to `make_update_request_source` everywhere where `current_journal` is required.

doajtest/unit/test_api_crud_application.py
```diff
@@ -408,7 +408,7 @@ def test_05_outgoing_application_do(self):
         oa = OutgoingApplication()
 
         # make one from an incoming application model fixture
```
Steven-Eardley: It's still unclear to me why this has been renamed in so many places.

portality/formcontext/emails.py
```diff
@@ -85,8 +85,41 @@ def send_assoc_editor_email(obj):
...
+        {
+            "name" : owner.name,
+            "email" : owner.email,
+            "sent_alert" : Messages.SENT_PUBLISHER_ASSIGNED_EMAIL,
+            "not_sent_alert" : Messages.NOT_SENT_PUBLISHER_ASSIGNED_EMAIL
+        }
+    ]
+
+    fro = app.config.get('SYSTEM_EMAIL_FROM', 'feedback@doaj.org')
+    subject = app.config.get("SERVICE_NAME","") + " - your application has been assigned an editor for review"
```
Steven-Eardley: this says application but the method is for an update request.

portality/formcontext/emails.py
```diff
@@ -218,8 +251,40 @@ def send_editor_completed_email(application):
...
+        {
+            "name" : owner.name,
+            "email" : owner.email,
+            "sent_alert" : Messages.SENT_PUBLISHER_IN_PROGRESS_EMAIL,
+            "not_sent_alert" : Messages.NOT_SENT_PUBLISHER_IN_PROGRESS_EMAIL
+        }
+    ]
+
+    fro = app.config.get('SYSTEM_EMAIL_FROM', 'feedback@doaj.org')
+    subject = app.config.get("SERVICE_NAME", "") + " - your application is under review"
```
Steven-Eardley: this says application but the method is for an update request.

portality/formcontext/formcontext.py
```diff
@@ -760,7 +760,11 @@ def finalise(self):
         # If this is the first time this application has been assigned to an editor, notify the publisher.
         old_ed = self.source.editor
         if (old_ed is None or old_ed == '') and self.target.editor is not None:
-            alerts = emails.send_publisher_editor_assigned_email(self.target)
```
Steven-Eardley: why don't we just decide which template and subject to send in the function `send_publisher_editor_assigned_email` rather than duplicating the send functions which are almost identical?

portality/templates/email/publisher_application_inprogress.txt
```diff
@@ -1,6 +1,6 @@
 Dear {{ publisher_name }},
 
-The application or update request for "{{ journal_title }}" is now being reviewed by an Associate Editor.
+The update request for "{{ journal_title }}" is now being reviewed by an Associate Editor.
```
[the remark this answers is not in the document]
richard-jones: good spot, I missed that one.  Yes, that needs a hotfix as soon as we can.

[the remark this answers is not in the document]
Steven-Eardley: I've released a fix for it, https://github.com/DOAJ/doaj/commit/5432eb6fe15925ad0c443cab621a4736b177578b

portality/formcontext/emails.py
amdomanska: I'll prepare a hotfix for this.

portality/formcontext/formcontext.py
amdomanska: They are almost identical but it has a different subject, different template and different receiver so basically all email parameters. Sure we can merge that method and just provide parameter but that's the case for pretty almost every email send (eg. inprogress, received, revisions_required or reject emails are also almost identical) so I believe it's a matter of convention and I followed the existing one.

doajtest/unit/test_api_crud_application.py
amdomanska: Because these tests assert if `current_journal is not None` and this is true only for update requests (that's the difference between the application and update request.

portality/formcontext/formcontext.py
Steven-Eardley: Excellent point :)

doajtest/unit/test_api_crud_application.py
Steven-Eardley: cool, I'm happy with that

PR #1691, opened 2019-10-08
API Key authentication raises error if invalid API key 

amdomanska:
Check for API key added
Fixes: https://github.com/DOAJ/doajPM/issues/2123

### DIFF
--- a/portality/app.py
+++ b/portality/app.py
@@ -311,13 +311,14 @@ def standard_authentication():
         user = models.Account.pull(remote_user)
         if user:
             login_user(user, remember=False)
-    # add a check for provision of api key
     elif 'api_key' in request.values:
-        res = models.Account.query(q='api_key:"' + request.values['api_key'] + '"')['hits']['hits']
-        if len(res) == 1:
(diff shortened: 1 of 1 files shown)


portality/app.py
```diff
@@ -311,13 +311,14 @@ def standard_authentication():
...
             login_user(user, remember=False)
-    # add a check for provision of api key
     elif 'api_key' in request.values:
-        res = models.Account.query(q='api_key:"' + request.values['api_key'] + '"')['hits']['hits']
-        if len(res) == 1:
-            user = models.Account.pull(res[0]['_source']['id'])
-            if user:
-                login_user(user, remember=False)
+        q = models.Account.query(q='api_key:"' + request.values['api_key'] + '"')
+        if q.has_key('hits'):
```
richard-jones: This is fine it doesn't need to be changed, though just to share an idiom that I tend to use in this situation.  I use:

```
q.get("hits", {}).get("hits", [])
```

This always gives you an array, doesn't fail if anything is missing, and then you can check the length of the array for results.

PR #1692, opened 2019-10-08
Add logger message to read an error details on doa pull

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2126


amdomanska:
I debugged it just to be sure and added `out.text`, `out.status_code` and also `out.reason` to the log.

PR #1693, opened 2019-10-08
Feature/python3 harvester

richard-jones:


portality/harvester/epmc/models.py
```diff
@@ -0,0 +1,454 @@
+from builtins import object
```
Steven-Eardley: more futurize

portality/harvester/epmc/queries.py
```diff
@@ -0,0 +1,51 @@
+from builtins import object
```
Steven-Eardley: more unnecessary `futuriz`ation

portality/lib/dataobj.py
```diff
@@ -1,5 +1,8 @@
 # -*- coding: UTF-8 -*-
 
+from builtins import str
```
Steven-Eardley: I really don't want python 2 noise in here.

portality/lib/dataobj.py
```diff
@@ -959,7 +970,7 @@ def construct_validate(struct, context=""):
     }
     """
     # check that only the allowed keys are present
-    keys = struct.keys()
+    keys = list(struct.keys())
```
Steven-Eardley: probably not necessary to cat to list() since we're iterating through them next, but hopefully these will be caught as conflicts on the merge. It'll still work like this, too.

portality/lib/http.py
```diff
@@ -0,0 +1,242 @@
+from future import standard_library
```
Steven-Eardley: more futurize stuff

portality/models/harvester.py
```diff
@@ -0,0 +1,243 @@
+from builtins import object
```
Steven-Eardley: more unnecessary `futurize`

portality/models/harvester.py
```diff
@@ -0,0 +1,243 @@
...
+        :return:
+        """
+        raise NotImplementedError()
+
+
+class HarvestState(dataobj.DataObj, DomainObject):
+    __type__ = 'harvester_state'
+
+    def __init__(self, **raw):
+        struct = {
```
Steven-Eardley: so far by convention we've defined our structs as constants rather than in __init__ - I guess it doesn't matter really, apart from that it's easier to find if it's not simply called `struct`

portality/models/harvester.py
```diff
@@ -0,0 +1,243 @@
...
+            cls.articles_saved_successfully[plugin] = 1
+
+    @classmethod
+    def record_error(cls, msg):
+        cls.error_messages.append(msg)
+
+    @classmethod
+    def write_report(cls):
+        report = ["Harvester ran from {d1} to {d2}.".format(d1=cls.harvester_started, d2=dates.now())]
+        for p_name in list(cls.last_harvest_dates_at_start_of_harvester.keys()):
```
Steven-Eardley: here we have a couple of unnecessary casts to list

portality/settings.py
```diff
@@ -266,6 +267,16 @@
             }
         }
     },
+    "str": {
```
Steven-Eardley: does this mean we now have both `str` and `unicode` in our mappings? That strikes me as a bad time.

portality/settings.py
```diff
@@ -889,3 +900,52 @@
...
 CONSENT_COOKIE_KEY = "doaj-cookie-consent"
 
+
+
+#############################################
+## Harvester Configuration
+
+## Configuration options for the DOAJ API Client
+
+DOAJ_SEARCH_BASE = "http://doaj.org"
```
Steven-Eardley: should be https

portality/decorators.py
```diff
@@ -99,3 +100,43 @@ def decorated_view(*args, **kwargs):
...
+class CaughtTermException(Exception):
+    pass
+
+
+def _term_handler(signum, frame):
+    app.logger.warning("Harvester terminated with signal " + str(signum))
+    raise CaughtTermException
+
+
+def capture_sigterm(fn):
```
Steven-Eardley: this could be generalised further in future, right now it's harvester-specific.

Steven-Eardley:
After a merge with the python3 upgade branch, tests are failing on this branch due to `rb` mode in `store.py` - @richard-jones it might be worth checking over the changes to the tets which made that incompatible, since `rb` may have been correct but it was changed to `r` to make the tests pass.

portality/api/v1/client/client.py
```diff
@@ -0,0 +1,224 @@
...
+################################################################
+# Old DOAJ client for use against the public search endpoint
+#
+# This is deprecated, and you should use the DOAJCv1API above for full
+# access to all Search and CRUD activities
+
+class DOAJSearchClient(object):
+
+    def __init__(self, search_base=None, query_endpoint=None, search_type=None, search_port=None):
+        self.search_base = search_base if search_base else app.config.get("DOAJ_BASE_URL", "http://doaj.org")
```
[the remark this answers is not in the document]
richard-jones: Done

portality/api/v1/client/client.py
```diff
@@ -0,0 +1,224 @@
...
+        qs = ""
+        if params is not None:
+            for k, v in params.items():
+                if qs != "":
+                    qs += "&"
+                qs += http.quote(k) + "=" + http.quote(str(v))
+        if qs != "":
+            url += "?" + qs
+
+        return url
```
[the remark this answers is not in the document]
richard-jones: wontfix - not enhancing the harvester just porting

portality/api/v1/client/client.py
```diff
@@ -0,0 +1,224 @@
...
+            raise DOAJException("Bad request against DOAJ API: '{x}'".format(x=resp.json().get("error", "no error provided")))
+        elif resp.status_code == 403:
+            raise DOAJException("Forbidden action - your API key was incorrect, or you are trying to add an article with an ISSN you don't own")
+        elif resp.status_code == 401:
+            raise DOAJException("Authentication failed, your API key was probably wrong")
+
+        j = resp.json()
+        return j.get("id"), j.get("location")
+
+class ANDQueryBuilder(object):
```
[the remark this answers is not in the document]
richard-jones: wontfix

portality/api/v1/client/models.py
```diff
@@ -0,0 +1,200 @@
+from portality.lib import dataobj
+
+BASE_ARTICLE_STRUCT = {
```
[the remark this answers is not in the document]
richard-jones: No, actually they should be identical to the article data objects, so I'm going to tie them together

portality/models/harvester.py
richard-jones: seems sensible to be consistent in the main codebase

portality/settings.py
richard-jones: yes, I added this because a "str" coerce type was added to the data objects.  I think we decided to back this out, so this can go.  I'll leave it in for the moment, until we've tracked that other change through.

portality/decorators.py
richard-jones: no fix required right now

portality/harvester/workflow.py
```diff
@@ -99,14 +98,11 @@ def process_issn(cls, account_id, issn):
...
+            article.is_api_valid()
+        except DataObjException as e:
             app.logger.info("Article for Account:{y} was not API valid ... skipping".format(y=account_id))
-
-            try:
-                doajclient.models.ArticleValidator(article.data)
-            except DataObjException as e:
-                Report.record_error(article.get_identifier("doi") + u" - " + e.message)
-
+            Report.record_error(article.get_identifier("doi") + " - " + e.message)
```
Steven-Eardley: python 3 doesn't have `Exception.message`, you want `str(e)`

PR #1694, opened 2019-10-14
abort(404) when retrieving invalid journal

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2129

### DIFF
--- a/portality/view/doaj.py
+++ b/portality/view/doaj.py
@@ -253,21 +253,6 @@ def toc(identifier=None, volume=None, issue=None):
     # identifier may be the journal id or an issn
     journal = None
     issn_ref = False
-    # if len(identifier) == 9:
-    #     js = models.Journal.find_by_issn(identifier, in_doaj=True)
-    #     if len(js) > 1:
-    #         abort(400)                             # really this is a 500 - we have more than one journal with this issn
-    #     if len(js) == 0:
-    #         abort(404)
-    #     journal = js[0]
-    #
-    #     issn_ref = True     # just a flag so we can check if we were requested via issn
-    # else:
-    #     try:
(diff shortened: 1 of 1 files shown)


amdomanska:
I added the check if the identifier is an id (len = 32) if not ISSN. I suggest postponing adding new regex checks for now as that could require further and more in-depth research.

portality/view/doaj.py
```diff
@@ -253,20 +253,27 @@ def toc(identifier=None, volume=None, issue=None):
     # identifier may be the journal id or an issn
     journal = None
     issn_ref = False
+
     if len(identifier) == 9:
```
Steven-Eardley: We have the opportunity here to fix another bug - `None` (the default value for `identifier`) will cause an exception if you give it to `len()`. This means there's an error when you go to https://doaj.org/toc/

portality/view/doaj.py
```diff
@@ -253,20 +253,27 @@ def toc(identifier=None, volume=None, issue=None):
...
-    else:
-        journal = models.Journal.pull(identifier)  # Returns None on fail
+        issn_ref = True  # just a flag so we can check if we were requested via issn
+    elif len(identifier) == 32:
+        js = models.Journal.pull(identifier)  # Returns None on fail
 
-    if journal is None:
-        abort(404)
+        if js is None:
+            abort(400)
```
amdomanska: it should be `abort(404)`

portality/view/doaj.py
```diff
@@ -253,20 +253,27 @@ def toc(identifier=None, volume=None, issue=None):
...
+        journal = js
+    else:
+        abort(400)
```
amdomanska: as above

PR #1695, opened 2019-10-15
Change keywords to lowercase before submitting to DOAJ

amdomanska:
#Fixes: https://github.com/DOAJ/doajPM/issues/2056

The information in the form is added, keywords are changed to lowercase before saving into bibjson struct.

### DIFF
--- a/portality/formcontext/forms.py
+++ b/portality/formcontext/forms.py
@@ -156,7 +156,7 @@ class JournalInformation(Form):
     )
     keywords = TagListField('Add keyword(s) that best describe the journal (comma delimited)',
         [validators.DataRequired(), MaxLen(6, message='You can only enter up to {max_len} keywords.')],
-        description='Maximum 6. Keywords must be in English. All keywords are stored in lowercases.'
+        description='Maximum 6. Keywords must be in English. All keywords are stored in lowercase.'
     )
     languages = DOAJSelectMultipleField('Select the language(s) that the Full Text of the articles is published in',
         [validators.DataRequired()],
--- a/portality/formcontext/forms.py
+++ b/portality/formcontext/forms.py
@@ -156,7 +156,7 @@ class JournalInformation(Form):
     )
     keywords = TagListField('Add keyword(s) that best describe the journal (comma delimited)',
         [validators.DataRequired(), MaxLen(6, message='You can only enter up to {max_len} keywords.')],
-        description='Maximum 6. Keywords must be in English.'
+        description='Maximum 6. Keywords must be in English and in lower case.'
     )
     languages = DOAJSelectMultipleField('Select the language(s) that the Full Text of the articles is published in',
         [validators.DataRequired()],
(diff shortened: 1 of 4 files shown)


portality/formcontext/forms.py
```diff
@@ -156,7 +156,7 @@ class JournalInformation(Form):
     )
     keywords = TagListField('Add keyword(s) that best describe the journal (comma delimited)',
         [validators.DataRequired(), MaxLen(6, message='You can only enter up to {max_len} keywords.')],
-        description='Maximum 6. Keywords must be in English.'
+        description='Maximum 6. Keywords must be in English. All keywords are saved in lowercases.'
```
richard-jones: Can we change the second sentence to `All keywords are stored in lowercase`

portality/models/bibjson.py
```diff
@@ -101,7 +101,10 @@ def add_keyword(self, keyword):
         self._add_to_list_with_struct("keywords", keyword)
 
     def set_keywords(self, keywords):
-        self._set_with_struct("keywords", keywords)
+        lowercase_keywords = []
+        for k in keywords:
+            lowercase_keywords.append(k.lower())
+        self._set_with_struct("keywords", lowercase_keywords)
```
richard-jones: Can we modify one of our existing unit tests to demonstrate this working.  Probably `test_models.py` would be the right one.

portality/formcontext/forms.py
Steven-Eardley: @amdomanska just `lowercases` to `lowercase` still to do

portality/models/bibjson.py
```diff
@@ -98,13 +98,16 @@ def keywords(self):
...
 
     def set_keywords(self, keywords):
-        lowercase_keywords = []
-        for k in keywords:
-            lowercase_keywords.append(k.lower())
-        self._set_with_struct("keywords", lowercase_keywords)
+        if type(keywords) is list:
+            lowercase_keywords = []
+            for k in keywords:
+                lowercase_keywords.append(k.lower())
```
Steven-Eardley: This is fine, but you could save lines and memory by using:
`keywords = map(str.lower, keywords)`
**or**
`keywords = [w.lower() for l in keywords]`

Both apply the transformation in-place. In this case it's very minor since we restrict keyword length!

portality/models/bibjson.py
```diff
@@ -98,13 +98,16 @@ def keywords(self):
...
+            self._set_with_struct("keywords", lowercase_keywords)
+        else:
+            self._set_with_struct("keywords", keywords.lower())
```
Steven-Eardley: `_set_with_struct()` will handle wrapping it in a list for you, so this is okay. (I had to look that up to check)

portality/models/bibjson.py
amdomanska: Cool! I'm gonna change it to remember next time ;)

portality/formcontext/forms.py
amdomanska: Sorry :/

portality/models/bibjson.py
Steven-Eardley: @amdomanska good choice with the list comprehension - @richard-jones gets uncomfortable when I use functional programming like `map` and `fold` :stuck_out_tongue:

PR #1697, opened 2019-10-16
prevent removing last author div from metadata form

amdomanska:
On the upload metadata form for a publisher it is possible to remove all the authors' fields. Unfortunately then the button "Add More Authors" doesn't work. Because at least one Author is required I think it is a good idea to just prevent the user from removing all the fields and give him proper error if she/he tries to do so - exactly the same one as she/he would like to submit the form without the authors.
Screenshot from 2019-10-16 13-54-49

### DIFF
--- a/portality/templates/publisher/metadata.html
+++ b/portality/templates/publisher/metadata.html
@@ -127,9 +127,9 @@ $(document).ready(function() {
     function showHideFirstRemoveButton() {
         // Hide delete button when there's just one author
         if ($('[id^=authors-][id$="container"]').length === 1) {
-            $('#remove_authors-0').hide();
+            $('#remove_authors-0').css('visibility', 'hidden')
         } else {
-            $('#remove_authors-0').show();
+            $('#remove_authors-0').css('visibility', 'visible')
         }
     }
     
@@ -183,7 +183,7 @@ $(document).ready(function() {
         rem_b.unbind("click");
         rem_b.click(removeAuthor);
         if (all_e.length === 1){
-            $('#remove_authors-1').show()
+            $('#remove_authors-1').css('visibility', 'visible')
         }
 
         showHideFirstRemoveButton()
--- a/portality/templates/publisher/metadata.html
+++ b/portality/templates/publisher/metadata.html
@@ -123,11 +123,18 @@ $(document).ready(function() {
             ce.attr('name', id);
         })
     }
+
+    function showHideFirstRemoveButton() {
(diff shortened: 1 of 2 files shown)


Steven-Eardley:
Looks like a sensible change - is there a related issue to this PR?

amdomanska:
@Steven-Eardley 
https://github.com/DOAJ/doajPM/issues/2175

portality/templates/publisher/metadata.html
```diff
@@ -127,13 +127,22 @@ <h4 class="red">There is a problem with the submitted form.</h4>
     
     function removeAuthor() {
         event.preventDefault();
-        
+
+        var all_e = $('[id^=authors-][id$="container"]');
+        if (all_e.length === 1) {
+            console.log($('#remove_warning') == 0)
+            if ($('#remove_warning').length == 0) {
+                all_e.last().append('<div id="remove_warning"><ul class="red"><li>Please provide at least one author</li></ul><div>')
```
Steven-Eardley: Doesn't the validator already do this? If so, would it be better to only display the delete button if there's more than one author field?

amdomanska: It does after clicking the submit button. But yes, that's actually a much better idea. Instead of displaying the warning it is better to just hide the X button when there is only one author. I'll change it and re-request the review when ready.

amdomanska:
New version:

Screenshot from 2019-10-18 14-28-27
Screenshot from 2019-10-18 14-28-32
Screenshot from 2019-10-18 14-28-35

amdomanska:
@richard-jones I was thinking about that but tbh I just assumed it's on purpose.

Steven-Eardley:
Before I release this I'm going to cleanup the code (lots of missing semicolons) and make it so that when you have 2 authors you can delete either of them, but not the last.

Steven-Eardley:
New behaviour:

Screenshot from 2019-10-24 13-05-52
Screenshot from 2019-10-24 13-06-32
Screenshot from 2019-10-24 13-06-53

Steven-Eardley:
Admittedly if I'd written stuff in those fields it'd make more sense, but trust me it's logical.

amdomanska:
I do agree with you. That definitely makes sense.

Steven-Eardley:
Finally I've changed hide / show for visible / hidden so the layout doesn't change when there's no delete box. Are you familiar with the term 'yak shaving'? :P I'm done now!

PR #1699, opened 2019-10-22
Hotfix/20191022 0947 issues 2176 2179

richard-jones:
Relates to:

https://github.com/DOAJ/doajPM/issues/2176
https://github.com/DOAJ/doajPM/issues/2179
https://github.com/DOAJ/doajPM/issues/2170

### DIFF
--- a/portality/static/css/doaj.css
+++ b/portality/static/css/doaj.css
@@ -77,6 +77,14 @@ blockquote {
     background-color: white;
 }
 
+.navbar .nav .active > a:focus,             
+.navbar .nav > li:focus,
+.navbar .nav > li > a:focus,
+.navbar .navbar-header a:focus {
+    color: #F68B1F;
+    background-color: #eeeeee;
+}
+
 .navbar .btn-navbar {
     background-color: #ffffff;
     background-image: none;
--- a/portality/static/css/doaj.css
+++ b/portality/static/css/doaj.css
@@ -68,8 +68,6 @@ blockquote {
 }
 
 .navbar .nav .active > a,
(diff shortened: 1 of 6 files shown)


portality/static/css/doaj.css
```diff
@@ -77,6 +77,14 @@ blockquote {
     background-color: white;
 }
 
+.navbar .nav .active > a:focus,             
+.navbar .nav > li:focus,
+.navbar .nav > li > a:focus,
+.navbar .navbar-header a:focus {
+    color: #F68B1F;
+    background-color: #eeeeee;
```
amdomanska: It probably should be `background-color: white`. This way you will have the same background for :focus and for :hover

amdomanska: Screenshot from 2019-10-23 10-22-18

richard-jones: Yes, so, interesting point.  That was my first consideration, but an active nav is also highlighted white, so is there a usability question around having the tab-based navigation highlight in the same way as the current active nav?  That is, if you tab through the nav, it's not possible to visually determine which is the current nav and which is the focused nav item?

Having said that, the grey looks rubbish, and I notice that when you click a link it briefly becomes grey before becoming white.

Perhaps an additional signifier, like also underlining the focused one, as well as highlighting in white?

Steven-Eardley: I agree, there should be different colours for focus and hover - you want navigating to the next item to be different to what you're currently on. The font colour change makes it readable and that's a good start.

Perhaps we want a lighter orange?

PR #1700, opened 2019-10-23
add test for upgrade.py method

amdomanska:


amdomanska:
It gives us 61% coverage on upgrade.py file

amdomanska:
@Steven-Eardley could you help me with that test failing here, please? It passes locally and the resources/migrate.json file is pushed but the test on Codeship fails with `FileNotFoundError: [Errno 2] No such file or directory: './resources/migrate.json'`. I don't want to merge it before it passes but I don't know why that's happening.

Steven-Eardley:
@amdomanska there were a few issues with that test run so I've reset the cache on codeship and re-run the tests.

doajtest/unit/test_upgrade.py
```diff
@@ -0,0 +1,78 @@
...
+            j.set_in_doaj(True)
+            bj = j.bibjson()
+            bj.add_identifier(bj.P_ISSN, "{x}000-0001".format(x=i))
+            bj.title = "Journal to Change"
+            bj.publisher = "Test Publisher {x}".format(x=i)
+            bj.add_url("http://homepage.com/{x}".format(x=i), "homepage")
+            j.save()
+            saved_journals.append(j.id)
+
+        created = j.last_updated
```
Steven-Eardley: `created` will only have the last_updated value of the very last of the 10 journals. Is that correct for the purposes of this test?

Steven-Eardley:
This test doesn't pass when I run it locally:

```
Failure
  File "/usr/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/lib/python3.7/unittest/case.py", line 628, in run
    testMethod()
  File "/home/steve/code/cl/doaj/doajtest/unit/test_upgrade.py", line 64, in test_upgrade
    assert j.last_updated == created
AssertionError
```

This could be due to the issue with `created` I mentioned above.

amdomanska:
Strange, as it passes when I run it locally at my machine :D I'll have a look at it.

PR #1706, opened 2019-10-30
check if valuelist is not None or empty

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2196

### DIFF
--- a/portality/formcontext/fields.py
+++ b/portality/formcontext/fields.py
@@ -61,17 +61,17 @@ class URLField(TextField):
     def process_formdata(self, valuelist):
         if valuelist is None or len(valuelist) == 0:
              self.data = self.data = u''
-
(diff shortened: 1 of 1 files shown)


portality/formcontext/fields.py
```diff
@@ -59,6 +59,9 @@ class URLField(TextField):
     widget = widgets.TextInput()
 
     def process_formdata(self, valuelist):
+        if valuelist is None or len(valuelist) == 0:
+             self.data = self.data = u''
```
Steven-Eardley: if you don't return early this will still fail on the next line when you try to access `valuelist[0]`

amdomanska: Stupid mistake. I put that in if-else statement.

PR #1707, opened 2019-10-30
Hotfix/2191 password keyerror

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2191
It happens that user is trying to log in that has no field 'password' in his account. As it is not a usual situation I added the logger error and call abort(404).

### DIFF
--- a/portality/app.py
+++ b/portality/app.py
@@ -338,14 +338,17 @@ if 'api' in app.config['FEATURES']:
         )
 
 
-@app.errorhandler(404)
+@app.errorhandler(400)
 def page_not_found(e):
-    return render_template('404.html'), 404
-
+    return render_template('400.html'), 400
 
 @app.errorhandler(401)
 def page_not_found(e):
     return render_template('401.html'), 401
+
+@app.errorhandler(404)
+def page_not_found(e):
+    return render_template('404.html'), 404
 
 @app.errorhandler(500)
 def page_not_found(e):
--- a/portality/app.py
+++ b/portality/app.py
@@ -347,6 +347,10 @@ def page_not_found(e):
 def page_not_found(e):
     return render_template('401.html'), 401
 
+@app.errorhandler(500)
+def page_not_found(e):
+    return render_template('500.html'), 500
+
 
 if __name__ == "__main__":
     pycharm_debug = app.config.get('DEBUG_PYCHARM', False)
--- a/portality/models/account.py
+++ b/portality/models/account.py
@@ -6,8 +6,6 @@ from werkzeug.security import generate_password_hash, check_password_hash
 from portality.dao import DomainObject as DomainObject
 from portality.core import app
 from portality.authorise import Authorise
-from flask import abort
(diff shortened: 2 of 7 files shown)


portality/models/account.py
```diff
@@ -92,7 +94,11 @@ def set_password(self, password):
         self.data['password'] = generate_password_hash(password)
 
     def check_password(self, password):
-        return check_password_hash(self.data['password'], password)
+        try:
+            return check_password_hash(self.data['password'], password)
+        except:
```
Steven-Eardley: this is an undesirable bare exception (it should be `KeyError`), and I think a 401 or 403 might be more appropriate in this case.

amdomanska: True. Good lesson. I haven't thought about it.

amdomanska:
As we agreed with @Steven-Eardley during the phone call it is 500 error that's going to be raised in that case. I implemented a 500 html template (slightly different to 404.html) and its handler.

amdomanska:
Screenshot from 2019-11-01 15-45-52

amdomanska:
I added 400 error (it's copied 404 html template) alongside to fix: https://github.com/DOAJ/doajPM/issues/2168

portality/app.py
```diff
@@ -338,15 +338,22 @@ def api_directory():
         )
 
 
-@app.errorhandler(404)
+@app.errorhandler(400)
```
Steven-Eardley: numerical order. Nice. It's the small things :)

portality/models/account.py
```diff
@@ -6,6 +6,8 @@
 from portality.dao import DomainObject as DomainObject
 from portality.core import app
 from portality.authorise import Authorise
+from flask import abort
```
Steven-Eardley: we don't need this import statement anymore

portality/templates/500.html
```diff
@@ -0,0 +1,10 @@
+{% extends "/base.html" %}
+
+{% block content %}
+
+<h2>Oops. Something went wrong.</h2>
+
+<p>An error has occurred. It looks like something has gone wrong.</p>
+<p>Please contact us using the Contact link at the bottom od the page.</p>
```
Steven-Eardley: Typo, and we could improve these templates by including the actual contact us URL (I'm not sure why we currently direct the user to click the other link, it's probably bad design or something)

portality/view/account.py
```diff
@@ -131,14 +131,17 @@ def login():
...
-            # return form.redirect('index')
-            # return redirect(url_for('doaj.home'))
-            return redirect(get_redirect_target(form=form))
-        else:
-            flash('Incorrect username/password', 'error')
+        try:
+            if user is not None and user.check_password(password):
+                login_user(user, remember=True)
+                flash('Welcome back.', 'success')
+                # return form.redirect('index')
```
Steven-Eardley: probably time for these commented-out lines to go.

PR #1708, opened 2019-10-31
Hotfix/2167 full content description

amdomanska:
This PR interferes with application/journals forms. Could you please check if these forms are displayed correctly in case I missed something?

### DIFF
--- a/portality/formcontext/formcontext.py
+++ b/portality/formcontext/formcontext.py
@@ -16,7 +16,7 @@ ACC_MSG = 'Please note you <span class="red">cannot edit</span> this application
 SCOPE_MSG = 'Please note you <span class="red">cannot edit</span> this application as you don\'t have the necessary ' \
             'account permissions to edit applications which are {0}.'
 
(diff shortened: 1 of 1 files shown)


richard-jones:
Also, couple of other comments:

* There's at least one URL field where this is missed out: Q53.  Possibly others are missed too, as a lot of URL fields are not visible unless other fields have been answered in a specific way

* Make sure the CSS for the description forces a word-break wrap, so that long URLs don't break the layout.

amdomanska:
> I don't think that this is quite going to scratch the itch. It's a good first step, and now can we make the URLs themselves actual clickable links?

As far as I understand from https://github.com/DOAJ/doajPM/issues/2167#issuecomment-545774353 this was meant to be a quick fix to ensure that the URL is entirely visible.

>  I wonder if instead, we can use the same solution as we have currently for Question 6 where the full value is written under the box? This was exactly why we did this for that question now I think of it.

We can make this full-description clickable but maybe it's worth to release it sooner as-is as a first step to improve editors' workflow and think about later improvements after that?

I'll double-check the other questions. CSS I believe should be fine - I tested it with long URLs by myself and it is an existing solution for other fields that we had no issues with so far.

dommitchell:
Clickable URL is almost a slightly different thing. The URL itself still needs to be displayed in full for reading.

amdomanska:
@dommitchell Could you please have a look at Q36 at the screenshot and confirm that is what you had in mind?

Screenshot from 2019-11-01 09-56-37

dommitchell:
Perfect! Thanks @amdomanska

amdomanska:
Clickable version released today at 9am

PR #1709, opened 2019-11-04
make url descriptions clickable

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2206
URL fields' full content descriptions are clickable URL and open in a new tab.

Screenshot from 2019-11-04 10-46-48


amdomanska:
Thanks!

PR #1714, opened 2019-11-07
add script and corresponding readme

amdomanska:
#Closes: https://github.com/DOAJ/doajPM/issues/2056#issuecomment-544894778


amdomanska:
@Steven-Eardley could you please let me know when I could run this migration on live?

amdomanska:
Improved. Thank you guys for your help, I really overcomplicated that. I removed articles from the script, added suggestions and used wrote simple json and operations to use those with upgrade.py

portality/migrate/20191128_2056_keywords_to_lower/operations.py
```diff
@@ -0,0 +1,9 @@
+from portality import models
+from portality.core import app
+import esprit
+import re
```
Steven-Eardley: I don't think any of these imports are used

amdomanska:
Now it's gonna be fine I think.

amdomanska:
@Steven-Eardley I guess it's finally ready to run

PR #1715, opened 2019-11-11
Turn off autosuggestion on new application form

amdomanska:
Fixes https://github.com/DOAJ/doajPM/issues/2211

### DIFF
--- a/portality/formcontext/formhelper.py
+++ b/portality/formcontext/formhelper.py
@@ -143,6 +143,7 @@ class FormHelperBS3(object):
             for k, v in kwargs.iteritems():
                 if k in ["class", "style", "disabled"] or k.startswith("data-"):
                     render_args[k] = v
+                render_args["autocomplete"] = "new-password"    # always turn off autocompletion at the field level
             frag += field(**render_args) # FIXME: this is probably going to do some weird stuff
 
             # FIXME: field.value isn't always set
--- a/portality/templates/formcontext/assed_application_review.html
+++ b/portality/templates/formcontext/assed_application_review.html
@@ -12,7 +12,7 @@
 {# include the table for comparing applications with journals that have changed #}
 {% include 'formcontext/_application_diff.html' %}
 
-<form method="post" action="#first_problem" class="form-horizontal application_journal_form" id="suggest_form">
+<form autocomplete="off" method="post" action="#first_problem" class="form-horizontal application_journal_form" id="suggest_form">
     {% if form_context.errors %}
     <h4 class="red form-status">There is a problem with the submitted form.</h4>
     {% endif %}
--- a/portality/templates/formcontext/assed_journal_review.html
+++ b/portality/templates/formcontext/assed_journal_review.html
@@ -9,7 +9,7 @@
 {% set heading_object_type = 'Journal' %}
 {% include 'formcontext/_lockable_header.html' %}
 
-<form method="post" action="#first_problem" class="form-horizontal application_journal_form" id="suggest_form">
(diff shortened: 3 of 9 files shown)


amdomanska:
> I'm curious as to why there's an extra input field; shouldn't disabling for the whole form be enough?

A few older browsers support `autosuggestion=off` only for input field.

amdomanska:
> Which if any of these fields https://www.w3.org/TR/WCAG21/#input-purposes is currently being mistaken?

The problem is mainly with URL for Open Access Statement and Platform I believe.

amdomanska:
> Also, according to https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion, we will also lose the local cache of filled fields, which could be annoying if e.g. a user accidentally clicks back then wants to go forward again to the form.

Good point. I believe @dommitchell needs to decide what's more important.

amdomanska:
@richard-jones I'm not sure if it matters but it think it was meant to be for application form only
https://github.com/DOAJ/doajPM/issues/2211#issuecomment-552799163

dommitchell:
> we will also lose the local cache of filled fields, which could be annoying if e.g. a user accidentally clicks back then wants to go forward again to the form.

I think that this would be *really* annoying actually. Can we hold off on this one until we can discuss it or is that tricky?

amdomanska:
@richard-jones your solution works with cache fine, right?

richard-jones:
>>we will also lose the local cache of filled fields, which could be annoying if e.g. a user accidentally clicks back then wants to go forward again to the form.

> I think that this would be really annoying actually. Can we hold off on this one until we can discuss it or is that tricky?

In my testing I didn't have any issues with the page caching, and I'm not sure how autocomplete as a user feature and "remembering what the user typed into the form" are the same. I suspect this concern is just the product of some dodgy online documentation somewhere.  All my research online about this hasn't turned up any explicit link between those two features.

I'll hold off until you give the go ahead, and if you want we can put it up on test so you can try it out first.

Steven-Eardley:
I've put this up on test as the branch `feature/2019-11-27_application_form_autocomplete` (because this branch is python2)

I'm seeing the issue I expected above - on the live site, you can part-edit the form, hit back in the browser, and then forward and your unsubmitted form is still there. This no longer happens on the test server.

Steven-Eardley:
I've now reinstated `develop` on the test server for the sake of the upcoming demo.

richard-jones:
This issue looks contentious, and also subject to differing whims of the browser manufacturers.  Reading this page https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html it suggested that you could specify autocomplete fields where they are relevant, but the problem we're having is that FFX and Chrome (and others?) are assuming fields should be autocompleted when they shouldn't be.

I also don't get the caching issue in Chrome (I have not tried it in FFX).

We could take a more targetted approach, and turn autocomplete off on fields which we know to be problematic, like `oa_statement`.

richard-jones:
One other useful resource: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill

I think the answer here is that browsers really want this to be down to users to control, but in the process they have built an autocomplete system which is not really very good, at least when it comes into contact with a complex form like ours.  That puts us in an essentially impossible position, and the best we can probably do is experiment with different autocomplete and field names to see if we can avoid erroneous offers of autocomplete.  That could be time consuming, so would be good to talk about it before doing more.

richard-jones:
This PR is no longer relevant, superseded by the new form, so closing

PR #1718, opened 2019-11-19
Feature/python3 old languages xml

Steven-Eardley:


portality/datasets.py
```diff
@@ -1,114 +1,156 @@
...
     try:
         return pycountry.countries.lookup(current_country).alpha_2
     except LookupError:
         return None if fail_if_not_found else current_country
-
+'''
 
 def get_country_name(code):
     """ Get the name of a country from its two-character code """
+    return legacy_datasets.get_currency_name(code)
```
amdomanska: It should be `return legacy_datasets.get_country_name(code)`

PR #1719, opened 2019-11-20
Feature/275 article metadata

amdomanska:
Closes: https://github.com/DOAJ/doajPM/issues/275


amdomanska:
@Steven-Eardley do you think it would be possible to put that up to the test server so we could test it more? There is a risk that if I made an error somewhere it will change existing articles so would be nice to test it throughout.

portality/formcontext/formcontext.py
```diff
@@ -1438,8 +1438,21 @@ def _set_choices(self):
         pass
     """
 
-### Journal form contexts ###
+class ManEdArticleReview(PrivateContext):
```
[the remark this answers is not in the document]
richard-jones: It looks like you started out with this approach and then abandoned it, but I think it has merit, because it's exactly what the formcontexts were designed for.  The parent class `Private Context` is heavily geared towards applications and journals, but you would get away with using it for articles too, I think.

portality/models/account.py
```diff
@@ -49,6 +49,12 @@ def pull_by_email(cls, email):
         else:
             return None
 
+    @classmethod
+    def pull_by_name(cls, name):
```
richard-jones: I'm against this method, and it looks like it doesn't get used anywhere, so can be removed.

portality/models/article.py
```diff
@@ -591,6 +596,34 @@ def save(self, *args, **kwargs):
         self._generate_index()
         return super(Article, self).save(*args, **kwargs)
 
+    def get_owner(self):
```
richard-jones: If we're going to introduce this method, I'd love to DRY up the code a bit and use it in `portality.bll.services.article` in `is_legitimate_owner`.  That function could call this function, and then handle the exceptions to give the return values that it wants.

portality/models/bibjson.py
```diff
@@ -41,6 +41,11 @@ def title(self, val):
     ## work with the identifiers
 
     def add_identifier(self, idtype, value):
+        all = self.get_identifiers()
+        existing = [x for x in all if x["type"] == idtype]
+        if existing:
+            for x in existing:
+                all.remove(x)
```
richard-jones: This is a good idea to check, and actually there's some underlying code which will do this for you.  Instead of doing this manually, just use

`self._delete_from_list("identifier", matchsub=idobj)`

portality/static/js/edges/admin.journalarticle.edge.js
```diff
@@ -52,7 +52,7 @@ $.extend(true, doaj, {
                 // if it's not a suggestion or a journal .. (it's an article!)
                 // we really need to expose _type ...
                 var result = '<a class="delete_article_link" href="';
-                result += "/admin/article/";
+                result += "/admin/delete/article/";
```
richard-jones: I realise this was like this before, but while we're at it can we treat this the same was as we do for `editArticle` - that is, take it from `doaj.adminJournalArticleSearchConfig.articleDeleteUrl` which is in turn provided in the template for this page?

portality/templates/admin/admin_site_search.html
```diff
@@ -77,6 +77,7 @@
     <script type="text/javascript">
     $.extend(true, doaj, {
         adminJournalArticleSearchConfig : {
+            articleEditUrl : "{{ url_for('admin.article_page', article_id='') }}",
```
richard-jones: so, as per above comment, add `articleDeleteUrl` here

portality/templates/admin/edit_article_metadata.html
```diff
@@ -0,0 +1,161 @@
...
+
+    {% if there_were_errors %}
+        <h4 class="red">There is a problem with the submitted form.</h4>
+    {% endif %}
+
+    {% from "_formhelpers.html" import render_field_horizontal %}
+    {% from "_formhelpers.html" import render_field %}
+    <form method="post" action="{{ url_for('admin.article_page', article_id=form.id) }}" class="form-horizontal" id="article_metadata_form">
+
+        {% include "formcontext/maned_article_review.html" %}
```
richard-jones: I'm not sure if this bit will change if we resurrect your initial FormContext approach, but probably we should give this a better name, since we wouldn't want to show the managing editor's form to the publisher (if they were different, which I realise they are not just now).  Probably call this `base_article_form.html` or something.

portality/templates/admin/edit_article_metadata.html
```diff
@@ -0,0 +1,161 @@
...
+
+<div class="row" style="margin-bottom: 20px">
+    <div class="col-md-9">
+        <h1>{{ heading_object_type }}: {{form_context.source.bibjson().title}}</h1>
+        {% if form_context.source.last_updated %}
+          <h4>Last updated: {{form_context.source.last_updated}}</h4>
+        {% endif %}
+        {% if form_context.source.current_journal %}
+            <h4 style="margin-top: 1em;"><a href="{{ url_for('doaj.journal_readonly', journal_id=form_context.source.current_journal) }}" target="_blank">View associated journal</a></h4>
+        {% endif %}
```
richard-jones: Some journal-related bits here which aren't relevant

portality/templates/publisher/metadata.html
```diff
@@ -3,81 +3,21 @@
...
+
+    {% if there_were_errors %}
+        <h4 class="red">There is a problem with the submitted form.</h4>
+    {% endif %}
+
+    {% from "_formhelpers.html" import render_field_horizontal %}
+    {% from "_formhelpers.html" import render_field %}
+    <form method="post" action="{{ url_for('publisher.metadata') }}" class="form-horizontal" id="article_metadata_form">
+
+    {% include "formcontext/maned_article_review.html" %}
```
richard-jones: see comment above about name of this file

portality/templates/admin/edit_article_metadata.html
```diff
@@ -0,0 +1,161 @@
...
+
+    </form>
+        </div>
+            </div>
+
+{% endblock %}
+
+{% block extra_js_bottom %}
+<script type="text/javascript">
+$(document).ready(function() {
```
richard-jones: I didn't look in detail, but I presume that the JS here and the JS in the publisher form is the same, so we should put that all in a single file, and use it in both places rather than having copies.

I'd also encourage us to move it to the new JS style, which involves creating or extending a `doaj` global variable and placing all the features in there.

portality/view/admin.py
```diff
@@ -144,6 +147,83 @@ def article_endpoint(article_id):
     resp.mimetype = "application/json"
     return resp
 
+def _validate_authors(form, require=1):
```
richard-jones: This should probably be a custom method on the FormContext, rather than defined in the base of the view

portality/view/admin.py
```diff
@@ -144,6 +147,83 @@ def article_endpoint(article_id):
...
+@ssl_required
+@write_required()
+def article_page(article_id):
+    if not current_user.has_role("edit_article"):
+        abort(401)
+    ap = models.Article.pull(article_id)
+    if ap is None:
+        abort(404)
+
+    form = ArticleForm(id=article_id)
```
richard-jones: Ok, it looks like we kind of are and kind of aren't using the FormContext here, and I'm not sure what your intent is.  If we are using the FormContext, then this form should be handled within that, rather than explicitly here.  So, you would do something like:

`ArticleFormFactory.get_form_context("admin", source=ap, form_data=request.form)`

Then you would substitute "admin" for "public" if you wanted to serve the public version instead.

richard-jones: Obviously you then need to write the `ArticleFormFactory` to return the appropriate FormContext instance, and plumb in the `ArticleForm` and templates in there.  I can help you with all of that.

test.cfg
```diff
@@ -3,7 +3,7 @@ APP_MACHINES_INTERNAL_IPS = ["localhost:5050"]
 
 # The app is served via nginx / cloudlflare - they handle SSL
 SSL = False
-DEBUG = True
+DEBUG = False
```
richard-jones: Probably don't include this change in this PR

portality/view/admin.py
```diff
@@ -144,6 +147,83 @@ def article_endpoint(article_id):
...
+            # if the form validates, then we have to do our own bit of validation,
+            # which is to check that there is at least one author supplied
+            if not enough_authors:
+                return render_template("admin/edit_article_metadata.html", form=form, author_error=True, form_context=form_context)
+            else:
+                xwalk = ArticleFormXWalk()
+                art = xwalk.crosswalk_form(form, id=article_id)
+                articleService = DOAJ.articleService()
+                try:
+                    articleService.create_article(art, current_user._get_current_object(), add_journal_info=True)
```
richard-jones: If we're going to use a FormContext, all this stuff should be in there rather than in the view

portality/view/forms.py
```diff
@@ -56,13 +56,25 @@ class ArticleForm(Form):
 
     def __init__(self, *args, **kwargs):
         super(ArticleForm, self).__init__(*args, **kwargs)
+        if "id" in kwargs:
+            a = models.Article.pull(kwargs["id"])
+            self.id = kwargs["id"]
+            if ("method" not in kwargs or kwargs["method"] != "post"):
+                bibjson = a.bibjson()
+                article_form.ArticleFormXWalk.obj2form(self, bibjson)
```
richard-jones: I'm not quite sure what's going on here, but I'm against embedding the form crosswalk into the WTForms instance, it seems the wrong place.  I think we should coalesce everything into a FormContext, in which case the crosswalks will sit comfortably there.

portality/view/publisher.py
```diff
@@ -5,7 +5,7 @@
 from portality.core import app
 from portality import models
 from portality.bll import DOAJ
-from portality.bll.exceptions import AuthoriseException, ArticleMergeConflict
+from portality.bll.exceptions import AuthoriseException, ArticleMergeConflict, ArticleExists
```
richard-jones: unnecessary import

portality/crosswalks/article_form.py
```diff
@@ -1,41 +1,53 @@
...
+
 
 class ArticleFormXWalk(object):
     format_name = "form"
 
-    def crosswalk_form(self, form, add_journal_info=True, limit_to_owner=None):
-        article = models.Article()
-        bibjson = article.bibjson()
+    def crosswalk_form(self, form, add_journal_info=True, limit_to_owner=None, id=None):
+        if id is not None:
```
richard-jones: I have suddenly realised what you were concerned about with regard to calling `prep` on an article, and it's because you're doing the crosswalk onto an existing article.

I think that's a high risk way to approach this, instead I would:

1. Take the data from the article form
2. Crosswalk it into a brand new article object with none of the old properties
3. Pass it to the BLL's `create_article` method, with an instruction to detect and merge duplicates

That way this will behave in exactly the same way as a publisher updating an article with a new XML upload behaves.

portality/api/v1/crud/articles.py
```diff
@@ -5,7 +5,7 @@
 from portality import models
 # from portality.article import XWalk
 from portality.bll.doaj import DOAJ
-from portality.bll.exceptions import ArticleMergeConflict, ArticleNotAcceptable
+from portality.bll.exceptions import ArticleMergeConflict, ArticleNotAcceptable, ArticleExists
```
richard-jones: unnecessary import

amdomanska:
@richard-jones it has been reimplemented using formcontext as we discussed. The code needs a bit of cleaning but it is quite ready to review.

Steven-Eardley:
This branch was merged into develop in error; the PR should probably not be closed yet.

PR #1720, opened 2019-11-20
Fix OAI-PMH access of legacy datasets

Steven-Eardley:
fix for https://github.com/DOAJ/doajPM/issues/2229

thanks to @amdomanska for reporting the bug


richard-jones:
This is quite an urgent fix and there's no one around to review it, so I'm going to approve it and deploy it immediately.

PR #1722, opened 2019-11-21
Feature/python3 crossref

amdomanska:
Closes https://github.com/DOAJ/doajPM/issues/1889
Continuation of: https://github.com/DOAJ/doaj/pull/1664


richard-jones:
We should provide some files to support the functional testing, Crossref equivalents of those in `doajtest/xml_upload_test_package` then we can add a note to the functional tests to say that they should be run with both doaj XML and Crossref XML files.

amdomanska:
I've added files and tested the branch manually on them:

image
image
image
image
image
image
image

doajtest/mocks/model_Article.py
```diff
@@ -31,4 +31,9 @@ def mock(self):
                 if j["pissn"] in pissns and j["eissn"] in eissns:
                     return j["instance"]
 
-        return mock
\ No newline at end of file
+        return mock
+
+    @classmethod
+    def batch_create(cls, *args, **kwargs):
+        raise RuntimeError("oops")
```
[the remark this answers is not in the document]
richard-jones: So I had a closer look and I think in order to achieve what we're trying to do with code organisation, this method should go in `doajtest.mocks.bll_article`

doajtest/mocks/response.py
```diff
@@ -0,0 +1,57 @@
+from doajtest.fixtures.article_crossref import CrossrefArticleFixtureFactory
```
richard-jones: I have not played with it, but @Steven-Eardley might be able to help - there's already a requests mocking library that we have installed as a dependency, so it may be worth replacing this class with that, or at least wrapping it.

portality/crosswalks/article_crossref_xml.py
```diff
@@ -0,0 +1,269 @@
+import os
```
richard-jones: Can we add an explicit crosswalk test in `test_xwalk.py`.

We also need to run the results of the xwalk by the DOAJ team.

amdomanska:
>     * Can you review the Functional Test Scripts and see which ones need to be augmented, and then list them in the issue so the DOAJ team can re-run them.  I guess this is specifically the article xml upload ones, and the instructions to run them using both the DOAJ and the Crossref xml examples.

This is done now.
https://docs.google.com/spreadsheets/d/1oOSC591NGA-05tXqTpwlYSelyYTORtorxU79sqN_xwE/edit#gid=887855589

amdomanska:
@dommitchell
I've just realized that there is quite a significant inconsistency between Crossref XSD and our current recommendations. Crossref XSD requires the title of the journal where the article can be found (which we don't) and at the same time, it doesn't require ISSN (that we do and we use it as a check at the later stage). That raises a few questions:
1) Should we require ISSN also for Crossref articles, even though the Crossref schema itself doesn't, to keep our requirements consistent?
2) If not: What metadata should we use for the check if the article should be added and to which journal attached? If the journal's title - what if that title will be later changed in update request?
3) Should we check if the provided title is consistent with the provided ISSNs?

Personally I don't think it will be possible to avoid the requirement of ISSN for Crossref articles. Without it, we would have to basically build the whole new system of authorizations, deduplication, etc just for articles provided in Crossref schema and I really wouldn't recommend it. I would suggest that how we treat articles in the system shouldn't depend on the schema it was provided at - even if that would require adding the requirement on top of Crossref schema.

doajtest/mocks/response.py
amdomanska: @Steven-Eardley could you please point me in the right direction?

portality/crosswalks/article_crossref_xml.py
amdomanska: There is no explicit crosswalk for doaj schema in `test_xwalk.py` as well. Do you think it would be worth adding alongside Crossref one?

portality/templates/publisher/uploadmetadata.html
```diff
@@ -8,6 +8,8 @@
             <p>Uploaded files must conform to one of the approved XML formats:</p>
             <ul>
                 <li><a href="/static/doaj/doajArticles.xsd">DOAJ Native XML</a></li>
+                <li><a href="/static/doaj/crossref4.4.2.xsd">Crossref 4.4.2 XML</a></li>
```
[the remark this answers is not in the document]
amdomanska: > I'm not sure what would happen if someone uploaded one of the examples from, say, here: https://www.crossref.org/xml-samples/ which conform to 4.3.7 rather than 4.4.2.

I've used one of these and it works ;)

> they may find it works with other versions, but we can't promise

The question is how closely we want to keep an eye on the changes in Crossref XSD schema. I would say that we should say exactly which version we support ("Crossref 4.4.2") and we cannot promise the support for the other versions. Crossref supports the whole range of documents (like, for example, books) and a lot of changes may actually do not touch articles at all and be compatible but unless we want to promise that we will check every single version of XSD I wouldn't promise it.

I haven't found the information on how breaking changes to Crossref schema are announced - maybe it would be enough to keep an eye on this.

dommitchell:
Hi @amdomanska yes, I agree with your suggestions. Thank you!

@ClaraDOAJ we should communicate this to Crossref.

portality/tasks/ingestarticles.py
```diff
@@ -220,7 +220,10 @@ def _download(self, file_upload):
         job.add_audit_message("Downloaded {x} as {y}".format(x=file_upload.filename, y=file_upload.local_filename))
 
         xwalk_name = app.config.get("ARTICLE_CROSSWALKS", {}).get(file_upload.schema)
-        xwalk = plugin.load_class(xwalk_name)()
+        try:
+            xwalk = plugin.load_class(xwalk_name)()
+        except IngestException:
+            raise RetryException(u"Unable to load schema {}".format(xwalk_name))
```
Steven-Eardley: we probably shouldn't have u'' in front of our strings nowadays

portality/app.py
```diff
@@ -170,6 +170,12 @@ def legacy_doaj_XML_schema():
             mimetype="application/xml", as_attachment=True, attachment_filename=schema_fn
             )
 
+@app.route("/isCrossrefLoaded")
+def is_crossref_loading():
```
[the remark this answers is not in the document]
amdomanska: Actually, it doesn't make sense that the route is `isCrossrefLoadED` and method `is_crossref_loadING`. I'm changing it to make it consistent.

doajtest/mocks/file.py
```diff
@@ -0,0 +1,12 @@
+class FileMockFactory(object):
+
+    def __init__(self, filename="filename.xml", stream=None):
```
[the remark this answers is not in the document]
amdomanska: It doesn't. But filename is required by the mocked method and not all tests pass filename argument. So theoretically we don't need it but practically we do.

doajtest/mocks/response.py
Steven-Eardley: I never got around to this before release, I can catch up on this later

PR #1723, opened 2019-11-21
Feature/2212 pyhon3 recaptcha

amdomanska:
Closes: DOAJ/doajPM#2212
Continuation: https://github.com/DOAJ/doaj/pull/1713


portality/app.py
```diff
@@ -342,6 +344,15 @@ def page_not_found(e):
...
 
+@app.route('/gettoken/<g_recaptcha_response>')
+def verify_recaptcha(g_recaptcha_response):
+    with urllib.request.urlopen('https://www.google.com/recaptcha/api/siteverify?secret=' + app.config.get("RECAPTCHA_SECRET_KEY", "") + '&response=' + g_recaptcha_response) as url:
+        data = json.loads(url.read().decode())
+        return data
+
+@app.route('/get_site_key')
+def get_site_key():
+    return app.config.get('RECAPTCHA_SITE_KEY')
```
richard-jones: Is there a better place we can define these routes?

My feeling is we could either introduce a `portality.view.recaptcha` view, giving us a clean place to put them, or maybe they would go in `portality.view.doajservices`. It feels weird having them in the main `app.py`.  There is a bunch of other stuff in app.py, but it's all error handlers, authentication or legacy routes, and I think it would be best to keep it that way.

portality/static/js/contact.js
```diff
@@ -1,18 +1,54 @@
...
+        window.console.log("now");
+        grecaptcha.render('html_element', {
+                'sitekey' : key,
+                'callback' : captchaCallback,
+            });
+        }
+    );
+
+
+};
```
richard-jones: Can we, since we're touching this file, bring it into line with how we want to do JS going forward?

So, introduce `doaj.contact` as a module by declaring this file like:

```
var doaj = doaj || {}
doaj.contact = {
   calculateRemaining : function() {},
   onLoadCallback : function() {}
}
```

portality/templates/base.html
```diff
@@ -40,7 +40,7 @@
 
     <!-- get the app's css -->
     <link rel="stylesheet" type="text/css" href="/static/css/doaj.css?v={{config.get('DOAJ_VERSION')}}">
-
+    <script type="text/javascript" src="/static/js/contact.js?v={{config.get('DOAJ_VERSION')}}"></script>
```
richard-jones: This should go in the `contact.html` template, in a section that looks like this:

```
{% block extra_js_bottom %}
<script type="text/javascript" src="/static/js/contact.js?v={{config.get('DOAJ_VERSION')}}"></script>
{% endblock extra_js_bottom %}
```

Just to clarify why:
* loading JS should be done as late in the page load as possible, so that it doesn't interrupt the page render.  That's why we load all our JS at the end, and much of it is in a template called `_js_includes.html`.
* We don't need this feature except on a contact page, so there's no point loading it on all the other pages, which is what would happen if you put it in the base template.

portality/templates/doaj/contact.html
```diff
@@ -45,8 +45,16 @@
     {{ render_field_horizontal(form.message, style="width: 100%; max-width: 400px; min-height: 200px", placeholder="Enter your message here") }}
 
     <div class="form-group">
-        <div class="col-md-9 col-md-offset-3">
-            <button class="btn btn-success" id="send">Send</button>
+        <div class="submit-with-recaptcha col-md-9 col-md-offset-2">
+            <form method="POST">
+              <div id="html_element"></div>
+            </form>
+            <script src="https://www.google.com/recaptcha/api.js?onload=onloadCallback&render=explicit"
```
richard-jones: If we modify the contact.js file in the way I suggested, then we also need to just change `onload=doaj.contact.onLoadCallback`

portality/templates/base.html
richard-jones: In fact, defined as it is right now, it doesn't work at all because jquery isn't loaded until later :)

richard-jones: Actually, I just looked, and it's already defined in contact.html, so you can just omit this.

portality/static/js/contact.js
```diff
@@ -1,18 +1,54 @@
...
+
+
+});
+
+var onloadCallback = function() {
+
+    $("#submitBtn").prop("disabled", true);
+
+    var captchaCallback = function(param) {
+        $.get( "gettoken/" + param, function( data ) {
```
richard-jones: Probably need a leading "/" here just to be safe on the URL path.

portality/app.py
richard-jones: Also, specifically on RECAPTCHA_SITE_KEY, could we add a placeholder in settings.py for that?

When I went to have a look, I noticed we also have RECAPTCHA_PUBLIC_KEY in settings.py, and it looks like that's not used anywhere so we should probably clear it out.

portality/templates/doaj/contact.html
```diff
@@ -45,8 +45,16 @@
     {{ render_field_horizontal(form.message, style="width: 100%; max-width: 400px; min-height: 200px", placeholder="Enter your message here") }}
 
     <div class="form-group">
-        <div class="col-md-9 col-md-offset-3">
-            <button class="btn btn-success" id="send">Send</button>
+        <div class="submit-with-recaptcha col-md-9 col-md-offset-2">
+            <form method="POST">
```
richard-jones: This is nesting a form inside a form, which isn't syntactically valid HTML.  Chrome just converts this to a div, so I'd probably do the same here.

portality/templates/doaj/contact.html
```diff
@@ -45,8 +45,16 @@
     {{ render_field_horizontal(form.message, style="width: 100%; max-width: 400px; min-height: 200px", placeholder="Enter your message here") }}
 
     <div class="form-group">
-        <div class="col-md-9 col-md-offset-3">
-            <button class="btn btn-success" id="send">Send</button>
+        <div class="submit-with-recaptcha col-md-9 col-md-offset-2">
```
richard-jones: The layout here doesn't quite line up with the rest of the form.  Looks like it needs an offset of about 3, possibly with some other padding.

portality/view/doaj.py
```diff
@@ -368,11 +368,26 @@ def contact():
...
+
+def _verify_recaptcha(g_recaptcha_response):
+    print(g_recaptcha_response)
+    with urllib.request.urlopen('https://www.google.com/recaptcha/api/siteverify?secret=' + app.config.get("RECAPTCHA_SECRET_KEY") + '&response=' + g_recaptcha_response) as url:
+        data = json.loads(url.read().decode())
+        return data
 
+@app.route('/get_site_key')
+def get_site_key():
+    return app.config.get('RECAPTCHA_SITE_KEY')
```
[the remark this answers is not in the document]
richard-jones: Error should be user friendly, and not tell them specifically what's wrong.  Something like "unable to load recaptcha" should be sufficient.

amdomanska:
I don't have a real working set of keys for Recaptcha, so we would need to test again with them.

@richard-jones to test it you would have to register localhost domain on admin site here: https://www.google.com/recaptcha/intro/v3.html
You will receive the secret key and the site key that you then would have to define as:
```
RECAPTCHA_SITE_KEY
RECAPTCHA_SECRET_KEY
```
in `dev.cfg` file. The same needs to be done with the credentials provided by Dom: https://github.com/DOAJ/doajPM/issues/2212#issue-517042886 before going live.

portality/settings.py
```diff
@@ -48,7 +48,7 @@
 ENABLE_EMAIL = True
 ENABLE_PUBLISHER_EMAIL = True
 MANAGING_EDITOR_EMAIL = "managing-editors@doaj.org"
-CONTACT_FORM_ADDRESS = "feedback+contactform@doaj.org"
+CONTACT_FORM_ADDRESS = "aga@cottagelabs.com"
 
 # Error logging via email
 SUPPRESS_ERROR_EMAILS = False
```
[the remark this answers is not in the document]
amdomanska: The public key can be added here but I believe the secret key should be added to the file not tracked by git? For now, I've added them both to `dev.cfg` file just as aws access keys?

portality/view/doaj.py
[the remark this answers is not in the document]
amdomanska: The question is what do we want to do if Recaptcha fails to load. Maybe we shouldn't block the possibility to send the contact form then? What about unblocking sending button in such a case and adding the warning to the mail message for DOAJ/us to catch it quickly? Not sure though how difficult that would be to implement.

What I'm worried about is that Recaptcha will stop loading properly for some reason in the future and noone will be able to contact us through the form till we will notice.

portality/app.py
amdomanska: > My feeling is we could either introduce a `portality.view.recaptcha` view, giving us a clean place to put them, or maybe they would go in `portality.view.doajservices`. It feels weird having them in the main `app.py`

I've added them now in doaj.py right under contact endpoint implementation. It's only now one method:
```
@app.route('/get_site_key')
def get_site_key():
    return app.config.get('RECAPTCHA_SITE_KEY')
```
used only by contact form. I think for clarity sake it's best to keep them together. Creating separate `portality.view.recaptcha` just for that one short method seems to be overreaction while `portality.view.doajservices` sounds like a good place but at the same time separates the functionality of one small module. What do you think?

amdomanska:
@richard-jones for testing purposes you may use:
```
RECAPTCHA_SITE_KEY = '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI'
RECAPTCHA_SECRET_KEY = "6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe"
```

amdomanska:
I've added replacements for recaptcha keys to settings.py. These are for dev tests purposes only and should be overriden in `dev.cfg` by the real keys.

portality/templates/doaj/contact.html
amdomanska: We could. But `onload=doaj.contact.onLoadCallback` is not accepted. The workaround would be to  do something like:
```
var script = document.createElement('script');

script.id = 'container'
script.src = '//www.google.com/recaptcha/api.js?render=explicit';
script.async = true;
script.defer = true;

script.onload = () => { ... };

document.body.appendChild(script);
```
in the contact.js file in `jQuery(document).ready(function($) `. I am not really sure if that will improve that code. And, tbh, I've struggled a bit to make it work. Could we maybe pair up sometime and have a look at it together?

amdomanska:
It was tested on the test server and works fine.

PR #1724, opened 2019-11-21
Change the code path back to cloo/doaj rather than cloo/doaj_python3

Steven-Eardley:
This is the prerequisite to running the `finalise_python3` upgrade script to finish the deployment. It changes the DOAJ code back to expecting the path to be `/home/cloo/doaj/src/doaj` - this means the previous deployment scripts and the wiki will be correct.

I've created this as a feature branch so that we can merge to develop and run the test server version of this script first.


Steven-Eardley:
I've now run the finalise script.

PR #1725, opened 2019-11-22
Feature/2201 reshuffle metadata form fields

amdomanska:
This branch is based on feature/275_article_metadata branch and therefor should be released with or after https://github.com/DOAJ/doajPM/issues/275.

Fields are reshuffled and ISSNs dropdowns content is restricted to the appropriate type of ISSN.

image

### DIFF
--- a/portality/dao.py
+++ b/portality/dao.py
@@ -376,6 +376,7 @@ class DomainObject(UserDict, object):
         r = None
         count = 0
         exception = None
+        data = json.dumps(qobj)
         while count < retry:
             count += 1
             try:
--- a/portality/models/journal.py
+++ b/portality/models/journal.py
@@ -27,21 +27,25 @@ class JournalLikeObject(dataobj.DataObj, DomainObject):
         return records
 
     @classmethod
-    def _return_one_type_issns(cls,q):
+    def _return_one_type_issns(cls,q, type):
+        one_type = []
         res = cls.query(q=q.query())
-        issns = [hit.get("fields", {}).get("bibjson.identifier.id")[0] for hit in res.get("hits", {}).get("hits", [])]
-        return issns
+        issns = [hit for hit in res.get("hits", {}).get("hits", [])[0].get("_source", {}).get("bibjson", {}).get("identifier", {})]
+        for issn in issns:
+            if issn.get("type") == type:
+                one_type.append(issn.get("id"))
+        return one_type
 
     @classmethod
     def eissns_by_owner(cls, owner):
         q = OneTypeIssnQuery(owner, "eissn")
-        return cls._return_one_type_issns(q)
+        return cls._return_one_type_issns(q, "eissn")
 
 
     @classmethod
     def pissns_by_owner(cls, owner):
         q = OneTypeIssnQuery(owner, "pissn")
-        return cls._return_one_type_issns(q)
+        return cls._return_one_type_issns(q, "pissn")
 
     @classmethod
     def issns_by_owner(cls, owner):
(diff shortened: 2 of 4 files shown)


portality/dao.py
```diff
@@ -376,6 +376,7 @@ def send_query(cls, qobj, endpoint='_search', recid='', retry=50, return_raw_res
         r = None
         count = 0
         exception = None
+        data = json.dumps(qobj)
```
richard-jones: Probably some hang-over from testing, that can be removed

portality/models/journal.py
```diff
@@ -26,10 +26,32 @@ def find_by_issn(cls, issns, in_doaj=None, max=10):
...
+    @classmethod
+    def pissns_by_owner(cls, owner):
+        q = OneTypeIssnQuery(owner, "pissn")
+        return cls._return_one_type_issns(q, "pissn")
+
     @classmethod
     def issns_by_owner(cls, owner):
         q = IssnQuery(owner)
         res = cls.query(q=q.query())
+        r = res.get("facets", {}).get("issns", {}).get("terms", [])
```
richard-jones: This line doesn't seem to be used

portality/models/journal.py
```diff
@@ -1339,6 +1361,26 @@ def all_in_doaj(self):
             q["sort"] = [{"bibjson.title.exact" : {"order" : "asc"}}]
         return q
 
+class OneTypeIssnQuery(object):
+    base_query = {
+       "query" : {
+            "bool":{
+                "must" : [
+                    {"term" : { "admin.owner.exact" : "<owner id here>" }},
+                    {"term" : { "bibjson.identifier.type" : "<issn type here>"}}
```
richard-jones: Probably best to use `bibjson.identifier.type.exact`

portality/models/journal.py
```diff
@@ -26,10 +26,32 @@ def find_by_issn(cls, issns, in_doaj=None, max=10):
...
 
+    @classmethod
+    def _return_one_type_issns(cls,q, type):
+        one_type = []
+        res = cls.query(q=q.query())
+        issns = [hit for hit in res.get("hits", {}).get("hits", [])[0].get("_source", {}).get("bibjson", {}).get("identifier", {})]
+        for issn in issns:
+            if issn.get("type") == type:
+                one_type.append(issn.get("id"))
+        return one_type
```
richard-jones: This approach has a few problems, which might mean we re-think how we query for individual issn types:

1. It's quite computationally expensive for large result sets.  It would be better to use an aggregation, but of course we can't because of the shape of the data

2. This particular implementation only gets the first 10 results, so no matter how many journals you have, you will only see a max of 10 issns in the list.  To do it properly you'll want to page through the results, which further to (1) adds more computational expense.

I'd suggest the way to do this instead is to index e and p issns in `index.eissn` and `index.pissn` and hit the index for an aggregation on those fields, constrained by owner.  We could add those fields easily enough in `Article._generate_index()`.

If we can make it so that that all happens in one request, with both aggregations on it, we can make it much more efficient.

portality/models/journal.py
richard-jones: (though see note above about changing this approach)

portality/view/forms.py
```diff
@@ -67,12 +67,15 @@ def __init__(self, *args, **kwargs):
         try:
             if not current_user.is_anonymous:
                 if "admin" in current_user.role and "id" in kwargs :
-                    issns = models.Journal.issns_by_owner(a.get_owner())
+                    eissns = models.Journal.eissns_by_owner(a.get_owner())
+                    pissns = models.Journal.pissns_by_owner(a.get_owner())
```
richard-jones: Yeah, so, taking my approach above of querying for them both at the same time, we could introduce a function which returns a tuple, like:

```
eissns, pissns = models.Journal.issns_by_type_by_owner(a.get_owner())
```

amdomanska:
Taking into consideration the complexity of pissn/eissn change I would suggest to remove this change from this PR (make it simply about reshuffling the fields and change of messages) and creating the separate one form pissn/eissn. That way we will be able to improve user experience sooner by simply changing the look of the form. @richard-jones if you think it's a good idea I will do it asap.

amdomanska:
Waiting for Orcid and new metadata form to be merged.

amdomanska:
This is going to be released with https://github.com/DOAJ/doajPM/issues/2340

PR #1726, opened 2019-11-22
Remove ISSN check in deduplication system

amdomanska:
The hotfix removes ISSN param and check in deduplication system - always checks the whole DB so that there is no possibility to introduce articles with the same DOI and URL but different ISSNs as an existing article.

Fixes: https://github.com/DOAJ/doajPM/issues/2228


amdomanska:
This branch is merged to https://github.com/DOAJ/doaj/pull/1719.

PR #1727, opened 2019-11-22
ORCID iD

amdomanska:
#Closes: https://github.com/DOAJ/doajPM/issues/659


amdomanska:
Article page with the Orcid ID link:

Screenshot from 2019-11-22 15-27-36

Article metadata form with Orcid ID field:

image

with validation (needs to be changed to red):
image

Article metadata retrieved via API:

image

amdomanska:
Done:

- Orcid ID added to ArticleBibjson model
- Orcid ID logo with a link to Orcid profile add to article landing page for authors when available
- Orcid ID dropdown option added in public and admin search interfaces
- Orcid ID added to incoming and outgoing API objects and API documentation updated
- Orcid ID field added to Article Metadata Form (with validation)
- Orcid ID element added to DOAJ XSD (with regex validation)
- Unit tests updated (these will need to be updated after new Admin Metadata Form release again)

doajtest/fixtures/bibjson.py
```diff
@@ -171,7 +171,8 @@ def article_bibjson(cls):
     "author" : [
         {
             "name" : "Test",
-            "affiliation" : "University of Life"
+            "affiliation" : "University of Life",
+            "orcid_id" : "0000-0001-1234-1234"
```
richard-jones: Exchange this for the canonical form

doajtest/fixtures/article.py
```diff
@@ -232,7 +240,8 @@ def make_article_apido_struct():
         "author" : [
             {
                 "name" : "The Author",
-                "affiliation" : "University Cottage Labs"
+                "affiliation" : "University Cottage Labs",
+                "orcid_id" : "0000-0001-1234-1234"
```
richard-jones: Exchange for canonical form

doajtest/unit/test_api_discovery.py
```diff
@@ -140,6 +140,7 @@ def test_02_articles(self):
             bj.add_identifier(bj.P_ISSN, "{x}000-0000".format(x=i))
             bj.add_identifier(bj.DOI, "10.test/{x}".format(x=i))
             bj.publisher = "Test Publisher {x}".format(x=i)
+            bj.add_author("Agnieszka Domanska", "CL University", "0000-0001-1234-1234")
```
richard-jones: exchange for canonical form

doajtest/unit/test_models.py
```diff
@@ -679,6 +679,8 @@ def test_16_article_bibjson(self):
         assert bj.journal_issns == ["1234-5678", "9876-5432"]
         assert bj.publisher == "IEEE"
         assert bj.author[0].get("name") == "Test"
+        assert bj.author[0].get("affiliation") == "University of Life"
+        assert bj.author[0].get("orcid_id") == "0000-0001-1234-1234", "received: {}".format(bj.author[0].get("orcid_id"))
```
richard-jones: exchange for canonical form

portality/static/js/edges/admin.journalarticle.edge.js
```diff
@@ -299,7 +299,8 @@ $.extend(true, doaj, {
                         {'display':'Publisher','field':'index.publisher'},
 
                         {'display':'Article: Abstract','field':'bibjson.abstract'},
-                        {'display':'Article: Author','field':'bibjson.author.name'},
+                        {'display':'Article: Author\'s name','field':'bibjson.author.name'},
+                        {'display':'Article: Author\'s Orcid ID','field':'bibjson.author.orchid_id'},
```
richard-jones: typo - `orchid_id` -> `orcid_id`

portality/static/js/edges/public.journalarticle.edge.js
```diff
@@ -268,7 +268,8 @@ $.extend(true, doaj, {
                         {'display':'Article: Abstract','field':'bibjson.abstract'},
                         {'display':'Article: Year','field':'bibjson.year'},
                         {'display':'Article: Journal Title','field':'bibjson.journal.title'},
-                        {'display':'Article: Author','field':'bibjson.author.name'},
+                        {'display':'Article: Author\'s name','field':'bibjson.author.name'},
+                        {'display':'Article: Author\'s Orcid ID','field':'bibjson.author.orcid_id'},
```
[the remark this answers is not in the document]
richard-jones: On the public search, should we also include the link to the orcid for the author where it's present?

portality/templates/_formhelpers.html
```diff
@@ -219,11 +219,11 @@
             <div class="col-md-9">
               {{ subfield(**kwargs) }}
               {% if subfield.errors %}
-              <ul class="errors">
                   {% for error in subfield.errors %}
-                  <li>{{ error }}</li>
+                      <ul class="errors">
+                        <li>{{ error }}</li>
+                      </ul>
```
richard-jones: This seems odd, is that right?  It looks like for each error you'll end up with an unordered list, containing a single list item?

portality/templates/publisher/metadata.html
```diff
@@ -28,13 +28,20 @@ <h4 class="red">There is a problem with the submitted form.</h4>
                         <button type="submit" id="remove_{{subfield.short_name}}" name="remove_{{subfield.short_name}}" class="btn btn-danger remove_button">
                           <i class="glyphicon glyphicon-remove"></i>
                         </button>
+                        {% for error in subfield.errors %}
+                            <ul class="errors">
+                                <li>{{ subfield.errors.get(error)[0] }}</li>
+                            </ul>
+                        {% endfor %}
```
richard-jones: Same comment as above here - won't you end up with multiple ul elements each with a single li inside them?

portality/templates/publisher/metadata.html
```diff
@@ -28,13 +28,20 @@ <h4 class="red">There is a problem with the submitted form.</h4>
...
                     <input type="submit" class="btn btn-info" name="more_authors" value="Add More Authors">
                     {% if author_error %}
-                    <ul class="errors">
-                        <li>Please provide at least one author</li>
-                    </ul>
+                    <div class="error">
+                        <ul class="errors">
+                            <li>Please provide at least one author</li>
+                        </ul>
+                    </div>
```
richard-jones: What's the necessity for this extra div with the same class as the enclosed ul?

portality/models/article.py
```diff
@@ -947,6 +951,7 @@ class ArticleIssueQuery(object):
             "bibjson.journal.number",
             "bibjson.title",
             "bibjson.author.name",
+            "bibjson.author.orcid_id",
```
richard-jones: It looks to me like this whole class is dead, and should probably be removed along with the function that calls it.  Unless something led you specifically to add this field here?

portality/view/forms.py
```diff
@@ -24,6 +24,8 @@
 
 DOI_REGEX = regex.DOI_COMPILED
 DOI_ERROR = 'Invalid DOI.  A DOI can optionally start with a prefix (such as "doi:"), followed by "10." and the remainder of the identifier'
+ORCID_REGEX = re.compile(r'^[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{4}$')
```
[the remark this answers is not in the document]
richard-jones: Once it's in the `regex` package it can be re-used in the validation on the API too.

portality/static/js/edges/public.journalarticle.edge.js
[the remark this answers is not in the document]
amdomanska: > On the public search, should we also include the link to the orcid for the author where it's present?

I would say that if so then only by id logo as at article landing page - check the screenshot provided above. There is not much space for the whole link, especially when a few authors. Personally I wouldn't do that - in my opinion the search page should only display the most important key infos about the article and for more we can go to the article's landing page. @dommitchell what do you think?

portality/models/article.py
amdomanska: It is used in `get_by_volume_issue` that's why I added that but I haven't checked that this method is not used anywhere else indeed. I'll remove it.

doajtest/unit/resources/article_uploads.xml
```diff
@@ -1657,5 +1656,89 @@
...
+        <startPage/>
+        <endPage/>
+        <documentType>article</documentType>
+        <title language="fre">Imaginaires autochtones contemporains. Introduction</title>
+        <authors>
+            <author>
+                <name>Papillon, Joëlle</name>
+                <email/>
+                <affiliationId>1</affiliationId>
+                <orcid_id>0000-0001-1234-1234</orcid_id>
```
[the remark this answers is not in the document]
amdomanska: I realized that `orcid` would be enough when I already had `orcid_id` everywhere so decided to stick with it ;) They use ORCID iD so I believe this should be fine.

portality/templates/_formhelpers.html
amdomanska: My mistake. Fixed.

portality/templates/publisher/metadata.html
amdomanska: It is not needed here apparently. But it is still needed above for orcid id error cause without it the red colour is not inherited. This is not the most elegant way I guess but the simplest one I think.

portality/templates/publisher/metadata.html
```diff
@@ -28,13 +28,20 @@ <h4 class="red">There is a problem with the submitted form.</h4>
                         <button type="submit" id="remove_{{subfield.short_name}}" name="remove_{{subfield.short_name}}" class="btn btn-danger remove_button">
                           <i class="glyphicon glyphicon-remove"></i>
                         </button>
+                        <div class="error">
```
amdomanska: @richard-jones  This div here is necessary to inherit red color for subfield.errors, tbh I don't entirely understand why it works like this. There would probably be some more elegant way to do this if we would rebuild CSS but this way it's much easier and less time-consuming. Maybe we could have a look at it later?

amdomanska:
@richard-jones I fixed/answered all of your comments but responsiveness. If you could have one more look at it while I'll try to do something with responsiveness that would be great.

amdomanska:
I played a bit with Author form fields and I think indeed moving them to columns will look best if only we will change `X` button to `Remove Author` button. My proposition is:
image

richard-jones:
I like your proposed change to the author entry section, yes.

portality/static/js/edges/public.journalarticle.edge.js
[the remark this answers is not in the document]
richard-jones: I agree, it should just be the "iD" logo next to the name, the same way you have done it for the article pages.

portality/templates/publisher/metadata.html
richard-jones: It doesn't seem to be causing the red colour to be inherited even so - on my screen the error came up in black text.

portality/templates/publisher/metadata.html
richard-jones: It doesn't seem to be causing the red colour to be inherited even so - on my screen the error came up in black text.  We can look at this together, yes.

amdomanska: It was. The red color disappeared after changing the orientation of the fields.

image

portality/static/js/edges/public.journalarticle.edge.js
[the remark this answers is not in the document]
amdomanska: @richard-jones 
image
What do you think? iD logo is a link ofc

amdomanska:
Author fields with error:
image

amdomanska:
>     * The form fields look much better the way they are now, can we just straighten them up so all the boxes are aligned vertically.  Also, if we can make the boxes wider that would be good, as the ORCIDs could be quite long.  Also, add a placeholder for the ORCID which indicates the format of the entry.

image


>     * If you submit an invalid ORCID you get an error message, but it isn't in red, and it appears after the "Remove author" button, but it should be under the ORCID field.

image

amdomanska:
>     * Could you review the relevant Functional Tests and make sure that any additions for ORCIDs are included, and then list the relevant tests on the issue so that we can ask the DOAJ team to run them?

I'm on it. Meanwhile ready for review.

portality/static/js/edges/public.journalarticle.edge.js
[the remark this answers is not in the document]
richard-jones: Looks excellent

amdomanska:
I've added suitable Functional Scripts. These can be found in the sheets:
Administrative Search/Journals and Articles
Article Metadata Upload Form
Article XML Upload
For now for easier identification they are added with green background.

amdomanska:
@richard-jones @Steven-Eardley if you could please review this PR I think we could put that up to the test server soon.

amdomanska:
It's on the test server now.

amdomanska:
TODO:
- [ ] logo on public search interface dissapeared - fix it

amdomanska:
TO DO: remove print statement

PR #1728, opened 2019-11-27
upgrade esprit version and make local store store in binary mode

richard-jones:
This PR introduces some improvements in the local and temp store implementations to handle byte or string streams, and also introduces buffered saves to limit local memory usage.  There's some basic tests for this.

We also set the version of esprit to the current master tip, which is the fully updated and working python 3 version.

### DIFF
--- a/portality/store.py
+++ b/portality/store.py
@@ -175,8 +175,13 @@ class StoreLocal(Store):
         if source_path:
             shutil.copyfile(source_path, tpath)
         elif source_stream:
-            with open(tpath, "wb") as f:
-                f.write(source_stream.read())
+            buffer_size = 16777216
+            data = source_stream.read(buffer_size)
+            mode = "w" if isinstance(data, str) else "wb"
+            with open(tpath, mode) as f:
+                while data:
+                    f.write(data)
+                    data = source_stream.read(buffer_size)
 
     def exists(self, container_id):
         cpath = os.path.join(self.dir, container_id)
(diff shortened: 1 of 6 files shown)


portality/store.py
```diff
@@ -175,8 +176,12 @@ def store(self, container_id, target_name, source_path=None, source_stream=None)
         if source_path:
             shutil.copyfile(source_path, tpath)
         elif source_stream:
-            with open(tpath, "w") as f:
-                f.write(source_stream.read())
+            data = source_stream.read(self.buffer_size)
+            mode = "w" if isinstance(data, str) else "wb"
```
amdomanska: Could you please explain to me briefly why these data may be of different type?

portality/store.py
```diff
@@ -217,12 +222,17 @@ def delete_file(self, container_id, target_name):
             else:
                 shutil.rmtree(cpath)
 
+    def size(self, container_id, target_name):
```
amdomanska: Is this method used anywhere else than the tests?

portality/store.py
richard-jones: unfortunately in python 3 `read` may return `bytes` or `str`, which are treated very differently.  It depends on how the source stream was initialised - for example StringIO returns a string, but BytesIO returns bytes, meanwhile opening a file with "b" or without will change the return type.  Also, we take streams off HTTP, and I don't think we can control whether that is a string or a byte output.

portality/store.py
richard-jones: no, just needed for a test, but could come in handy in the future.  StoreLocal specific.

PR #1731, opened 2020-01-03
Hotfix/2020 01 02 iss2253 api too many keywords

Steven-Eardley:
Relates to issue : https://github.com/DOAJ/doajPM/issues/2253

This fix prevents the dataobj custom Exception from intercepting the message text, which was stopping it from propagating in chained exceptions.

### DIFF
--- a/portality/lib/dataobj.py
+++ b/portality/lib/dataobj.py
@@ -222,12 +222,16 @@ def string_canonicalise(canon, allow_fail=False):
 ############################################################
 
 ############################################################
-## The core data object which manages all the interactions
-## with the underlying data member variable
+# The core data object which manages all the interactions
+# with the underlying data member variable
+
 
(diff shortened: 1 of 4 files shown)


portality/lib/dataobj.py
```diff
@@ -222,12 +222,16 @@ def sn(val):
...
+
 
 class DataObjException(Exception):
-    def __init__(self, msg, *args, **kwargs):
-        self.message = msg
+    def __init__(self, *args, **kwargs):
+        try:
+            self.message = args[0]
+        except IndexError:
+            self.message = ''
```
Steven-Eardley: I was trying to decide between 2 approaches here; I could have retained the signature and supply the message back into `args` with `args = (msg,) + args` but I decided since we already assume our message is the first positional argument this is equivalent and  cleaner.

PR #1733, opened 2020-01-08
UI Automated Tests

amdomanska:
Connects to: https://github.com/DOAJ/doajPM/issues/2173

This PR is created to give some insights into UI Automated Tests (with Selenium and Pytest). 2 simple tests have been implemented: basic public search (from the main site) and test of metadata form (assuming existing publisher account with a correct journal in DOAJ).

Pytest is a user-friendly test framework. Currently we use notests. The great majority of our test set can be run with pytest out of the box. 9 tests fail connected to email sending. However, this would have to be taken into consideration when implementing future UT. Maybe switching to pytest would be worth considering.
Screenshot from 2020-01-08 21-59-24

Selenium framework is easy to learn, straightforward in implementation and doesn't need extensive setup. The implementation itself is quite time-consuming but all UI functionality currently covered by our functionality tests could be automated.

The biggest downside of the UI tests is that they are really slow. They actually perform all actions on browser instance. Even though parallelization is easy they are still time-consuming. In my opinion, adding them to our "normal" test set is not an option and we would have to figure out another system (scheduling?).

In my opinion, we should start implementing the UI tests alongside our normal implementation. That way we could build our UI tests set slowly avoiding one-time big cost (implementation of the test of one UI feature is easy and short) and at the same time saving a lot of time for future functional tests, especially bearing in mind new application form.

As a great starting point for learning the technology and framework, I suggest this tutorial: https://blog.testproject.io/2019/07/16/open-source-test-automation-python-pytest-selenium-webdriver/
Going through it takes no more than a couple of hours and is enough to start implementing real-system tests.


amdomanska:
@Steven-Eardley do you think it would be reasonable to merge these tests to develop?

PR #1734, opened 2020-01-10
Fix publisher limiter search

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/1964

index.publisher array field is separated into two string fields: index.publisher and index.institution (see: https://github.com/DOAJ/doajPM/issues/1964#issuecomment-568100960 for discussion). To make the search work it is needed to run the simple migration that will regenerate index for all journals and suggestions. Run by: `python portality/upgrade.py -u portality/migrate/1964_separate_publisher_institution/migrate.json`


docs/openurl/outline.md
```diff
@@ -287,6 +287,7 @@ genre : "journal|issue|article|proceeding|conference|preprint|unknown" # journal
         "country" : "<country of journal publication>",
         "license" : [<list of titles of licences>],
         "publisher" : "<publisher>",
+        "institution" : "<institution>",
```
Steven-Eardley: There's something a bit odd about those old OpenURL docs, we should probably remove the DOAJ models from the 'useful info' section and point them to the main system documentation rather than repeating the models. They could be outdated in other ways.

docs/system/DATA_MODELS.md
```diff
@@ -88,6 +88,7 @@
         "country" : "<country of journal publication>",
         "license" : [<list of titles of licences>],
         "publisher" : "<publisher>",
+        "institution" : "<institution>",
```
Steven-Eardley: should we have these doc changes in the DataObj docs? docs/system/Journal.md and docs/system/Suggestion.md

Steven-Eardley: In addition, it looks here like you've added institution to the index of Article, but in the DataObj code below it's been added to bibjson.journal

amdomanska: It should be in bibjson.journal as it is already implemented there. Actually it also should be added to Article.index.institution to make this pr work correctly. And therefore the migration should be run on the articles as well.

amdomanska:
To sum it up:
1. array `index.publisher` is replaced by string fields `index.publisher` and `index.institution` for articles, suggestions and journals.
2. documentation is up to day for `bibjson` and `index` fields for above data obj models
3. openURL documentation points to other data obj docs instead of copying data models.

Steven-Eardley:
https://github.com/DOAJ/doajPM/wiki/How-To:-Review-a-data-model-change

amdomanska:
Model:
- [ ]  Has the field been added to the model struct
- [ ] Does the model have all the relevant getters/setters for the new field

    Storage:
- [ ] How will the ElasticSearch mapping update? (dynamic vs on app initialisation)
- [ ] Fields added - how should this be indexed and are there search considerations
- [ ] Fields removed - will the mapping fail, is reindex necessary
- [ ] Fields repurposed - will existing data still be valid, is a migration required

    Forms:
- [ ] Is the new field present in all relevant forms
- [ ] Does the new field have appropriate validation in the forms

    API(s):
- [ ] Has the field been added to the API incoming/outgoing model struct
- [ ] Does the new field have appropriate validation in the API
- [ ] Are the mappings to OpenURL correct still?

    Crosswalks:
- [ ] Is there a crosswalk from the forms to the core model object
- [ ] Is there a crosswalk from the core model object to the forms
- [ ] Is there a crosswalk from any incoming XML to the model object
- [ ] Is there a crosswalk from the model object to OAI DC
- [ ] Is there a crosswalk from the model object to OAI DOAJ XML
- [ ] Is there a crosswalk from the incoming API format to the model object
- [ ] Is there a crosswalk from the model object to the outgoing API format

    Data Exports
- [ ] Is there a crosswalk to the Journal CSV, and/or has the CSV generation been updated

    UI:
- [ ]  Is the new data displayed on the view page for the object
- [ ] Is the new data displayed in the search result view of the object

    Testing:
- [ ] Are the model fixtures updated to include the new data
- [ ] Are the form fixtures updated to include the new data
- [ ] Are the XML fixtures updated to include the new data
- [ ] Is the API for depositing the value tested
- [ ] Is the API for retrieving the value tested
- [ ] Are the crosswalks tested
- [ ] Are the model objects creation and getters/setters tested
- [ ] Is the formcontext tested
- [ ] Are the Functional Test Scripts updated

    Documentation
- [ ] Has the field been added to the API documentation
- [ ] Has the field been added to the OAI-PMH documentation
- [ ] Have the auto-generated model documents been regenerated?
- [ ] Are there any other places the field needs to be documented?

    Migration
- [ ]  Is a migration required for this field?
- [ ] Has a migration script been produced and tested

amdomanska:
@Steven-Eardley I believe it's all good. If only you could also run a few search tests locally (it will require the migration first) just to be sure I haven't missed anything I think it will be ready to go.

amdomanska:
@dommitchell now when the Institution field is excluded from searching by Publisher limiter would you like to add separate Institution limiter to the search? There is now no direct way to search by Institution.

portality/models/article.py
```diff
@@ -864,6 +864,7 @@ def vancouver_citation(self):
                         "volume" : {"coerce" : "unicode"},
                         "number" : {"coerce" : "unicode"},
                         "publisher" : {"coerce" : "unicode"},
+                        "institution" : {"coerce" : "unicode"},
```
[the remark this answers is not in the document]
amdomanska: We currently don't use Institution on Journals as well. I would wait for Dom's response and then keep Journals and Articles consistent (with either Institution in `index` for soon-developed search or Institution removed from index field everywhere if Institution search limiter is not going to be developed anytime in the future)

portality/migrate/1964_separate_publisher_institution/migrate.json
```diff
@@ -0,0 +1,20 @@
...
+			"init_with_model" : true,
+			"keepalive" : "20m"
+		},
+		{
+			"type": "journal",
+			"init_with_model": true,
+			"keepalive": "20m"
+		},
+		{
+			"type": "article",
```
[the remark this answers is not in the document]
amdomanska: Migration should be run on articles with `add_journal_metadata()`. The decision about adding (or not) institution to index.article I would leave till decision on https://github.com/DOAJ/doaj/pull/1734#issuecomment-575081725. I would need it consistent with journals. But of course, you're right: either `generate_index()` needs to assign a correct value to index.institution or `index.institution` should be removed from structs.

richard-jones:
This is now superseded by #1744

PR #1737, opened 2020-01-28
500 API script

amdomanska:
@Steven-Eardley could you please have a look at it and run in on live to produce csv file that contains all the journals that result with 500 when searched via API?

Thank you in advance

Connects to: https://github.com/DOAJ/doajPM/issues/2268

### DIFF
--- a/portality/scripts/all_returning_500_via_api.py
+++ b/portality/scripts/all_returning_500_via_api.py
@@ -1,0 +1,38 @@
+import requests
+
+from portality import models
+from portality.core import app
+import esprit
+import re
+import csv
+
+INVALID_URLS = {
+        "query": {
+            "match_all": {}
+        }
+}
+
+if __name__ == "__main__":
+
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-o", "--out", help="output file path")
+    args = parser.parse_args()
+
+    if not args.out:
+        print("Please specify an output file path with the -o option")
+        parser.print_help()
+        exit()
+
+    conn = esprit.raw.make_connection(None, app.config["ELASTIC_SEARCH_HOST"], None, app.config["ELASTIC_SEARCH_DB"])
+
+    with open(args.out, "w", encoding="utf-8") as f:
+        writer = csv.writer(f)
+        writer.writerow(["ID"])
+
+        for j in esprit.tasks.scroll(conn, models.Journal.__type__, q=INVALID_URLS, page_size=100, keepalive='5m'):
+            journal = models.Journal(_source=j)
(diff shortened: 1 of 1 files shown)


portality/scripts/all_returning_500_via_api.py
```diff
@@ -0,0 +1,38 @@
...
+    if not args.out:
+        print("Please specify an output file path with the -o option")
+        parser.print_help()
+        exit()
+
+    conn = esprit.raw.make_connection(None, app.config["ELASTIC_SEARCH_HOST"], None, app.config["ELASTIC_SEARCH_DB"])
+
+    with open(args.out, "w", encoding="utf-8") as f:
+        writer = csv.writer(f)
+        writer.writerow(["ID"])
```
Steven-Eardley: your header row has a different number of columns than your data, which could cause problems

portality/scripts/all_returning_500_via_api.py
```diff
@@ -0,0 +1,38 @@
...
+if __name__ == "__main__":
+
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-o", "--out", help="output file path")
+    args = parser.parse_args()
+
+    if not args.out:
+        print("Please specify an output file path with the -o option")
```
Steven-Eardley: Argparse has its own `required` which will do exactly this https://docs.python.org/3/library/argparse.html#required

portality/scripts/all_returning_500_via_api.py
```diff
@@ -0,0 +1,38 @@
...
+
+    conn = esprit.raw.make_connection(None, app.config["ELASTIC_SEARCH_HOST"], None, app.config["ELASTIC_SEARCH_DB"])
+
+    with open(args.out, "w", encoding="utf-8") as f:
+        writer = csv.writer(f)
+        writer.writerow(["ID"])
+
+        for j in esprit.tasks.scroll(conn, models.Journal.__type__, q=INVALID_URLS, page_size=100, keepalive='5m'):
+            journal = models.Journal(_source=j)
+            r = requests.get("http://doaj.org/api/v1/journals/{}".format(journal.id))
```
Steven-Eardley: you're going to be knocking on the API quite a lot with this, and there's no rate limiting here.

I'd prefer it if the script just tried to initialise the journal as an OutgoingJournal object (although since we've relaxed this recently we may need to include the old one in this script explicitly)

richard-jones:
Just held for a week, so that we can get moving on the application form work.  We should revisit this w/c 2020-03-16.

Note that now that the fix for the API issue has been deployed, the script won't error as expected, so we will need to get under the hood to evaluate whether the records would have raised a 500.

PR #1738, opened 2020-01-30
Relax requirements for outgoing journal struct

amdomanska:
This PR does not require any URL for outgoing journals - unicode is accepted. That way even if url is not correct the journal metadata still will be displayed.

#Connects to: https://github.com/DOAJ/doajPM/issues/2268#issuecomment-579190600


amdomanska:
I can confirm that now outgoing journal struct is the same as mappings in elasticsearch so the typing error shouldn't cause 500 anymore.

PR #1739, opened 2020-02-06
make regex.py read the orcid regex from xsd

amdomanska:
Fixes: https://github.com/DOAJ/doajPM/issues/2275

1. The main places where orcid and doi regexes are saved is our XSD
2. regex.py instead of duplicating them reads them from XSD and process them so that they are usable by our system:
  2a. adds `^` and `$` which are default in XSD
  2b. for DOI adds capturing group (not supported in XSD but used later to normalise DOI)

### DIFF
--- a/portality/regex.py
+++ b/portality/regex.py
@@ -4,8 +4,7 @@ from lxml import etree
 
 from portality.core import app
 
-DOI = r"^((https?://)?((dx\.)?doi\.org/|hdl\.handle\.net/)|doi:|info:doi/|info:hdl/)?(?P<id>10\.\S+/\S+)$"
-DOI_COMPILED = re.compile(DOI, re.IGNORECASE)
+DOI_XPATH = '/xs:schema[@version="1.2"]/xs:complexType[@name="recordType"]/xs:sequence/xs:element[@name="doi"]/xs:simpleType/xs:restriction/xs:pattern/@value'
 ORCID_XPATH = '/xs:schema[@version="1.2"]/xs:complexType[@name="recordType"]/xs:sequence/xs:element[@name="authors"]/xs:complexType/xs:sequence/xs:element[@name="author"]/xs:complexType/xs:sequence/xs:element[@name="orcid_id"]/xs:simpleType/xs:restriction/xs:pattern/@value'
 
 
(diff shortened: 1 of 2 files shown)


portality/regex.py
```diff
@@ -1,16 +1,39 @@
 import re
 
-DOI = r"^((https?://)?((dx\.)?doi\.org/|hdl\.handle\.net/)|doi:|info:doi/|info:hdl/)?(?P<id>10\.\S+/\S+)$"
-DOI_COMPILED = re.compile(DOI, re.IGNORECASE)
-ORCID = r"^https://orcid\.org/[0-9]{4}-[0-9]{4}-[0-9]{4}-\d{3}[\dX]$"
+from lxml import etree
+
+from portality.core import app
+
+DOI_XPATH = '/xs:schema[@version="1.2"]/xs:complexType[@name="recordType"]/xs:sequence/xs:element[@name="doi"]/xs:simpleType/xs:restriction/xs:pattern/@value'
```
Steven-Eardley: I don't think specifying all the way from the root is the best way to do this - not least because we'll be changing XSD version in future and we don't want to change our code with that. I recommend using a `find` which lets you use a more minimal XPATH.

portality/regex.py
```diff
@@ -1,16 +1,39 @@
...
+ORCID_XPATH = '/xs:schema[@version="1.2"]/xs:complexType[@name="recordType"]/xs:sequence/xs:element[@name="authors"]/xs:complexType/xs:sequence/xs:element[@name="author"]/xs:complexType/xs:sequence/xs:element[@name="orcid_id"]/xs:simpleType/xs:restriction/xs:pattern/@value'
+
+
+def _read_value(id):
```
Steven-Eardley: we shouldn't use `id` as that's a builtin function

portality/regex.py
```diff
@@ -1,16 +1,39 @@
...
+
+ORCID = r"^" + _read_value("orcid")[0] + r"$"
 ORCID_COMPILED = re.compile(ORCID)
+doi_data = _read_value("doi")[0]
+DOI = r"^" + doi_data[0] + r"(?P<id>" + doi_data[1] + doi_data[2] + r")$"
+DOI_COMPILED = re.compile(DOI, re.IGNORECASE)
 
 def is_match(pattern, string, *args, **kwargs):
     match = re.match(pattern, string, *args, **kwargs)
+    print(match)
```
Steven-Eardley: debug print statement should be removed

PR #1740, opened 2020-02-17
add new application/journal data model proposal

richard-jones:


portality/models/v2/bibjson.py
```diff
@@ -564,6 +568,14 @@ def publication_time(self):
     def publication_time(self, weeks):
         self.publication_time_weeks = weeks
 
+    @property
+    def publisher(self):
+        return self.publisher_name
```
amdomanska: I'm not sure why you want to return publisher_name for publisher getter. If we now have publisher_name getter maybe here it would be better to return the whole publisher struct? For me it would be more intuitive.

richard-jones: Agreed that this will remain, and we will add the deprecation library to track these deprecated methods

PR #1743, opened 2020-03-02
Feature/af redesign  api v2

amdomanska:
Connects to: https://github.com/DOAJ/doajPM/issues/2289

If you could please review new API data models I would be grateful.


portality/api/v2/data_objects/application.py
```diff
@@ -0,0 +1,379 @@
...
+                                "currency": {"coerce": "unicode"},
+                                "price": {"coerce": "integer"}
+                            }
+                        }
+                    }
+                },
+                "article": {
+                    "fields": {
+                        "embedded_license": {"coerce": "bool"},
+                        "embedded_license_url": {"coerce": "unicode"},
```
richard-jones: `embedded_example_license_url`

portality/api/v2/data_objects/application.py
```diff
@@ -0,0 +1,379 @@
...
+                    }
+                },
+                "editorial": {
+                    "fields": {
+                        "review_url": {"coerce": "unicode"},
+                        "board_url": {"coerce": "unicode"}
+                    },
+                    "lists": {
+                        "review_process": {"contains" : "field", "coerce": "unicode",
+                                           "allowed_values": ["Editorial review", "Peer review", "Blind peer review",
```
richard-jones: I've not specified allowed values in the main model struct, but perhaps we should.

portality/api/v2/data_objects/application.py
```diff
@@ -0,0 +1,379 @@
...
+                "waiver": {
+                    "requred": ["has_waiver"]
+                },
+            }
+        }
+    }
+}
+
+
+class IncomingApplication(dataobj.DataObj, swagger.SwaggerSupport):
```
richard-jones: I'll just make the comment here, but it applies everywhere that we use `DataObj`, is that we should use `SeamlessMixin` for the new version of the API (the v1 should continue to use `DataObj`).

portality/api/v2/data_objects/common_journal_application.py
```diff
@@ -0,0 +1,30 @@
...
+
+class OutgoingCommonJournalApplication(dataobj.DataObj, swagger.SwaggerSupport):
+
+    @classmethod
+    def from_model(cls, journal_or_app):
+        d = deepcopy(journal_or_app.data)
+
+        # we need to re-write the preservation section
+        # joa = d.get("bibjson", {}).get("preservation")
+        joa = journal_or_app.bibjson().preservation
```
richard-jones: These kinds of translations should go away with the new version of the API, and we should stick to the core model format as much as possible.

portality/api/v2/data_objects/journal.py
```diff
@@ -0,0 +1,201 @@
+from portality import models
+
+from portality.api.v2.data_objects.common_journal_application import OutgoingCommonJournalApplication
+
+# we only have outgoing journals for the moment
+JOURNAL_STRUCT = {
```
richard-jones: There's a bit of repetition here, as the structs for different model objects are declared independently.  Structs can overlay, so we should take this opportunity to normalise them in the v2 of the API.

See an example here: https://github.com/DOAJ/doaj/blob/feature/af_redesign__datamodel/portality/models/v2/journal.py#L404 (and follow your nose into the struct declarations)

portality/models/v2/shared_structs.py
```diff
@@ -57,7 +57,7 @@
                 "article" : {
                     "fields" : {
                         "embedded_license" : {"coerce" : "bool"},
-                        "embedded_license_example" : {"coerce" : "unicode"},
+                        "embedded_license_url" : {"coerce" : "unicode"},
```
richard-jones: I have fixed this in my branch, so you should get that on merge

portality/models/v2/bibjson.py
```diff
@@ -184,7 +184,7 @@ def add_subject(self, scheme, term, code=None):
         self.__seamless__.add_to_list_with_struct("subject", sobj)
 
     @property
-    def apcs(self):
+    def apc(self):
```
richard-jones: I'll change this in my branch and update all the downstream stuff like the tests

portality/api/v2/data_objects/application.py
amdomanska: @richard-jones or `embedded_license_example_url`?

portality/api/v2/data_objects/journal.py
amdomanska: Yep, that's the plan. It was just easier to write them down first in a full version.

PR #1747, opened 2020-03-17
remove orcid_id logo, add orcid profile link to authors name, support…

amdomanska:
… AND and OR in api ES queries correctly

### DIFF
--- a/portality/api/v1/discovery.py
+++ b/portality/api/v1/discovery.py
@@ -89,7 +89,10 @@ def escape(query):
         # colons after keywords are: first one and every first one after & or |
         parts = q.split(":")
         for i in range(1, len(parts)-1):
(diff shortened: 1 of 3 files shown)


portality/api/v1/discovery.py
```diff
@@ -89,7 +89,10 @@ def colon_escaper(q):
         # colons after keywords are: first one and every first one after & or |
         parts = q.split(":")
         for i in range(1, len(parts)-1):
-            if '&' not in parts[i] and '|' not in parts[i] and not parts[i].endswith('\\'):
+            if '&' not in parts[i] \
```
Steven-Eardley: we've lost the check for `'|'` here - is that intentional? (the comment above has remained the same)

PR #1750, opened 2020-04-21
Hotfix/2302 2344 article metadata errors

amdomanska:
Fixed errors:

1. for empty abstract elements the Type Error is not raised
https://github.com/DOAJ/doajPM/issues/2344#event-3252405301

2. An empty form is returned after successful upload of article's metadata:
https://github.com/DOAJ/doajPM/issues/2402

If ANY error occurred (including duplicate errors) returned form is pre-filled with last metadata entered for the easy proof-reading. Is that correct behaviour?

I am not able to reproduce the error from p1: https://github.com/DOAJ/doajPM/issues/2402#issue-603246918

3. If admin uploads the article without upload_article_id (like via publisher's metadata form or API) it is treated as publishers upload) - even though not used in practical context it makes testing easier for admin accounts.

### DIFF
--- a/portality/formcontext/formcontext.py
+++ b/portality/formcontext/formcontext.py
@@ -1885,7 +1885,7 @@ class PublisherMetadataForm(MetadataForm):
         self.template = "publisher/metadata.html"
 
     def render_template(self, **kwargs):
-        self._check_for_author_errors()
+        self._check_for_author_errors(**kwargs)
         if "validated" in kwargs and kwargs["validated"] == True:
             self.blank_form()
         return render_template(self.template, form=self.form, form_context=self, author_error=self.author_error)
@@ -1899,6 +1899,6 @@ class AdminMetadataArticleForm(MetadataForm):
         self.template = "admin/article_metadata.html"
 
     def render_template(self, **kwargs):
-        self._check_for_author_errors()
+        self._check_for_author_errors(**kwargs)
         return render_template(self.template, form=self.form, form_context=self, author_error=self.author_error)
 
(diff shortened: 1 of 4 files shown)


portality/formcontext/formcontext.py
```diff
@@ -1885,6 +1884,11 @@ def __init__(self, source, form_data, user):
     def set_template(self):
         self.template = "publisher/metadata.html"
 
+    def render_template(self, **kwargs):
+        self._check_for_author_errors()
```
Steven-Eardley, line 1888: you're not calling `_check_for_author_errors()` with kwargs, so will it find anything?

portality/view/publisher.py
```diff
@@ -196,15 +196,17 @@ def metadata():
...
             try:
                 fc.finalise()
+                validated = True
             except ArticleMergeConflict:
                 Messages.flash(Messages.ARTICLE_METADATA_MERGE_CONFLICT)
             except DuplicateArticleException:
                 Messages.flash(Messages.ARTICLE_METADATA_UPDATE_CONFLICT)
 
-        return fc.render_template()
+        return fc.render_template(validated=validated)
```
Steven-Eardley, line 209: oh this is rather cool!

portality/formcontext/formcontext.py
amdomanska: Good point.

portality/view/publisher.py
amdomanska: is it? I'm glad :D

PR #1754, opened 2020-06-09
Hotfix/2433 sage update

amdomanska:
script for:

https://github.com/DOAJ/doajPM/issues/2433


portality/scripts/sage_update.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if not args.out:
+        print("Please specify an output file path with the -o option")
+        parser.print_help()
+        exit()
+
+    with open(args.out, "w", encoding="utf-8") as f:
+        writer = csv.writer(f)
+
+        file = './resources/SAGE_jnls_in_DOAJ.edited.xlsx'
```
Steven-Eardley, line 21: probably best to keep the file outside our version control. upload it to /tmp/ or the home dir or similar on the server. There's no need to have it persist and bulk up our repo.

(since it's now been committed to this branch, it may be worth making a new branch & PR without it so it isn't in version control at all, or you can go through a 'purge file from repo' procedure)

requirements.txt
```diff
@@ -11,3 +11,30 @@
 
 # now run the normal setup.py and get all the normal PyPI dependencies from there
 -e .
+
+boto3~=1.10.5
```
Steven-Eardley, line 15: most of these will already be in setup.py - and any new requirements should go there rather than here in requirements.txt

portality/scripts/sage_update.py
```diff
@@ -0,0 +1,94 @@
...
+                                         sa=current_license["SA"],
+                                         nc=current_license["NC"],
+                                         nd=current_license["ND"],
+                                         embedded=current_license["embedded"],
+                                         embedded_example_url=current_license["embedded_example_url"])
+                    j.save(blocking=True)
+
+            else:
+                #if journal's id is not found in the system
+                writer.writerow(["Journal not found: " + sheet.cell(row=r, column=1).value])
```
Steven-Eardley, line 94: we're only ever writing errors, so if everything is successful our mandatory output will be an empty file

amdomanska: That is meant to be an error. It's to indicate that SAGE wants us to update the journal that does not exist in the system. I believe it won't be called but put that here just in case I missed something.

requirements.txt
```diff
@@ -1,13 +0,0 @@
...
-
-# first get esprit, specifically the version we know will work (+ it's
-# not on PyPI)
--e git+https://github.com/CottageLabs/esprit.git@6fc6b24450627077f1d38145e14765f776feba8b#egg=esprit
--e git+https://github.com/CottageLabs/dictdiffer.git@cc86c1ca1a452169b1b2e4a0cb5fc9e6125bc572#egg=dictdiffer
--e git+https://github.com/DOAJ/flask-swagger.git@f1dbf918d9903a588eed3cce2a87eeccc9f8cc0e#egg=flask-swagger
--e git+https://github.com/CottageLabs/combinatrix.git@740d255f0050d53a20324df41c08981499bb292c#egg=combinatrix
-
-# now run the normal setup.py and get all the normal PyPI dependencies from there
--e .
```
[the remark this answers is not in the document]
amdomanska: yes, sorry, I know it should be here, missed it somehow.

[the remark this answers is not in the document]
amdomanska: I cannot see why the difference is shown here now.

dommitchell:
@amdomanska and @Steven-Eardley -  a guesstimate of the time you took here, please?

amdomanska:
About 4 hours on my side.

Steven-Eardley:
And about an hour from me

PR #1755, opened 2020-06-30
add custom API route for JCT in progress queries

richard-jones:
Adds a basic approach to generate data about applications that are in progress with DOAJ.  This is limited to accounts with a particular permission (jct_inprogress), so when this is deployed we will need to set up a user account for JCT and give it that permission.  It only provides essential information on the following statuses:

* in progress
* completed
* ready
* update_request
* revisions_requested

Before we release this just pending confirmation from @dommitchell that the last two statuses can be included, so don't merge just yet.

### DIFF
--- a/portality/view/jct.py
+++ b/portality/view/jct.py
@@ -1,6 +1,7 @@
 from flask import Blueprint, make_response
 from flask_login import current_user, login_required
 from flask import abort
+from portality.decorators import api_key_required
 
 from portality import models as models
 from portality import constants
@@ -19,7 +20,7 @@ INCLUDE_STATUSES = [
 ]
 
 @blueprint.route('/inprogress')
-@login_required
+@api_key_required
 def inprogress():
     if not current_user.has_role("jct_inprogress"):
         abort(404)
--- a/portality/view/jct.py
+++ b/portality/view/jct.py
@@ -1,0 +1,33 @@
+from flask import Blueprint, make_response
+from flask_login import current_user, login_required
+from flask import abort
+
+from portality import models as models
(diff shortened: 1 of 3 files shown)


portality/view/jct.py
```diff
@@ -0,0 +1,34 @@
+from flask import Blueprint, make_response
+from flask_login import current_user, login_required
+from flask import abort
+from portality.decorators import api_key_required
+
+from portality import models as models
```
Steven-Eardley, line 6: Hi Richard, or shall I call you... Richard?

portality/view/jct.py
```diff
@@ -0,0 +1,34 @@
...
+    if not current_user.has_role("jct_inprogress"):
+        abort(404)
+
+    records = []
+    for application in models.Suggestion.list_by_status(INCLUDE_STATUSES):
+        record = JCTInProgressXWalk.application2jct(application)
+        records.append(record)
+
+    data = json.dumps(records)
+    return make_response((data, 200, {'Content-Type': 'application/json'}))
```
Steven-Eardley, line 34: `util.make_json_resp` should be used here

portality/view/jct.py
richard-jones: ha, weird, that's just copied from somewhere else in the code, who knows how it wound up like that originally

richard-jones:
@Steven-Eardley this one is good to merge and deploy, if you could do that when you have a moment.

I checked the test error on codeship, and it is unrelated and works locally, so just a ccodeship performance issue there.

PR #1760, opened 2020-07-24
Feature/af redesign  apc charges field

amdomanska:


portality/templates/application_form/_form_js.html
```diff
@@ -308,8 +309,8 @@
         });
     }
 
-    let currentTab = 0; // Current tab is set to be the first tab (0)
-    let previousTab = 0;
+    let currentTab = 4; // Current tab is set to be the first tab (0)
```
Steven-Eardley, line 312: 4 and 3? why these?

Steven-Eardley: A: I used it for easier development to not switch between the tabs constantly and usually forget to set to (0,0). I will just set it correctly at the end of work. But I remember about it.

PR #1762, opened 2020-07-28
Additional validators: OnlyIf, NotIf, GroupMember

Steven-Eardley:


amdomanska:
Assigning myself to remember about implementing this in the front end.

PR #1767, opened 2020-09-14
Adds support for env vars when setting default service hosts

mauromsl:
Just a nice to have:

Setting the Redis and Elasticsearch host/port settings via environment variables helps with dockerized setups.

The alternative is to set these values via the environment specific `.cfg` files, but we find this a bit cumbersome for our use case (Throw-away environments for a continuous integration pipeline).

### DIFF
--- a/portality/settings.py
+++ b/portality/settings.py
@@ -69,7 +69,7 @@ SSL = True
 VALID_ENVIRONMENTS = ['dev', 'test', 'staging', 'production', 'harvester']
 
 # elasticsearch settings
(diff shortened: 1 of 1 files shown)


Steven-Eardley:
@mauromsl thanks for this, sorry it's taken a while to get back to you.

Are you running the DOAJ app completely without an `app.cfg` instance override file in your CI pipeline? We just create one as an initialisation step since the app needs its secrets anyway. I can see the utility of these env vars even though it's not necessary for us upstream, so I'll merge it in.

Steven-Eardley (approved):

PR #1768, opened 2020-09-15
Detail pages UI

sssoz:
Cleaned up some styles and markup for both the article detail pages & journal table of contents. 

See https://github.com/DOAJ/doajPM/issues/2292

Some other things to do, I think would be better in a separate PR:

Articles: 
- ~make sure to include the Google metatags~ done in this commit of this PR f1a8855812bf646e42062baa6eeb441debfcaa86 
- ~the `<title>` field also isn’t showing~
  - ~both of these are because of template inheritance not working as it used to~
- the secondary header should also have the highlighted class 
- the `{{ bibjson.flattened_archiving_policies|join(", ") }}` variable isn’t working and gives an `TypeError: 'NoneType' object is not iterable` error
  - every single policy should also be in its own list item, not all listed in one 
  - I don’t think I can fix this myself though

Journals:
- the tabs are kind of annoying  -- whenever you click on them, because each tab section is an anchor, it forces the page the scroll back up 
  - this is due to a bit of JS I added, need to look into it 

Closes https://github.com/DOAJ/doajPM/issues/2465


richard-jones:
@sssoz I've been through this and fixed a bunch of the TODOs and stuff that wasn't working, and got it merged down to af_redesign.  You might want to just run over it from the UI and check in the template for the things that I changed.

In particular, I've added handling for multiple APC currencies/prices, and multiple licenses.

PR #1769, opened 2020-09-15
Cleanup search styles & markup

sssoz:
See doaj/DOAJPM#2395 

Other things left to do / I need help with: 

- [x] **Whenever there is no data, do not display the filter** to avoid this:  
Screenshot 2020-09-15 at 23 36 43

---

- [ ] **In the Publisher dashboard, whenever there are no journals / articles in a search UI, do not display the tab for that UI at all** to avoid this:  
Screenshot 2020-09-15 at 23 37 24

---

- [x] **Update the markup for `SortRenderer`** so to reduce unused `<div>`s and make it accessible  

Instead of [the markup here](https://github.com/CottageLabs/edges/blob/0db83e3101ca1a6fd1b105792b2816c83ad9c2d4/src/renderers/bs3.SortRenderer.js#L40), display the following:

```
sortOptions = '' + label + '\
' + direction + ' \
<select name="' + selectName + '" class="form-control input-sm ' + sortFieldClass + '" id="' + selectName + '"> \
<option value="_score">Relevance</option>';

for (var i = 0; i < comp.sortOptions.length; i++) {
  var field = comp.sortOptions[i].field;
  var display = comp.sortOptions[i].display;
  var dir = comp.sortOptions[i].dir;
  if (dir === undefined) {
    dir = "";
  }
  dir = " " + dir;
  sortOptions += '<option value="' + field + '' + dir + '">' + edges.escapeHtml(display) + '</option>';
}

sortOptions += ' </select>';
```


richard-jones:
>Whenever there is no data, do not display the filter to avoid this:

That's a config option on the filter, so we should be able to fix that quickly

> In the Publisher dashboard, whenever there are no journals / articles in a search UI, do not display the tab for that UI at all

That's quite hard, as it involves hitting the index every time we load the publisher navigation.  Otherwise, you can't know that there aren't update requests if the user isn't on the update requests tab.

Perhaps a middle ground here is to offer the user some instructions on how to create an update request?


Also, I notice that a few classes have got dropped from the elements they were attached to - we will probably need to put those back in, as they will be used for javascript event bindings.

sssoz:
> That's a config option on the filter, so we should be able to fix that quickly

Great! Can you show me so I can do it? 

> That's quite hard, as it involves hitting the index every time we load the publisher navigation. Otherwise, you can't know that there aren't update requests if the user isn't on the update requests tab.
>
>Perhaps a middle ground here is to offer the user some instructions on how to create an update request?

Sounds good. Yes, a prompt to create one could work! 


> Also, I notice that a few classes have got dropped from the elements they were attached to - we will probably need to put those back in, as they will be used for javascript event bindings.

Depends which classes. I want to keep anything related to visual styling separate from interactions. We should only be using BEM-styled classes for styling, not binding to JS -- we should use either the id or create a class with the nomenclature `js-...`. BEM elements or modifiers can't exist outside of their block (or component), otherwise, the styling gets messed up. e.g. the class `input-group__input` (element is `input`) can only be nested inside a tag with the class `input-group` (the block).

richard-jones:
> > That's a config option on the filter, so we should be able to fix that quickly
> Great! Can you show me so I can do it?

My mistake - there is a config option for this on the old AND facet, but not on the new OR facet, I'll need to add it.  What behaviour do you want exactly, just for the facet to disappear and leave an empty space?

sssoz:
@richard-jones If the filter doesn’t apply to a certain query & there are no facets available to it, then we shouldn’t display anything. If the query changes and the filter is available, it should be displayed again.

richard-jones:
> Update the markup for SortRenderer so to reduce unused <div>s and make it accessible

I've done this, and pushed it to this branch

richard-jones:
>  Whenever there is no data, do not display the filter to avoid this:

I have now pushed a change to this branch that does this

PR #1770, opened 2020-09-17
Update header & footer content

sssoz:
See all tasks relating to header/footer content in DOAJ/doajPM#2481. 

The markup is fine now, but I’m having an issue figuring out how to get the latest styles: there have been changes done to the SCSS in various branches and `af_redesign__static_content` as well. The ones done on the `af_redesign__static_content` branch have not yet been pulled in to `af_redesign`, so I’m missing a few styles (you can see the footer doesn’t look quite right). 

## This PR:

Screenshot 2020-09-17 at 13 08 15

---

## What we expect to see, as it is on doaj-static.netlify.app / `af_redesign__static_content`

Screenshot 2020-09-17 at 13 13 28

The content is as it should be though.


portality/templates/layouts/base.html
```diff
@@ -71,12 +71,92 @@
...
+      target = target.length ? target : $('[name=' + hash.slice(1) +']');
+      if (target.length) {
+        $('html,body').animate({
+          scrollTop: target.offset().top - headerHeight //offsets for fixed header
+        }, 50, 'linear');
+      }
+    } );
+
+  </script>
+
```
richard-jones, line 160: Just so you know, all this code was moved to `_js_includes.html` to keep it together with all the other base JS.

There were also a few bugfixes to it, so unless this code here has changed, and the changes need porting over to the updated version, I'd probably just drop this from the merge.

sssoz: Hmm it seems that without that JS there, some small features don’t appear though, like the "scroll to top" button: 

1ae2413f11a8070b4a71f9c24c53bbd9

richard-jones: ok, I'll have to diagnose.  In theory whether the code is there or in `_js_includes.html` shouldn't matter, it's just keeping the JS all together in the same file.  Maybe there's a loading order thing that needs looking at.

richard-jones: (incidentally, and maybe or maybe not related, I get a lot of errors from the scroll code on my machine, maybe it's crashing out on some pages and stopping working.  I'll look when I get a moment)

richard-jones: (I say "a lot of errors", it's always the same error, but I get it a lot :) )

sssoz: Ah, just saw that! I get the  error too. JS is definitely not my thing... 😬

PR #1776, opened 2020-10-14
Update static files

sssoz:
This is mostly updates to the Jekyll Markdown files **but** I realised I was still making some changes to the SASS on this branch (I no longer do). I wanted to make sure they were merged to the main codebase too. Sorry I didn’t do this sooner!

Let me know if there’s a better way to do this -- like maybe cherry-pick the stylesheet-related commits to its own branch.


richard-jones:
@sssoz I can merge this one as-is if you like, the only question I had is that I noticed there were changes to the static site templates in `_layout` (e.g. `nosidenav.html`) and I wonder if they need to be incorporated into the jinja templates too (or instead)?

sssoz:
@richard-jones I’m not quite sure how the templates and layouts have been organised in portality, so I’ll have to take a moment to look into it. If we do need to update them, is it okay if I do it on this branch?

richard-jones:
Yes, that's fine, in fact that would make the most sense, as then I can check that everything is consistent.

sssoz:
@richard-jones it’s all yours -- no need to do anything to display the static pages (see `portality/view/doaj.py`), they display the ones that are generated by Jekyll. I did just add a few URLs that weren’t on portality yet.

PR #1778, opened 2020-10-20
Rebuild static assets

sssoz:
@Steven-Eardley I’ve rebuilt the static assets (with jekyll’s `jekyll build` command); from what I see, the Accessibility, Privacy, and T&Cs are in here already.


sssoz:
This works now (there’s no content in the Accessibility page yet but the page is now displayed)

PR #1779, opened 2020-10-23
FEEDBACK - Minor styling & templating stuff

sssoz:
...all from various feedback tasks (each commit is tagged with its corresponding issue). 

This also includes the feedback task to fix the "Thank you" fragment. 

Also took the opportunity to also remove all the pages we no longer need (dynamic pages like the app. form that are now in `portality/`) -- this is why so many lines have been removed. It was unnecessarily recompiling these pages every time.


portality/templates/account/forgot.html
```diff
@@ -1,26 +1,27 @@
...
+                <h1>Reset your password</h1>
+                <p>If you forgot your password, or if your password reset link has expired,
+                    enter your email address in the box and click "reset your password".</p>
+                <p><strong>NOTE: </strong>You must know your correct email, and you must have access to the account associated
+                to the email address you used, as your password reset instructions will be sent there.</p>
+                <p>If you do not know these things, <a href="/contact">contact us directly</a> for further help.</p>
+            </div>
+            <div class="col-md-6">
+                <form method="POST" action="">
+                    <label for="email">Email address</label>
```
Steven-Eardley, line 19: This should include username - I'll change the field name to make more sense too https://github.com/DOAJ/doajPM/issues/2554

portality/templates/api/v2/api_docs.html
```diff
@@ -8,8 +12,6 @@
 
 {% endblock %}
 
-{% extends "layouts/base.html" %}
```
Steven-Eardley, line 11: This breaks the swagger UI stuff, so it's not working on this branch. I agree it should definitely be at the top though.

portality/templates/includes/footer.html
```diff
@@ -96,19 +96,19 @@ <h2>
                 </h2>
                 <ul>
                     <li>
-                        <a href="/apply/">Application form</a>
+                        <a href="">Application form</a>
```
Steven-Eardley, line 99: missing URL (Loving the rest of the `url_for()`!!)

portality/view/doaj.py
```diff
@@ -427,127 +427,127 @@ def google_webmaster_tools():
 
 @blueprint.route("/accessibility/")
 def accessibility():
-    return render_template("layouts/static_page.html", page_frag="/accessibility-fragment.html")
+    return render_template("layouts/static_page.html", page_frag="/accessibility-fragment.html", page_title="Accessibility statement for doaj.org")
```
Steven-Eardley, line 430: I can see that adding these to the templates necessitates passing in the `page_title` - but I don't like that string constants have been added to the code. Perhaps we should have a lookup from config for these page titles.

static_content/_includes/footer.html
```diff
@@ -5,7 +5,7 @@
       <div class="col-xs-6 col-sm-4 col-lg-2">
         <h2>Apply</h2>
         <ul>
-          <li><a href="/apply/">Application form</a></li>
+          <li><a href="#">Application form</a></li>
```
Steven-Eardley, line 8: Not sure where this comes in, but it might be our missing url in the html footer

static_content/_site/apply/thank-you-fragment/index.html
```diff
@@ -1,25 +1,19 @@
...
+          <div class="row">
+            <div class="col-md-8">
+              <p>You will receive an automated email from us, confirming that your application has been received, in the next few minutes. Please <a href="contact/">contact us</a> if you don’t receive it.</p>
 
 <p>You will receive two more status updates if your application is assigned to an editorial group for review. If a Managing editor processes your application, this step is skipped.</p>
 
 <p>Further emails from the reviewer may be sent asking questions about your application.</p>
 <ul>
-  <li>The person reviewing your application may not be using a doaj.org email address. If you are not sure that the person you are writing to is authorised to represent DOAJ, please check the <a href="about/volunteers.md %}">list of volunteers</a>.</li>
+  <li>The person reviewing your application may not be using a doaj.org email address. If you are not sure that the person you are writing to is authorised to represent DOAJ, please check the <a href="about/volunteers.md%20%%7D">list of volunteers</a>.</li>
```
Steven-Eardley, line 16: This link isn't working - comes out as ...`/apply/about/volunteers.md %}`

Steven-Eardley (approved):
Great work, I'll merge but there's a few comments that may require another pass

static_content/_includes/footer.html
sssoz: Oops, accidentally changed this.

sssoz: Nevermind this comment: I manually update the footer/header templates from `portality/`; the footer & header in `static_content` are outdated now! But fixed and pushed :)

portality/templates/api/v2/api_docs.html
sssoz: Hmm  I tried putting it back but  the API content still isn’t being displayed...

Steven-Eardley: I'll take this one on

portality/view/doaj.py
sssoz: I don’t know how to set that up :\

Steven-Eardley: not a problem, we can bring it up in the meeting. `doaj.py` does have almost all of the static page routes, so maybe it's okay to keep them here for now. Otherwise it'd be something like a yaml file with all of the titles.

PR #1787, opened 2020-11-15
Hotfix/2020 11 15 2567 script delete users without journals

Steven-Eardley:
Delete users script and tests for issue https://github.com/DOAJ/doajPM/issues/2567

[edited for enhancements below]

To run non-destructively use the `-r` option and supply a filename for the CSV of accounts that will be deleted.

Run again with `-i` with the same filename to be prompted to delete the user accounts. CSV should still contain header rows but user rows can be deleted from CSV if they are to be retained in the system

### DIFF
--- a/doajtest/unit/test_scripts_accounts_delete_pub_no_journal.py
+++ b/doajtest/unit/test_scripts_accounts_delete_pub_no_journal.py
@@ -1,0 +1,97 @@
+import time
+import esprit
+
+from doajtest.helpers import DoajTestCase
+from doajtest.fixtures.accounts import AccountFixtureFactory
+from doajtest.fixtures.journals import JournalFixtureFactory
+
+from portality import models
+from portality.scripts.accounts_delete_pub_no_journal import accounts_with_no_journals, delete_accounts_by_id
+
+
+class TestScriptsAccountsDeletePubNoJournal(DoajTestCase):
+
+    def setUp(self):
+        super(TestScriptsAccountsDeletePubNoJournal, self).setUp()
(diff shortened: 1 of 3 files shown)


Steven-Eardley:
This will need to be ported to the af_redesign branch before release separately.

Steven-Eardley:
I might've thought of a modification that might improve this - rather than adding on the CSV report as an extra, the CSV report should be fed back in to perform the deletes. That way you could edit the CSV to remove accounts you want to keep.

Quicker and dirtier would be to add an exception list of user IDs if there are any accounts in the list that should be retained.

However, If they're all to be deleted every time, what's here should suffice.

dommitchell:
> rather than adding on the CSV report as an extra, the CSV report should be fed back in to perform the deletes. That way you could edit the CSV to remove accounts you want to keep.

Is this in light of my comment here: https://github.com/DOAJ/doajPM/issues/2567#issuecomment-725302602

Steven-Eardley:
> Is this in light of my comment here: DOAJ/doajPM#2567 (comment)?

Yep. I think I'll go ahead and make that enhancement since we'll be using the script on an ongoing basis - it may as well be customisable. It's about half an hour's work

Steven-Eardley:
CSV read-in enhancement is done, cheers to @DhileasHeywood for the help

Steven-Eardley:
I've added further checks for applications in the system before account deletion. On my system (not up-to-date) that resulted in ~200 accounts been retained: 540 deletes down from 740

richard-jones (approved):

PR #1789, opened 2020-11-16
Feedback style & template updates

sssoz:


portality/templates/application_form/assed_application.html
```diff
@@ -10,32 +10,27 @@
 
 {% block editor_content scoped %}
 <body id="apply">
-    <main class="container">
+    <main class="container page-content">
+        <h1>{{ obj.bibjson().title }}</h1>
```
Steven-Eardley, line 14: Due to the chance of some odd characters in titles it might be worth using the `| safe` filter so they're not escaped by jinja (you should be able find some examples elsewhere)

Steven-Eardley (approved):
Looks good, might be a change needed for un-escaping title text.

portality/templates/application_form/assed_application.html
sssoz: NP! Will have to change it everywhere -- this was already being used in other places, I think.

sssoz: Also, what kind of odd characters? Is there sometimes markup in journal titles?

Steven-Eardley: or maybe escaping it is correct - let's keep it how you've done it

sssoz: No problem! I had started collecting different "edge cases" for journal titles in non-Latin scripts. Perhaps we could collect any other weird cases too where there is potentially  "bad" data? 

https://docs.google.com/document/d/1koBSLp4Zj9ktLeOqwX5o5q9YaJeKubOQXZleK2UDyQU/edit#heading=h.yp6yhhe31rv8

PR #1790, opened 2020-11-18
Display basic diff table

sssoz:
Display basic diff table

See DOAJ/doajPM#2555


richard-jones:
@sssoz I can't merge this automatically as it's reporting conflicts with the af_redesign branch.  Do you want to resolve those?  I might not resolve them in the right way, since it's in the `main.css`.

sssoz:
@richard-jones main.css is just the compiled assets (we don’t modify it directly), so it'll often conflict. It's fine to delete it and recompile with `bundle exec jekyll build`! I did this and pushed it again!

PR #1799, opened 2020-12-02
Add CC definitions to app. form help

sssoz:
And add alphanumeric inputs for article metadata form.

### DIFF
--- a/portality/formcontext/forms.py
+++ b/portality/formcontext/forms.py
@@ -672,8 +672,8 @@ class ArticleForm(Form):
     pissn = DOAJSelectField("Print", [ThisOrThat("eissn", "Either this field or Online ISSN is required")], choices=[]) # choices set at construction
     eissn = DOAJSelectField("Online", [ThisOrThat("pissn", "Either this field or Print ISSN is required")], choices=[]) # choices set at construction
 
-    volume = IntegerField("Volume", [validators.Optional()])
-    number = IntegerField("Issue", [validators.Optional()])
+    volume = StringField("Volume", [validators.Optional()])
+    number = StringField("Issue", [validators.Optional()])
     start = StringField("Start", [validators.Optional()])
     end = StringField("End", [validators.Optional()])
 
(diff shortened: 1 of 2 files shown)


portality/formcontext/forms.py
```diff
@@ -674,8 +674,8 @@ class ArticleForm(Form):
 
     volume = IntegerField("Volume", [validators.Optional()])
     number = IntegerField("Issue", [validators.Optional()])
-    start = IntegerField("Start", [validators.Optional()])
-    end = IntegerField("End", [validators.Optional()])
+    start = StringField("Start", [validators.Optional()])
```
Steven-Eardley, line 677: Good find. Is there an issue open for this? volume and issue are also stored as unicode in the data structure, so they should probably also have the same treatment.

portality/forms/application_forms.py
```diff
@@ -531,7 +531,25 @@ class FieldDefinitions:
         "help": {
             "long_help": ["The journal must use some form of licensing to be considered for indexing in DOAJ. ",
                           "If Creative Commons licensing is not used, then select <em>Publisher's own license</em> and enter "
-                          "more details below."],
+                          "more details below.",
```
Steven-Eardley, line 534: Looks good, worth checking rendering on the test server

Steven-Eardley (changes requested):

portality/formcontext/forms.py
sssoz: https://github.com/DOAJ/doajPM/issues/2654

sssoz: This is how it's shown on the commit: 

Screenshot 2020-12-03 at 17 02 46

Steven-Eardley: great, if you use 'Fixes' rather than 'See' Github will know to close the issue in question on merge https://guides.github.com/features/issues/#notifications

sssoz: Yeah, I just don’t use it here because I figured that Dom might want to check / test them before closing them!

Steven-Eardley: good point, we'll keep the issues open and not use `fixes`. (although Dom had already closed these prematurely as soon as you'd issued the PR!)

portality/forms/application_forms.py
sssoz: (This line adds another paragraph)

Let me know when it’s on test, but I’ve already viewed it locally.

PR #1801, opened 2020-12-04
Hotfix/20201204 1330 issues2642 2641

richard-jones:


Steven-Eardley (approved):
lovely stuff.

dommitchell:
Thanks @Steven-Eardley and @richard-jones If we can get these on test first, please. That would be great!

Steven-Eardley:
@dommitchell on test now

PR #1804, opened 2020-12-08
Feature/2555 urs

richard-jones:
Covers both: https://github.com/DOAJ/doajPM/issues/2555 and https://github.com/DOAJ/doajPM/issues/2609 as two sides of the same coin.


Steven-Eardley (approved):
Looks good

PR #1805, opened 2020-12-10
modify link on upload history to point to new help page and fix quick search field lists

richard-jones:
For convenience I have also put the changes for https://github.com/DOAJ/doajPM/issues/2669 in here


Steven-Eardley (approved):
nifty, looks good

PR #1807, opened 2020-12-14
add monkey patch to flask redirect to handle ssl issue in live

richard-jones:


Steven-Eardley:
Thanks for this, seems the ProxyFix solution is working, so closing.

PR #1808, opened 2020-12-14
fix issues with crosswalk and form definition which were causing vali…

richard-jones:
…dation failures.  Also adjust the bulk add note test to include the failing case as a regression test


Steven-Eardley (approved):
fab

PR #1809, opened 2020-12-14
remove requirement for subject classification when application is in …

richard-jones:
…progress


Steven-Eardley (approved):
yep. So simple, what a beautiful thing.

PR #1811, opened 2020-12-14
Update Sponsors copy

sssoz:


sssoz:
@Steven-Eardley Not sure if I still have to recompile assets. Let me know!

Steven-Eardley:
@sssoz it's correct to not recompile assets, avoids conflicts with the content moving about. To clarify, should this be a hotfix branch (based on master, ready to go live) or should it be a feature branch (based on develop, released in future after other changes that go on test first)?

sssoz:
@Steven-Eardley You’re right, it should be a hotfix. I will prefix those branches with `hotfix/` then? And base them off master

PR #1812, opened 2020-12-15
Update sponsors copy

sssoz:


portality/view/doaj.py
```diff
@@ -447,7 +447,7 @@ def support():
 
 @blueprint.route("/support/sponsors/")
```
Steven-Eardley, line 448: would it be worth changing the route and adding (yet another) legacy redirect? (so long we we've used `url_for('sponsors') everywhere we can change the route without issue)

Steven-Eardley (approved):
Looks good, will generate and release

PR #1818, opened 2020-12-21
Feature/2486 Add back link

sssoz:
See DOAJ/doajPM#2486, comment https://github.com/DOAJ/doajPM/issues/2486#issuecomment-703701032

### DIFF
--- a/portality/static/doaj/css/application_form.css
+++ b/portality/static/doaj/css/application_form.css
@@ -55,10 +55,6 @@ li.parsley-type {
     color: #982E0A;
 }
 
-.cannot-submit {
-    color: #982E0A;
-}
-
 /*.formulaic-clickableurl-visit {*/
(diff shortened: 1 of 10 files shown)


richard-jones:
We need to do a little more here.  As there is already a button on the page with the id `prevBtn` then technically the HTML is invalid, but more importantly it seems to break the event bindings on the other back button.

What we should do is have a `prevBtn` class, which the js events binds instead, and then we can have as many buttons as we want.

@sssoz do you want me to make those changes, or do you want to take a shot?

sssoz:
@richard-jones yes, I was quite tired when I did this! My bad. It looks like there is a lot of IDs used as JS event binders, which isn't great for semantic HTML. I can do it now.

sssoz:
Done and pushed

PR #1819, opened 2020-12-21
Explicitly state when there is no waiver policy

sssoz:
See DOAJ/doajPM#2612


richard-jones:
Although this is tagged as a hotfix, the PR was against develop, so I have just merged it to develop, and it will go out with our next push to live.

PR #1822, opened 2021-01-11
turn off validation on front and beckend if application status is rejected

amdomanska:
https://github.com/DOAJ/doajPM/issues/2683


portality/lib/formulaic.py
```diff
@@ -728,6 +728,7 @@ def __init__(self, formdata=None, source=None, parent: FormulaicContext=None):
         self._alert = []
         self._info = ''
         self._formulaic = parent
+        self._status_not_requiring_validation = ['rejected']
```
richard-jones, line 731: This shouldn't be in the formulaic library file - that ought to be totally independent of the actual form configuration.

portality/lib/formulaic.py
```diff
@@ -910,11 +911,14 @@ def draft(self, *args, **kwargs):
     def validate(self):
         self.pre_validate()
         f = self.form
-        valid = False
         if f is not None:
-            valid = f.validate()
+            if f.application_status.data in self._status_not_requiring_validation:
+                return True
+            else:
+                return f.validate()
```
richard-jones, line 918: This shouldn't be in the formulaic library file.  Instead override the validation method in the application_processor module for the form instances where it is relevant.  In this case, I'd copy the style ManEdJournalReview.validate method to the AdminApplicationReview class, and make it aware that if the form.application_status contains `rejected` then it shouldn't validate.

portality/static/js/application_form.js
```diff
@@ -563,6 +564,11 @@ doaj.af.EditorialApplicationForm = class extends doaj.af.BaseApplicationForm {
         }
         this.form.submit();
     }
+
+    setAllFieldsOptionalIfAppropriate() {
+        return this.statusesNotRequiringValidation.includes(this.jq("#application_status").val());
```
richard-jones, line 569: I think this means you can now no longer explicitly set "all fields optional" with that checkbox.

It's possible now that we have different behaviours for journal-like and application-like forms that it would be sensible to add an extra class layer between EditorialApplicationForm and the specific implementations that handles validation exceptions according to the form.

richard-jones (changes requested):

portality/lib/formulaic.py
amdomanska: > that ought to be totally independent of the actual form configuration.
You're right, it shouldn't be in the formulaic. It should be in the Form class - why independent of the actual form configuration? I would argue this IS the form configuration - that we may want to change in the future (ex. add other states)

portality/lib/formulaic.py
amdomanska: The error is that I do not check whether the `f.application_status` exists at all. I would just add that check and leave it here. If any form gives the user possibility to change the status to "rejected" it should also not require validation in such case. And it might be the case that the scope of responsibility will be changed one day.

amdomanska (commented):
This implementation does not require validation on Admin form for statuses: ['rejected', 'pending', 'in progress', 'on hold'];

portality/static/js/application_form.js
richard-jones: ok, I'd need to test the equivalent journal form directly, which I can do tomorrow, but it looks like the "make_all_fields_optional" checkbox is no longer looked at anywhere, so the ManEd Journal form won't be able to override validation by selecting that checkbox.

richard-jones: I have confirmed this by manual testing.  If you go to the Journal admin form and select the "Allow save without validation" with checkbox, it still validates the form, and does not allow you to save.

richard-jones (changes requested):
Need to add back in support for "make_all_fields_optional" on the Journal form

portality/static/js/application_form.js
amdomanska: Cannot reproduce it. I've saved the journal without title:

image
Confirmed by search:
image
Not saved when checkbox is unchecked:
image

portality/static/js/application_form.js
```diff
@@ -563,6 +564,11 @@ doaj.af.EditorialApplicationForm = class extends doaj.af.BaseApplicationForm {
         }
         this.form.submit();
     }
+
+    setAllFieldsOptionalIfAppropriate() {
+        return (this.statusesNotRequiringValidation.includes(this.jq("#application_status").val()) || this.jq("#make_all_fields_optional").is(":checked"))
```
amdomanska, line 569: check for `make_all_fields_optional` checkbox added

portality/static/js/application_form.js
amdomanska: fixed

richard-jones (changes requested):
This all appears to be working, there are just a few issues with the PR:

* It's got stuff from a Crossref bugfix mixed in with it
* It's got some development files from your IDE, I think - doaj.css, doaj.css.map and main.css, and main.css also has error messages in it

I think the development files can just be removed, but the crossref stuff will need extracting to another branch

amdomanska:
Done. Sorry for that, shouldn't happen.

richard-jones (approved):

portality/forms/application_processors.py
```diff
@@ -447,6 +447,16 @@ def _send_application_approved_email(self, journal_title, publisher_name, email,
             self.add_alert('Sending the journal acceptance information email didn\'t work. Please quote this magic number when reporting the issue: ' + magic + ' . Thank you!')
             app.logger.exception('Error sending application approved email failed - ' + magic)
 
+    def validate(self):
+        _statuses_not_requiring_validation = ['rejected', 'pending', 'in progress', 'on hold']
```
Steven-Eardley, line 451: Usually we'd use `constants.APPLICATION_STATUS_REJECTED` etc. but I don't think we use a js equivalent, this is no worse!

Steven-Eardley (approved):

PR #1825, opened 2021-01-12
A new script to merge duplicate accounts into the one last updated. A…

Steven-Eardley:
…dded a todo for tests of journallikeobject functions

https://github.com/DOAJ/doajPM/issues/2632


Steven-Eardley:
I've taken a bit more time and added a test for this since RJ doesn't have the time to review and it's urgent, I'll dry run first, verify and give it a whirl.

PR #1829, opened 2021-01-16
Hotfix/2723 accept no media type and add human-friendly exception error for no authors

amdomanska:
https://github.com/DOAJ/doajPM/issues/2723
https://github.com/DOAJ/doajPM/issues/2745

`issn.attrib["media_type"] is None` does not work and raises `key_error` if in metadata the ISSN metadata has no "media_type" attribute.

**With no media_type the issn is treated as EISSN by default in this PR**

The Croswalk Exception is raised if no contributors are specified.

image


richard-jones:
There are a couple of logic issues with this which can cause us to lose ISSNs and/or confuse the UI

I made the following truth table to check what happens:

* CrP - crossref data contains a print issn
* CrE - crossref data contains an electronic issn
* P1st - our processing code sees the print issn first (1 for true, 0 to see the E ISSN first)
* Pmt - print issn has media type attribute
* Emt - electronic issn has media type attribute
* Outcome - the result of the algorithm in this PR
* Fix - if necessary how to fix

| CrP | CrE | P1st | Pmt | Emt | Outcome | Fix | 
| ----- | ----- | ------ | ------ | ----- | ------------ | ------- |
| 0 | 0 | 0 | 0 | 0 | no issns| expected behaviour |
| 0 | 1 | 0 | 0 | 0 | CrE stored as EISSN | expected behaviour |
| 0 | 1 | 0 | 0 | 1 | CrE stored as EISSN | expected behaviour |
| 1 | 0 | 0 | 0 | 0 | CrP stored as EISSN | would be better as PISSN, but there is no way to know, so OK |
| 1 | 0 | 0 | 1 | 0 | CrP stored as PISSN | expected behaviour |
| 1 | 1 | 1 | 0 | 0 | CrP and CrE both stored as EISSN | no way to know which is which, but at least put them in separate fields |
| 1 | 1 | 0 | 0 | 0 | CrP and CrE both stored as EISSN | no way to know which is which, but at least put them in separate fields |
| 1 | 1 | 1 | 0 | 1 | CrP and CrE both stored as EISSN | EISSN is identified, so place other in PISSN |
| 1 | 1 | 0 | 0 | 1 | CrP and CrE both stored as EISSN | EISSN is identified, so place other in PISSN |
| 1 | 1 | 1 | 1 | 0 | CrP stored as PISSN, CrE stored as EISSN | expected behaviour |
| 1 | 1 | 0 | 1 | 0 | CrP stored as PISSN, CrE stored as EISSN | expected behaviour |
| 1 | 1 | 1 | 1 | 1 | CrP stored as PISSN, CrE stored as EISSN | expected behaviour |
| 1 | 1 | 0 | 1 | 1 | CrP stored as PISSN, CrE stored as EISSN | expected behaviour |

The result is that we need a bit of extra code to determine what to do when one of the identifiers has a media_type and the other does not.  Not fixing those cases could result in 2 EISSNs being registered, and I'm not sure of the downstream consequences of that.

portality/crosswalks/article_crossref_xml.py
```diff
@@ -235,6 +235,8 @@ def crosswalk_article(self, record, journal):
 
         # authors
         contributors = record.find("x:contributors", NS)
+        if contributors is None:
+            raise CrosswalkException(message="No contributors found.", inner_message="DOAJ requires at least one author for each article")
```
richard-jones, line 239: Could we place these bits of text in the `Messages` file.  Otherwise, that's fine with me

amdomanska:
> The result is that we need a bit of extra code to determine what to do when one of the identifiers has a media_type and the other does not. Not fixing those cases could result in 2 EISSNs being registered, and I'm not sure of the downstream consequences of that.

I would discuss two options:
1. If there are two ISSNs, both without media_type then we can treat the first one as EISSN and the second one as PISSN, if only one has media_type defined then the second one is the one missing
2. If only one has media_type defined then the second one is the one missing but if there are two and both without media_type then we could show the error to avoid guessing.

I suspect this is the question to @dommitchell?

amdomanska:
I added the help to publisher/help page for the new error message. As usual: could you please have a look and fix any wording that needs fixing? Thank you!
image

dommitchell:
@amdomanska thanks for the text changes.  Please can you change the text to the following?

'Even though the Crossref schema accepts XML with no contributors, contributors are required for all articles in DOAJ'

'Please add at least one author to each article that you upload'

amdomanska:
Done.
image

amdomanska:
https://github.com/DOAJ/doajPM/issues/2723#issuecomment-771540166

portality/crosswalks/article_crossref_xml.py
```diff
@@ -171,13 +172,37 @@ def crosswalk_article(self, record, journal):
...
+                        if attrs[1] == "electronic":
+                            attrs[0] = "print"
+                        else:
+                            attrs[0] = "electronic"
+                elif attrs[0] == 0:
+                    attrs[0] = "electronic"
+                    attrs[1] = "print"
+
+                bibjson.add_identifier(bibjson.P_ISSN if attrs[0] == "print" else bibjson.E_ISSN, issns[0].text.upper())
+                bibjson.add_identifier(bibjson.P_ISSN if attrs[1] == "print" else bibjson.E_ISSN, issns[1].text.upper())
```
richard-jones, line 205: A couple more thoughts here about error resilience.  What happens if there are 3 issns (I'm not sure what the schema allows), and what happens if both issns have the same media_type (like, if they both say they are print issns?).

I tried to imagine a bit of code that handled all the cases, and I haven't tested it but something like this might work:

```
unalloc = []
pissns = []
eissns = []

for issn in issns:
    if issn.attrib["media_type"] == 'print':
       pissns.append(issn)
    elif issn.attrib["media_type"] == 'electronic':
       eissns.append(issn)
    else:
        unalloc.append(issn)

if len(pissn) > 1 or len(eissn) > 1: raise Exception("too many issns or issns of the same type")

for issn in unalloc:
    if len(eissn) == 0: 
        eissn.append(issn)
        continue
     if len(pissn) == 0:
        pissn.append(issn)
        continue
      raise Exception("too many issns)

# now eissn has 1 issn, and pissn has 0 or more issns, so they can be added using add_identifier if present
```

richard-jones (changes requested):
Added a note about handling error cases, with some rough code that might do it

portality/crosswalks/article_crossref_xml.py
amdomanska: That's a very good point! I've just checked that and Crossref schema allows max 6(!) ISSNs! But I am wondering why? Are there really cases of journals with 6 valid ISSNs? On the other hand: our schema does not limit the max number of ISSNs at all. If we want to introduce the error like you suggest it is (again) the case of more strict schema then "original" crossref one. I suspect we should also inform the users... and Crossref? Other solution would be changed to our data model and introduction of ISSN arrays...

On the third hand, I do not think we ever had such a problem. Maybe it would be wise to leave it alone for now?

amdomanska: I'm opening a new issue for this. I would suggest deploying this change and solving that problem separately.
https://github.com/DOAJ/doajPM/issues/2806

richard-jones: > On the third hand, I do not think we ever had such a problem. Maybe it would be wise to leave it alone for now?

Before we do that we should have a hard think about what the possible consequences are.  Cleaning up broken article data is hard and time consuming so I'd err on the side of caution.

Let me make sure I've got this right - your code above only works if there are 1 or 2 issns.  If there are more than that, does the article simply fail to import?

amdomanska: I do agree with you it should be handled somehow. I just think we could make it in the separate issue to not delay the deployment of this issue. Especially that it does not "break" the too-many-issns-handling, it just does not fix it (situation is the same as before this issue has been raised). I can pick that up as soon as we've got a decision from Dom - implementation shouldn't be difficult,

>Let me make sure I've got this right - your code above only works if there are 1 or 2 ISSNs. If there are more than that, does the article simply fail to import?

I explained that in the created issue and I added the additional XML upload file. If there is more then 2 issns the last two are taken into consideration.

amdomanska:
@richard-jones and I made the decision to add the exception for more than 2 issns situation to avoid any problems before the decision will be made about what to do in such a situation. It will be changed later if needed.

portality/crosswalks/article_crossref_xml.py
```diff
@@ -171,13 +172,44 @@ def crosswalk_article(self, record, journal):
...
-            if issns is not None:
-                for issn in issns:
-                    if issn.attrib["media_type"] is None or issn.attrib["media_type"] == 'print':
-                        bibjson.add_identifier(bibjson.P_ISSN, issn.text.upper())
-                    elif issn.attrib["media_type"] == 'electronic':
-                        bibjson.add_identifier(bibjson.E_ISSN, issn.text.upper())
 
+            # if more than 2 issns raise the exception
+            if len(issns) > 2:
+                raise CrosswalkException(message="Too many ISSNs. Only 2 ISSNs are allowed")
```
richard-jones, line 178: Let's put the text itself in the Messages file, and then I think this is good to release

richard-jones (changes requested):
One minor change about using the messages file for the exception text, then this is ready to go

portality/crosswalks/article_crossref_xml.py
dommitchell: Journals can't have 6 ISSNs! That's madness. There are only three types of ISSN: P-ISSN, E-ISSN and ISSN-L.

We haven't adopted ISSN-L because its uptake was low.

richard-jones (approved):

Steven-Eardley (approved):
looks good, apart from CSS being added again, I'll clean that out on merge

PR #1830, opened 2021-01-16
activate parsley on form construction and fix Next button behaviour on select2 fields

amdomanska:
https://github.com/DOAJ/doajPM/issues/2626

The issue was the result of parsley not being active till the next button is first clicked. Activating parsley in constructor should prevent the issue. Requires more testing but I cannot reproduce the problem anymore.

For some reason button NEXT stays active after first click on the first tab. It does not affect its functionality and most probably is separate issue.


richard-jones:
It's worth putting this one on the test server and asking the DOAJ team to run some functional tests.  Just in case having Parsley activated all the time on the page causes some other side-effects that we haven't thought of

richard-jones (approved):
Don't forget to remove the console.log statement in formulaic.js before merge

portality/static/js/application_form.js
```diff
@@ -380,6 +374,11 @@ doaj.af.TabbedApplicationForm = class extends doaj.af.BaseApplicationForm {
         }).fail(() => {
             // $("#validated-" + this.currentTab).val("False");
             this.tabValidationState[this.currentTab].state = "invalid";
+            let that = this;
+            let errFields = $(this.parsley.fields).filter(function(){return (this.validationResult !== true && this.domOptions.group === "block-" + that.currentTab)});
+            let errFirst = $(errFields.first()[0].element);
+            errFirst.trigger("focus");
+            errFirst.trigger("focus");
```
richard-jones, line 381: I think it would be worth adding a bit of detail in a comment as to what this is fixing and why, since it's a bit weird.

portality/static/js/formulaic.js
```diff
@@ -1930,6 +1931,8 @@ var formulaic = {
 
                 let selector = "[name='" + this.fieldDef.name + "']";
 
+                $(selector).on("focus", formulaic.widgets._select2_shift_focus);
```
richard-jones, line 1934: We will need to do all the other select2-based widgets too: `Select`, `TagEntry`, `Autocomplete`

richard-jones (changes requested):

portality/static/js/application_form.js
richard-jones: I did a bit more reading around this, and this might work with one hit: (from https://api.jquery.com/focus/):

`errFirst.triggerHandler("focus")`

portality/static/js/formulaic.js
amdomanska: They are added to the code, just for some reason not shown in this diff. Super weird:

Autocomplete: https://github.com/DOAJ/doaj/blob/0580a5541ea039da5960b0b84dd1cbb0b68b09fe/portality/static/js/formulaic.js#L1934
Select: https://github.com/DOAJ/doaj/blob/0580a5541ea039da5960b0b84dd1cbb0b68b09fe/portality/static/js/formulaic.js#L1736
Taglist: https://github.com/DOAJ/doaj/blob/0580a5541ea039da5960b0b84dd1cbb0b68b09fe/portality/static/js/formulaic.js#L1802
Tag Entry: https://github.com/DOAJ/doaj/blob/0580a5541ea039da5960b0b84dd1cbb0b68b09fe/portality/static/js/formulaic.js#L1827

richard-jones (approved):
I've previously reviewed this, and I think now it's ready to go.

Before it is merged I would still like to get the functional tests run on the test server for the application form in case parsley being always active has any other side-effects

amdomanska:
@Steven-Eardley 

I merged the latest develop into this. Definitely requires good (functional?) testing on the test server as lots of conflicts in important areas were resolved.

PR #1832, opened 2021-01-18
check for shared ancestors when removing child classifications

richard-jones:
For https://github.com/DOAJ/doajPM/issues/2650


Steven-Eardley (approved):
🐜

Steven-Eardley:
How can we write bug free software when we're dealing with aunts

PR #1834, opened 2021-01-19
Fix for https://github.com/DOAJ/doajPM/issues/2722 and remove obsolet…

Steven-Eardley:
…e code

https://github.com/DOAJ/doajPM/issues/2722


amdomanska (approved):

richard-jones (approved):
So we completely overlooked the lock pages, oops!  Good catch, thanks.

PR #1835, opened 2021-01-19
Add static thank you for support page

sssoz:
See DOAJ/doajPM#2725


Steven-Eardley (changes requested):
This new route isn't referred to in any code, rather `apply.application_thanks()` is used.

dommitchell:
After you have submitted payment via Paypal buttons, you are referred to this page from the Paypal app @Steven-Eardley

PR #1840, opened 2021-02-01
Hotfix/2708 update journal by csv

Steven-Eardley:
This is the script I ran for issue #2708, I ran it in this branch on the second checked-out DOAJ on the background server - the one used for release - so that I didn't need to worry about the running instance.

This PR has been opened for review since there are changes to the xwalks and processors, and to bring the script into the codebase.

### DIFF
--- a/portality/forms/application_processors.py
+++ b/portality/forms/application_processors.py
@@ -9,7 +9,7 @@ from portality.ui.messages import Messages
 
 from portality.crosswalks.application_form import ApplicationFormXWalk
 from portality.crosswalks.journal_form import JournalFormXWalk
-from flask import url_for, request
+from flask import url_for, request, has_request_context
 from flask_login import current_user
 
 from wtforms import FormField, FieldList
@@ -282,13 +282,14 @@ class AdminApplication(ApplicationProcessor):
 
         # if this application is being accepted, then do the conversion to a journal
         if self.target.application_status == constants.APPLICATION_STATUS_ACCEPTED:
-            j = applicationService.accept_application(self.target, current_user._get_current_object())
+            j = applicationService.accept_application(self.target, account)
             # record the url the journal is available at in the admin are and alert the user
-            jurl = url_for("doaj.toc", identifier=j.toc_id)
-            if self.source.current_journal is not None:  # todo: are alerts displayed?
-                self.add_alert('<a href="{url}" target="_blank">Existing journal updated</a>.'.format(url=jurl))
-            else:
-                self.add_alert('<a href="{url}" target="_blank">New journal created</a>.'.format(url=jurl))
+            if has_request_context():
+                jurl = url_for("doaj.toc", identifier=j.toc_id)
+                if self.source.current_journal is not None:  # todo: are alerts displayed?
+                    self.add_alert('<a href="{url}" target="_blank">Existing journal updated</a>.'.format(url=jurl))
+                else:
+                    self.add_alert('<a href="{url}" target="_blank">New journal created</a>.'.format(url=jurl))
 
             # Add the journal to the account and send the notification email
             try:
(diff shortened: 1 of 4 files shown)


portality/forms/application_processors.py
```diff
@@ -282,13 +282,14 @@ def finalise(self, account, save_target=True, email_alert=True):
...
         if self.target.application_status == constants.APPLICATION_STATUS_ACCEPTED:
-            j = applicationService.accept_application(self.target, current_user._get_current_object())
+            j = applicationService.accept_application(self.target, account)
             # record the url the journal is available at in the admin are and alert the user
-            jurl = url_for("doaj.toc", identifier=j.toc_id)
-            if self.source.current_journal is not None:  # todo: are alerts displayed?
-                self.add_alert('<a href="{url}" target="_blank">Existing journal updated</a>.'.format(url=jurl))
-            else:
-                self.add_alert('<a href="{url}" target="_blank">New journal created</a>.'.format(url=jurl))
+            if has_request_context():
```
richard-jones, line 287: I'm very conflicted by this.  I see why you need it, but it also makes me think that the abstraction is wrong somewhere - our core processing code shouldn't require a request context to run.  Don't know what to do about it, probably just have to keep this as-is for the moment, but a note that we should think again.

portality/forms/application_processors.py
```diff
@@ -300,7 +301,8 @@ def finalise(self, account, save_target=True, email_alert=True):
                 owner.save()
 
                 # for all acceptances, send an email to the owner of the journal
-                self._send_application_approved_email(j.bibjson().title, owner.name, owner.email, self.source.current_journal is not None)
+                if has_request_context():
+                    self._send_application_approved_email(j.bibjson().title, owner.name, owner.email, self.source.current_journal is not None)
```
richard-jones, line 305: I have a more concrete objection to this one - we're defining the behaviour of the code according to its request context, even though the request context has no particular bearing on the execution of the code it protects.  If the code would fail in the absence of a request context, fair enough (notwithstanding comment above about our abstraction here).  But this code, I think, would succeed, so what we really are saying is that there are circumstances where during acceptance we don't want to send the approved email, and in which case that should be a switch that's passed in to the function, not assumed based on call context.

richard-jones (changes requested):
I haven't reviewed the "new" code as that seems to be working well.  All I've done is reviewed and commented on the bits that intersect with the application processors.  I think change is minor, but important for code health

portality/forms/application_processors.py
Steven-Eardley: I agree with you there, I thought the same thing when I was running it and it failed due to the alerts. This is where a notification processor would help, so that context doesn't interfere with the logic laid out in the processor.

portality/forms/application_processors.py
Steven-Eardley: Agreed, I was working around the fact that we'd copied across from formcontext here, so wasn't going into signatures. However it appears there's an `email_alert=True` flag on the function already which I'd failed to notice. I might use that, depending on prior use.

portality/forms/application_processors.py
richard-jones: I think that's the right answer - it's separating the notification code from the form processing code that's required.  I would also like to make it, if possible, that the form processors are "below" the BLL level, so that they are independent of call context; I think that would make a variety of things cleaner and easier, so maybe that's something we can work towards.  Then the notifications system can spin up a test context if it needs to generate urls, etc.

richard-jones: So, what I mean is, lets not change this now, but know how to change it in the future

portality/forms/application_processors.py
richard-jones: Yes, that's better, if we can use that flag then this is good to go

PR #1841, opened 2021-02-02
Feature/2703 reorganise form

richard-jones:
Reorganises the admin forms for journal and applications.  Has been reviewed for functionality, so ready to merge . Related issue https://github.com/DOAJ/doajPM/issues/2703


amdomanska (approved):
ok for me.

amdomanska:
Just noticed - probably would be good to remove the `doaj.css` and `doaj.css.map` files from that PR.

richard-jones:
Yes, good point about doaj.css and doaj.css.map.  I believe these have, at some point, made their way into `develop`, so I have explicitly removed them here, so perhaps that will get rid of them in develop.  I've also added them to `.gitignore` in this PR

portality/templates/application_form/_field.html
```diff
@@ -37,7 +37,7 @@
     {% endif %}
 
     {% if form_diff %}
-        {% if f.name in form_diff %}
+        {% if f.help("update_requests_diff", True) and f.name in form_diff %}
```
Steven-Eardley, line 40: I don't really like how this looks but I know it works.

portality/templates/application_form/editor_application.html
```diff
@@ -7,32 +7,13 @@
 {% set js_validation = True %}
 {% set auto_save = 0 %}
 {% set notabs = true %}
+{% set diff_table = false %}
```
Steven-Eardley, line 10: This file only only differs from the associate form by the form_id - can we further deduplicate these files by handling these variables in the view by passing in a config object?)

Steven-Eardley (approved):

Steven-Eardley:
Merging to develop and deploying to test server for release next week.

PR #1845, opened 2021-02-17
Tag flash notifs as <aside>s instead of <section>s

sssoz:
Better semantically, also reduces spacing

DOAJ/doajPM#2685

### DIFF
--- a/portality/templates/includes/header.html
+++ b/portality/templates/includes/header.html
@@ -327,7 +327,7 @@
                   {% if category == "error" %}
                       {% set category = "danger" %}
                   {% endif %}
-                  <section class="container flash_container">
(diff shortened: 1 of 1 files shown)


richard-jones (approved):

richard-jones:
Just one aside, not related to the issue - @sssoz could you avoid using the auto-close features, as these close the issues in the project board, but usually merging the code doesn't mean that the issue is finished.

sssoz:
@richard-jones No problem! I usually use the "See #..." keyword to reference issues, but for some reason, it made sense to me last night to auto-close it because it was a hotfix / minor fix. I’ll make sure to not use it now though!

PR #1847, opened 2021-02-18
Feature/2718 urls with underscores

richard-jones:
This replaces the default parsley type validator with one customised for our usage by replacing the url regular expression with the same one that we use in the back end

https://github.com/DOAJ/doajPM/issues/2718


portality/static/js/application_form.js
```diff
@@ -866,3 +866,80 @@ window.Parsley.addValidator("notIf", {
...
+  digits: /^\d+$/,
+
+  alphanum: /^\w+$/i,
+
+  date: {
+    test: value => Utils.parse.date(value) !== null
+  },
+
+  url: new RegExp(
+      "^https?://([^/:]+\.[a-z]{2,63}|([0-9]{1,3}\.){3}[0-9]{1,3})(:[0-9]+)?(\/.*)?$"
```
Steven-Eardley, line 895:
```suggestion
      url: /^https?:\/\/([^/:]+\.[a-z]{2,63}|([0-9]{1,3}\.){3}[0-9]{1,3})(:[0-9]+)?(\/.*)?$/
```
Since this regex isn't changing, I think it's preferred to use the literal notation like those above

Steven-Eardley (changes requested):

portality/static/js/application_form.js
richard-jones: because of the extra complexity that escaping the slashes adds to this, I'm going to disagree with you on this one - while the style might be preferred, the readability is better this way round.

PR #1848, opened 2021-02-18
Fix up Paypal form to match new pricing

sssoz:
Hotfix because it’s already in the page content
and the current form is now broken!

See DOAJ/doajPM#2759


richard-jones (approved):
@Steven-Eardley another hotfix for you.  I have reviewed, just needs a merge

PR #1849, opened 2021-02-18
modify the API so that GET on a journal does not require the api key,…

richard-jones:
… and that if the journal is not public it is not served, even if the requester is the owner

This is an updated version of https://github.com/DOAJ/doaj/pull/1717 for the v2 of the API

Once this has been merged, that branch can also be deleted


Steven-Eardley (approved):
looks good

PR #1850, opened 2021-02-18
Improve error messages

sssoz:
See DOAJ/doajPM#2822


richard-jones (approved):
@sssoz I made a couple of minor changes to application_forms.py, in case you want to check them.  Otherwise this is good to go.  Do you want it to go onto the test server for review before we deploy?

sssoz:
Looks good @richard-jones (d’oh on that markup error haha)

Steven-Eardley (approved):

PR #1856, opened 2021-02-19
Fix other charges / waiver policy links

sssoz:
See DOAJ/doajPM#2707


richard-jones (approved):
@Steven-Eardley another hotfix for you, reviewed as good to go

PR #1857, opened 2021-02-23
Fix for failed test

amdomanska:
The crossref tests after the media_type issue merged had race issue. It should be fixed here.


richard-jones (approved):
Without having played around with the race condition issue myself, it's difficult to give a good review of this.  If the race condition seems to have gone, and this only affected the test, this is all good by me.

Note that our doaj.css and doaj.css.map files are in again here

PR #1863, opened 2021-03-02
Update Fixed Query Widget to Edges, both widgets styles to new website style and docs

amdomanska:
localhost:5004/widget_demo is a demo site on our server - the endpoint should be removed before the deployment,
test_site.html is an "external" page with widget embedded. 

for test/prod check TODO comments in fqw_demo.html

Static sites must be regenerated

After inital review requires more tests on the test server.


portality/static/widget/fixed_query.js
```diff
@@ -1,6 +1,7 @@
 (function() {
     // Load the widget's js scripts then HTML content. We assume jQuery is available.
     jQuery(document).ready(function($) {
+        console.log("this is js");
```
richard-jones, line 4: A note to remove console.log statements too

portality/static/widget/fixed_query_body_dev_edges.html
```diff
@@ -0,0 +1,68 @@
...
+        }
+        if (QUERY_OPTIONS.hasOwnProperty('sort_direction')) {
+            widget_fv_opts['sortFilter']['sort_direction'] = QUERY_OPTIONS.sort_direction;
+        }
+
+        if (QUERY_OPTIONS.hasOwnProperty('search_type')) {
+            widget_fv_opts['es_type'] = QUERY_OPTIONS.search_type;
+        }
+
+    }
```
richard-jones, line 64: As much as possible of this should be in the widget js file, rather than embedded in the html file

portality/lib/query_filters.py
```diff
@@ -13,13 +13,12 @@ def public_query_validator(q):
...
-    # queries with zero results returned
-    if q.has_facets():
-        return True
-    else:
+    # we do not expect the fqw
+
+    if not q.has_facets():
         return q.size() == 0
+    else:
+        return True
```
richard-jones, line 21: I don't understand this change - it looks like you've just reversed the if/else, which has the same overall effect?  If so, can we keep the original to avoid unnecessary file changes and to keep the original comment?  Or am I missing something?

portality/settings.py
```diff
@@ -475,7 +475,7 @@
         "journal,article" : {
             "auth" : False,
             "role" : None,
-            "query_validator" : "public_query_validator",
+            # "query_validator" : "public_query_validator",
```
richard-jones, line 478: Relating to my above comment, since you are not using the public_query_validator here you shouldn't need to update the `public_query_validator` function.

portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -2081,7 +2093,8 @@ $.extend(true, doaj, {
                 }
 
                 this.component.context.html(frag);
-                feather.replace();
+                //TODO: bring it back for non-widget usage!
+                //feather.replace();
```
richard-jones, line 2097: Just a reminder note to do this :)

portality/static/js/doaj.js
```diff
@@ -1,165 +1,165 @@
 /** base namespace for all DOAJ-specific functions */
 var doaj = {
-    init : function() {
```
richard-jones, line 3: Is this all commented out as part of your workflow, and that it's all coming back in again when you are done?

portality/static/widget/fixed_query_body_dev_edges.html
```diff
@@ -0,0 +1,68 @@
+<script type="text/javascript" src="/static/vendor/jquery-3.4.1/jquery-3.4.1.min.js"></script>
```
richard-jones, line 1: Just a reminder for the cleanup: we should either put this code back in the original `fixed_query_body_dev.html` file, or remove the old files that we no longer need.

portality/static/widget/fixed_query_body_dev_edges.html
richard-jones: I think this is the code that we were discussing how to map from the options to the ES query.  Can we get rid of this whole conversion to the intermediate `widget_fv_opts` object, and just generate a query directly that we pass into the init function.  Since all these options are in a global `QUERY_OPTIONS` object, I don't think we need the `widget_fv_options` object at all, the widget init can just make the direct coversion to an ES query object.

portality/static/widget/fixed_query_dev.js
```diff
@@ -1,12 +1,56 @@
...
-    // Load the widget's js scripts then HTML content. We assume jQuery is available.
+// Localize jQuery variable
+let jQuery;
+
+/******** Load jQuery if not present *********/
+if (window.jQuery === undefined || window.jQuery.fn.jquery !== "1.9.1") {
+    let script_tag = document.createElement("script");
+    script_tag.setAttribute("type","text/javascript");
+    script_tag.setAttribute("src",
+        "http://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js");
```
richard-jones, line 11: Can we serve our own static version of jquery here so we remain consistent

portality/static/widget/fixed_query_dev.js
```diff
@@ -1,12 +1,56 @@
...
+    jQuery = window.jQuery.noConflict(true);
+    // Call our main function
+    main();
+}
+
+/******** Our main function ********/
+function main() {
     jQuery(document).ready(function($) {
-        $('#doaj-fixed-query-widget').load(doaj_url + "/static/widget/fixed_query_body_dev.html");
+        let url = doaj_url + "/static/widget/fixed_query_body_dev_edges.html";
```
richard-jones, line 41: It would be lovely to find a DRYer way of doing this, so we don't have to have different versions of the files for different environments.  I don't know if that's possible - we obviously didn't find a way of doing it before, but perhaps we can pull all this stuff out at least to an environment-localisation file like `widget-dev.js` which contains all the dev-specific urls

portality/static/widget/fixed_query_src_edges.js
```diff
@@ -0,0 +1,68 @@
...
+                        alert("There was an unexpected error.  Please reload the page and try again.  If the issue persists please contact us.");
+                    }
+                }
+            });
+            doaj.publicSearch.activeEdges[selector] = e;
+        }
+    }
+
+});
+    doaj.publicSearch.init(widget_fv_opts);
```
richard-jones, line 66: Yeah, so here, I don't think you need `widget_fv_opts` since `QUERY_SETTINGS` is present in global scope

richard-jones: Or you could pass `QUERY_SETTINGS` in directly to make the function independent of global scope, even if the overall module isn't (which is essentially what passing `widget_fv_opts` is doing anyway

portality/static/widget/fixed_query_src_edges.js
```diff
@@ -0,0 +1,68 @@
...
+                }
+            }
+            return query;
+        },
+
+        init : function(params) {
+            if (!params) { params = {} }
+
+            let current_domain = document.location.host;
+            let current_scheme = window.location.protocol;
```
richard-jones, line 30: I don't think this is going to work, is it?  Won't this create a URL that's localised to the user's URL rather than the doaj url?

portality/static/widget/fqw_js.js
```diff
@@ -0,0 +1,7404 @@
+/** @namespace */
```
richard-jones, line 1: I couldn't see anywhere where this file was autogenerated, so I'm guessing this is manually created?  Since it's basically copies of all the core dependencies, we should have this compiled from those dependencies.  There's a script that does this already for the old version which you can modify, it's just a few lines of bash that cats all the files together, see `fixed_query_build.sh`

portality/templates/doaj/widgets.html
```diff
@@ -1,138 +1,84 @@
...
+        if (QUERY_OPTIONS.hasOwnProperty('sort_field')) {
+            widget_fv_opts['sortFilter']['field'] = QUERY_OPTIONS.sort_field;
+        }
+        if (QUERY_OPTIONS.hasOwnProperty('sort_direction')) {
+            widget_fv_opts['sortFilter']['sort_direction'] = QUERY_OPTIONS.sort_direction;
+        }
+
+        if (QUERY_OPTIONS.hasOwnProperty('search_type')) {
+            widget_fv_opts['es_type'] = QUERY_OPTIONS.search_type;
+        }
```
richard-jones, line 78: We should be able to get rid of this duplicated code from this file if we get rid of `widget_fv_opts` completely

richard-jones (changes requested):
First major pass on the code review.  I have some front-end testing that I will also do, which is:

1. Look at the widget demo
2. Look at the public search to make sure changes haven't affected it
3. Attempt to create my own widget by reading the docs page
4. Check that the "share" feature of the public search works correctly

portality/static/widget/fixed_query_src_edges.js
```diff
@@ -0,0 +1,68 @@
...
+                })
+            ];
+            let e = edges.newEdge({
+                selector: selector,
+                template: doaj.templates.newPublicSearch({
+                    titleBar: false,
+                    resultsOnly: true
+                }),
+                search_url: search_url,
+                manageUrl : true,
```
richard-jones, line 52: For the widget we should set `manageUrl: false` otherwise it will write stuff into the host system's URL bar

portality/static/widget/fixed_query_src_edges.js
```diff
@@ -0,0 +1,68 @@
...
+
+
+            let components = [
+
+                // results display
+                edges.newResultsDisplay({
+                    id: "results",
+                    category: "results",
+                    renderer : doaj.renderers.newPublicSearchResultRenderer()
+                })
```
richard-jones, line 43: There's no pager component, like there was in the original fixed query widget.  You should be able to copy one from the admin edges that would be suitable

portality/view/doaj.py
```diff
@@ -58,13 +58,21 @@ def news():
     news = blog.News.latest(app.config.get("NEWS_PAGE_NEWS_ITEMS", 20))
     return render_template('doaj/news.html', news=news, blog_url=app.config.get("BLOG_URL"))
 
+@blueprint.route("/widgets_demo")
```
richard-jones, line 61: I wasn't able to get this to render, I got CORS errors instead:

Screenshot from 2021-03-05 11-09-12

portality/view/doaj.py
```diff
@@ -58,13 +58,21 @@ def news():
...
+                           lcc_tree=lcc_jstree
+                           )
 
-# @blueprint.route("/widgets")
-# def widgets():
-#     return render_template('doaj/widgets.html',
-#                            env=app.config.get("DOAJENV"),
-#                            widget_filename_suffix='' if app.config.get('DOAJENV') == 'production' else '_' + app.config.get('DOAJENV', '')
-#                            )
+@blueprint.route("/widgets_demo_empty")
```
richard-jones, line 69: This worked ok, but there were a couple of issues:

1. I got a JS error on load which didn't seem to block the page, but we should figure it out

Screenshot from 2021-03-05 11-04-32


2. The page overflowed the window giving me a horizontal scroll.  This might be resolved when the styles of this are looked at.

richard-jones:
From what I can tell, I don't think the build is quite ready for testing an actual embed.  I copied the share/embed frag from the search interface, and with a bit of tweaking came up with this:

```
<!DOCTYPE html>
<html dir="ltr" lang="en">
    !window.jQuery && document.write(
        '<scr' + 'ipt type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></scr' + 'ipt>'
    );
</script>
    var doaj_url="https://doaj.org";
    var SEARCH_CONFIGURED_OPTIONS={"query":{"filtered":{"filter":{"bool":{"must":[{"terms":{"index.schema_codes_tree.exact":["LCC:S"]}}]}},"query":{"match_all":{}}}},"size":50,"sort":[{"created_date":{"order":"desc"}}]}
</script>
<script src="http://localhost:5004/static/widget/fqw_js.js" type="text/javascript"></script>
<div id="doaj-fixed-query-widget"></div>
</html>
```

When this loads it complains that `doaj` does not exist when a bit of code in `fqw_js.js` tries to extend that object.  I think this is probably waiting on the stuff you said you needed to do with the doaj library

portality/lib/query_filters.py
amdomanska: That is an error. Should be 
```
if q.has_facets()
  return q.size() == 0
else:
  return True
```
as the fqw should not have any facets.

portality/static/widget/fixed_query_body_dev_edges.html
amdomanska: Sure, it will be done. Kept so far as a reference.

portality/templates/doaj/widgets.html
amdomanska: Yes, as mentioned in my inital comment this is only to make debugging easier.

portality/view/doaj.py
amdomanska: Interesting! Works well for me. I will look into that.

portality/static/widget/fqw_js.js
amdomanska: It is not exactly the copy of all the core dependencies as it is much smaller - I have removed all the methods from the core that are not used by the widget. For doaj.js for example it is quite a lot.

amdomanska:
> From what I can tell, I don't think the build is quite ready for testing an actual embed. I copied the share/embed frag from the search interface, and with a bit of tweaking came up with this:
> 
> ```
> <!DOCTYPE html>
> <html dir="ltr" lang="en">
> <script type="text/javascript">
>     !window.jQuery && document.write(
>         '<scr' + 'ipt type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></scr' + 'ipt>'
>     );
> </script>
> <script type="text/javascript">
>     var doaj_url="https://doaj.org";
>     var SEARCH_CONFIGURED_OPTIONS={"query":{"filtered":{"filter":{"bool":{"must":[{"terms":{"index.schema_codes_tree.exact":["LCC:S"]}}]}},"query":{"match_all":{}}}},"size":50,"sort":[{"created_date":{"order":"desc"}}]}
> </script>
> <script src="http://localhost:5004/static/widget/fqw_js.js" type="text/javascript"></script>
> <div id="doaj-fixed-query-widget"></div>
> </html>
> ```
> 
> When this loads it complains that `doaj` does not exist when a bit of code in `fqw_js.js` tries to extend that object. I think this is probably waiting on the stuff you said you needed to do with the doaj library

It might be also because you used `doaj_url="https://doaj.org"` while that is not there yeat and because the script source should not be `fqw_js.js` but  `fixed_query_dev.js ` as before (it is going to be kept to ensure the back-compatibility)

portality/static/widget/fqw_js.js
richard-jones: There's a size vs maintenance trade-off here, and I would come down on the side of ease of maintenance.  Otherwise any change that we make to `doaj.js` we need to go and check if that change is relevant to the fqw and make the change there.  Otherwise, we just introduce a build step which recompiles fqw on release and we no longer have to worry about it.

amdomanska: You are right, this is quite a burden in terms of maintenance. But we need to remember that this is the js that clients machine needs to upload. I do understand your point of view but I am not fully convinced that simply compiling everything is the way to go. Cause why not just import all for js files?

Let me think about it.

portality/lib/query_filters.py
richard-jones: I'm still a bit confused by this change.  In the old logic it rejected a query if it did not have facets but did have a non-zero size parameter.  In the new logic it rejects a query if it has facets and has a non-zero size parameter.

The old logic would work against the fix query widget, but doesn't the new logic work against the public search?

I think what we actually want now is:

```
if q.has_facets():
    return q.size() > 0
else:
    return True
```

Which will just reject all queries which have a zero size and facets.

portality/static/widget/fixed_query.js
```diff
@@ -1,9 +1,11 @@
 (function() {
     // Load the widget's js scripts then HTML content. We assume jQuery is available.
     jQuery(document).ready(function($) {
-        $('#doaj-fixed-query-widget').load(doaj_url + "/static/widget/fixed_query_body.html");
+        $('body').load(doaj_url + "/static/widget/fixed_query_body.html");
```
richard-jones, line 4: What's the reason for this change?  It worked before, and I'm a bit wary of injecting dom modifications outside the div that we've asked embedders to create.

portality/static/widget/fixed_query_src_edges.js
```diff
@@ -0,0 +1,127 @@
...
+                        widget_fv_opts.sortFilter.sort_direction = QUERY_OPTIONS.sort_direction;
+                    }
+
+                    if (QUERY_OPTIONS.hasOwnProperty('search_type')) {
+                        widget_fv_opts.es_type = QUERY_OPTIONS.search_type;
+                    }
+
+                }
+
+                return widget_fv_opts;
```
richard-jones, line 73: We talked about getting rid of this intermediate stage, is this just waiting for cleanup?

portality/static/widget/fqw_js.js
richard-jones: What was your final position on this in the end?  Does the updated build script above mean that this is now being compiled from the other files?

portality/static/widget/public_search_config.js
```diff
@@ -0,0 +1,6 @@
+$.extend(true, doaj, {
+    publicSearchConfig : {
+        publicSearchPath : 'http://localhost:5004/query/journal,article/_search?ref=fqw',
```
richard-jones, line 3: Is this just for developers, or do we need to update this before finalising for release?

portality/view/doaj.py
richard-jones: My fault, developer config to blame!

portality/view/doaj.py
```diff
@@ -58,15 +58,15 @@ def news():
     news = blog.News.latest(app.config.get("NEWS_PAGE_NEWS_ITEMS", 20))
     return render_template('doaj/news.html', news=news, blog_url=app.config.get("BLOG_URL"))
 
+@blueprint.route("/widgets_demo")
+def widgets_demo():
+    return render_template('doaj/fqw_demo.html',
```
richard-jones, line 63: This page does include some of the main search template on the right, which should be hidden as there aren't any facets.

Screenshot from 2021-03-15 14-39-18

richard-jones: Styles for the pager are also a bit wonky, but that might not be worth looking at until Sophy has done a pass over it.

static_content/css/main.css
```diff
@@ -0,0 +1,39 @@
+/*
```
richard-jones, line 1: Unnecessary file for the PR (which also contains a sass error by the looks of it)

richard-jones (changes requested):
Looking good, only a few odds and ends to look at I think.

One query about all the different "versions" of the widget (dev, test, prod) - all the files for those have been deleted.  Have you found a way so that we don't need those files any more or do they need to be reintroduced once everything is ready?

portality/static/widget/fqw_js.js
amdomanska: Autogenerated as it used to be. Even though I don't like it ;)

sssoz:
@amdomanska and I agreed that it would be more simple for me to just create a static version of my modifications to the widget’s layout. The change (fairly minimal) has been committed and pushed to this branch: [feature/1679-widgets-styling)](https://github.com/DOAJ/doaj/tree/feature/1679-widgets-styling). 

You can see the markup in: `/Users/sophy/git/doaj/static_content/static-widget-layout.md` (so everything within the `<div id="doaj-fixed-query-widget">`). I’ve removed all of the automatically-generated classes & only displayed one result in the example.

portality/static/widget/fixed_query_src_edges.js
amdomanska: no, it doesn't, The widget_fv_options is now the local variable that puts together the params no matter whether they were provided by `QUERY_OPTIONS` or `SEARCH_CONFIGURED_OPTIONS`. Discussed now on call.

portality/static/widget/public_search_config.js
amdomanska: that is the error. It has to be changed to `doaj_url + '\query\journal,article\_search?ref=fqw'`

portality/view/doaj.py
amdomanska: I left it for now and I will just update that to the final version. I will create the separate template for the widget to make sure this is really separated from the public search template

amdomanska:
Merged with https://github.com/DOAJ/doaj/pull/1866

tested with the generated snippet

portality/static/js/edges/public.article.edge.js
```diff
@@ -4,8 +4,16 @@ $.extend(true, doaj, {
         activeEdges : {},
 
         embedSnippet : function(renderer) {
-            var snip = '<script type="text/javascript">!window.jQuery && document.write("<scr" + "ipt type=\"text/javascript\" src=\"http://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js\"></scr" + "ipt>"); </script><script type="text/javascript">var doaj_url="https://doaj.org"; var SEARCH_CONFIGURED_OPTIONS={{QUERY}}</script><script src="https://doaj.org/static/widget/fixed_query.js" type="text/javascript"></script><div id="doaj-fixed-query-widget"></div></div>';
-            var query = renderer.component.edge.currentQuery.objectify({
+            var snip = `<script type="text/javascript">
```
amdomanska, line 7: This is change from https://github.com/DOAJ/doajPM/issues/2843 with jquery script removed

portality/static/js/edges/public.journal.edge.js
```diff
@@ -4,8 +4,17 @@ $.extend(true, doaj, {
         activeEdges : {},
 
         embedSnippet : function(renderer) {
-            var snip = '<script type="text/javascript">!window.jQuery && document.write("<scr" + "ipt type=\"text/javascript\" src=\"http://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js\"></scr" + "ipt>"); </script><script type="text/javascript">var doaj_url="https://doaj.org"; var SEARCH_CONFIGURED_OPTIONS={{QUERY}}</script><script src="https://doaj.org/static/widget/fixed_query.js" type="text/javascript"></script><div id="doaj-fixed-query-widget"></div></div>';
-            var query = renderer.component.edge.currentQuery.objectify({
+            var snip =
```
amdomanska, line 7: This is change from https://github.com/DOAJ/doajPM/issues/2843 with jquery script removed

portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -125,8 +132,12 @@ $.extend(true, doaj, {
                         </div>\
                     </header>';
                 }
-
-                var frag = titleBarFrag + '\
+                var frag = "";
```
amdomanska, line 135: I'm not entirely sure why github shows that as a difference. The publicSearch template has not been shown, `if(this.resultsOnly)` has been here since the beginning. The only difference to this file should be FQWidget template added.

richard-jones: It looks like it's just because of the `resultsOnly` if statement that's been inserted.  it's all fine.

portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -109,6 +115,7 @@ $.extend(true, doaj, {
             this.title = edges.getParam(params.title, "");
 
             this.titleBar = edges.getParam(params.titleBar, true);
+            this.resultsOnly = edges.getParam(params.resultsOnly, false);
```
richard-jones, line 118: Is this now necessary, if the FQWidget has its own template?

richard-jones: Answering for myself - this template is not used with this arg, so technically it can come out.

richard-jones (commented):
Just done a first pass through as I pick this up.  A few things for me to do:

* Review whether we need the query_filters bit at all
* Remove `resultsOnly` from the PublicSearch template as no longer needed
* Consider moving the fqw source out from the static directory, in line with other changes we have made in other branches
* Rationalise this with the new static pages layout; whether that's waiting to merge this, or merging this down and the develop back up to static pages

richard-jones (approved):
I've now reviewed this and made a few final changes, so I think it's ready for release.

I would like to put it up on test and try making some embeds from there just to make sure everything is in order before release

sssoz:
@amdomanska Just going to review the stylesheets added now.

cms/sass/components/_fqw.scss
```diff
@@ -0,0 +1,68 @@
+.facetview {
+    height: 100%;
+    background-color: #FEFBF8;
```
sssoz, line 3: Need to use colour variables here.

cms/sass/components/_fqw.scss
```diff
@@ -0,0 +1,68 @@
...
+                text-transform: none;
+
+                :last-child {
+                    margin-bottom: 0;
+                }
+            }
+        }
+    }
+
+    @import "../base/_fonts";
```
sssoz, line 49: @amdomanska Is this because the widget _only_ uses this stylesheet and none of the other ones?

cms/sass/components/_fqw.scss
```diff
@@ -0,0 +1,68 @@
+.facetview {
```
sssoz, line 1: Does not use BEM convention for naming -- and it’s not a component. I am not quite sure though because it seems that the widget will only be using this stylesheet, right? If so, I would place elsewhere (outside of the components directory).

cms/sass/components/_fqw.scss
amdomanska: Yes.

cms/sass/components/_fqw.scss
amdomanska: It has to start with the facet view class to make sure we are back-compatible with previous versions. Where would you like to put it? To base? 

And yes, the widget will use only this stylesheet. It cannot import main.css because it had an impact on the host site that, obviously, we cannot do.

cms/sass/components/_fqw.scss
amdomanska: done.

cms/sass/components/_fqw.scss
amdomanska: Moved to: `cms/sass/fq_widget.scss`

amdomanska:
@sssoz for some reason when we embed the simple widgets on the doaj site the radiobuttons do not work. I tried to embed them on the dynamic site instead of the static one and it does not make any difference, so it is not connected in anyway to how they are generated. When tested on different sites they work fine.
Here is where you can check how radiobuttons (do not) work on doaj site:
https://testdoaj.cottagelabs.com/docs/widgets/
And here is the test site that just embed the widget but mimic the 3rd party:
https://testdoaj.cottagelabs.com/static/widget/ssw_test.html
Could you possibly have a look at that? It seems to be connected to how the radiobuttons are styled on doaj. It would be useful to know why it is like that not only to possibly fix the docs site (where the example simple search widget is not fully functioning) but also to make sure that it won't happen for any other host. 
Thank you!

sssoz:
@amdomanska I will have to look at it later this week or at the beginning of next week, is that alright?

PR #1868, opened 2021-03-12
only show archiving policy box on toc if has_preservation is True

richard-jones:
UI parts for: https://github.com/DOAJ/doajPM/issues/2812

### DIFF
--- a/portality/templates/doaj/toc.html
+++ b/portality/templates/doaj/toc.html
@@ -306,7 +306,7 @@
                                 </div>
                             </article>
 
-                            {% if bibjson.preservation %}
+                            {% if bibjson.has_preservation %}
                                 <article class="card card--with-icon">
                                     <span data-feather="archive" aria-hidden="true"></span>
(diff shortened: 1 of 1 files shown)


portality/templates/doaj/toc.html
```diff
@@ -306,7 +306,7 @@ <h3 class="label label--underlined">Best practice</h3>
                                 </div>
                             </article>
 
-                            {% if bibjson.preservation %}
+                            {% if bibjson.has_preservation %}
```
Steven-Eardley, line 309: I'm seeing `bibjson.preservation.has_preservation` in the live data

Steven-Eardley (changes requested):

portality/templates/doaj/toc.html
Steven-Eardley: Richard clarified it was a property, oops.

Steven-Eardley (approved):

PR #1877, opened 2021-03-29
Display links to ISSN Portal records on journal TOCs

sssoz:
See DOAJ/doajPM#2794


richard-jones (approved):
Code wise this is fine, just wondering what happens if the record isn't in ROAD, and if that's even possible?

sssoz:
If a non-existent ISSN is input, it’ll just tell the user that this ISSN is not on ROAD, which is fine. @dommitchell

dommitchell:
Not possible for an ISSN to be in DOAJ and not in the Portal.

sssoz:
@Steven-Eardley Updated text, good to go

PR #1881, opened 2021-04-02
add issn is empty string option to coerce to_issn method

amdomanska:
Patch for:
https://github.com/DOAJ/doajPM/issues/2882


richard-jones (approved):

richard-jones (changes requested):
What effect does this change have on the data?  It looks to me like `to_issn` is being called here with an invalid value.  The rest of the `to_issn` does a valiant if naive job of actually returning a 9 character ISSN, and when given "" return `0000-0000`.  This change would mean it would return the empty string.

My preference would be to also raise a `ValueError` on an empty string, and then for the code that calls this to sanitise its ISSNs before this method is called, if possible.

If we already have empty strings in the ISSN data we might need to strip them when we make this change.

amdomanska:
Changed. ISSN provided as an empty string will cause ValueError to be thrown and then utilized by code. Tested and such ISSN is omitted in such condition. I haven't found any journals or applications with empty string ISSN but we've got 62 applications with ISSN `0000-0000`. I would suggest if we want to clean them up programmatically to open the new issue.

portality/lib/coerce.py
```diff
@@ -73,7 +73,8 @@ def to_country_code(val):
     return uc(nv)
 
 def to_issn(issn):
-    if len(issn) > 9:
+
+    if len(issn) > 9 or issn == '':
```
richard-jones, line 77: Can we look upstream to whichever crosswalks are writing issns into the data, and make sure that they are now checking for empty issns, and just omitting them?  That will prevent error cases in most normal "user error" kind of ways, while leaving us a defence here for anywhere they slip through the net.  I'd guess the journal form crosswalk is the place to do it?

richard-jones (changes requested):

portality/lib/coerce.py
amdomanska: I tested the code and now if the empty string is provided as issn it is just omitted. I will add some UT to make sure it works fine  :)

richard-jones: It looks to me that `allow_coerce_failure` is set on the `IncomingApplication` data object, which means if there's an exception here, that will be swallowed and the empty string used as the issn value.  This implies that the empty string would be entered into the issn field, and I can't see anywhere that it checks for the empty string, so I'd want to double check that it isn't writing that empty string into the issn field.

PR #1882, opened 2021-04-02
simple widget functionality

amdomanska:
https://github.com/DOAJ/doajPM/issues/2653

Just functionality, waiting for @sssoz for styles.


portality/static/widget/simple_search.js
```diff
@@ -37,9 +37,31 @@ function scriptLoadHandler() {
 
 /******** Our main function ********/
 function main() { 
-    jQuery(document).ready(function($) {
```
richard-jones, line 40: In this file, near the top we explicitly import an old version of jquery from a google cdn.  We should load this as the latest version from our domain if we need it.  Though I'm not actually convinced we need jquery here at all

portality/static/widget/simple_search_body.html
```diff
@@ -1,12 +1,35 @@
-<link rel="stylesheet" href="https://doaj.org/static/widget/simple_search_widget.css">
-
-<form id="doaj-search-form" name="topsearch" action="https://doaj.org/search" method="POST">
-    <input type="hidden" name="ref" value="ssw">
-    <div class="control-group" id="doaj-search-container">
-        <a href="https://doaj.org/" target="_blank"><img id="doaj-simple-search-logo" src="https://doaj.org/static/doaj/images/logo_cropped.jpg" alt="Directory of Open Access Journals" title="Directory of Open Access Journals"></a>
-        <div>
-            <input type="text" name="q" placeholder="Search the DOAJ" id="doaj-search-input"/><span id="doaj-search-button-container"><button type="submit" formtarget="_blank" id="doaj-search-button">Search</button></span>
+<div class="modal__dialog" role="document">
+    <form class="quick-search__form" role="search" id="doaj-search-form" action="http://localhost:5004/search" method="POST">
```
richard-jones, line 2: Don't hard-code the form action

richard-jones (changes requested):

portality/static/widget/simple_search.js
```diff
@@ -37,9 +37,31 @@ function scriptLoadHandler() {
...
+
+        let scr  = document.createElement('script');
+        scr.src = doaj_url + '/static/vendor/feather/feather.min.js';
+        scr.async = false;
+        scr.defer = false;
+        head.insertBefore(scr, head.firstChild);
+
+        scr = document.createElement('link');
+        scr.rel = 'stylesheet';
+        scr.href = doaj_url + '/static_content/_site/css/main.css';
```
sssoz, line 58: This URL has changed with https://github.com/DOAJ/doaj/pull/1880

portality/static/widget/simple_search_body.html
```diff
@@ -1,12 +1,35 @@
-<link rel="stylesheet" href="https://doaj.org/static/widget/simple_search_widget.css">
-
-<form id="doaj-search-form" name="topsearch" action="https://doaj.org/search" method="POST">
-    <input type="hidden" name="ref" value="ssw">
-    <div class="control-group" id="doaj-search-container">
-        <a href="https://doaj.org/" target="_blank"><img id="doaj-simple-search-logo" src="https://doaj.org/static/doaj/images/logo_cropped.jpg" alt="Directory of Open Access Journals" title="Directory of Open Access Journals"></a>
-        <div>
-            <input type="text" name="q" placeholder="Search the DOAJ" id="doaj-search-input"/><span id="doaj-search-button-container"><button type="submit" formtarget="_blank" id="doaj-search-button">Search</button></span>
+<div class="modal__dialog" role="document">
```
sssoz: It’s not a modal, so we should not use this class (also with BEM, we should not be using element classes, `modal__dialog` without their corresponding blocks, `modal`).

portality/static/widget/simple_search_body.html
```diff
@@ -1,12 +1,35 @@
...
+    <form class="quick-search__form" role="search" id="doaj-search-form" action="http://localhost:5004/search" method="POST">
+        <h2 class="modal__title">
```
sssoz: Idem here

portality/static/widget/simple_search.js
sssoz: Stylesheets and classes have also been moved around -- possible to rebase this branch off of this one? https://github.com/DOAJ/doaj/pull/1880  It will be much easier to style (which should be pretty minimal)

sssoz:
@amdomanska I’ve left a few comments for both you and Richard. Did go over the `simple_search_body.html` template quickly and fix some of the things I  mentioned. It should work for now; if nothing else needs to be added, I think we can have it reviewed by Dom?

amdomanska:
@sssoz Thank you a lot for your input. Although I would be grateful if in the future you could create a PR or mention what needs to be changed in the comment instead of just making the changes directly to my branch. Thank you.

portality/static/widget/simple_search.js
amdomanska: As discussed: we do jquery here.

amdomanska: >We should load this as the latest version from our domain if we need it.

done 👍🏼

amdomanska:
Requested changes made. Will have to be merged with https://github.com/DOAJ/doaj/pull/1880 after that is merged with master

amdomanska:
merged into: https://github.com/DOAJ/doaj/pull/1863

PR #1885, opened 2021-04-09
make subject tree sort work across chrome and firefox (and in general…

richard-jones:
…, in fact)

### DIFF
--- a/portality/static/js/doaj.fieldrender.edges.js
+++ b/portality/static/js/doaj.fieldrender.edges.js
@@ -67,7 +67,7 @@ $.extend(true, doaj, {
                             }
                             displayTree.push(entry);
                         }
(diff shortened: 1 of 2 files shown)


portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -67,7 +67,7 @@ $.extend(true, doaj, {
                             }
                             displayTree.push(entry);
                         }
-                        displayTree.sort((a, b) => a.display > b.display);
+                        displayTree.sort((a, b) => a.display > b.display ? 1 : -1);
```
Steven-Eardley, line 70: such a teeny tiny change, but I've learnt a heck of a lot so I can understand it.

Steven-Eardley (approved):

PR #1887, opened 2021-04-22
fix cms edit url and data button for pages where data is in preface r…

richard-jones:
…ather than include frontmatter


Steven-Eardley (approved):
Oops, yes. I did notice the correct url comment on review then promptly forgot it on release.

PR #1888, opened 2021-04-22
Refactor Harvester to the background task

amdomanska:
https://github.com/DOAJ/doajPM/issues/2153

READY FOR INITIAL REVIEW.

Still to be done: 
- [x] Unit Tests
- [x] Change of app.logger to job.audit_logs


portality/tasks/harvester.py
```diff
@@ -0,0 +1,128 @@
...
+        if app.debug:
+            # Augment the default flask debug log to include a timestamp.
+            app.debug_log_format = (
+                    '-' * 80 + '\n' +
+                    '%(asctime)s\n'
+                    '%(levelname)s in %(module)s [%(pathname)s:%(lineno)d]:\n' +
+                    '%(message)s\n' +
+                    '-' * 80
+            )
+            flask.logging.create_logger(app)
```
richard-jones, line 95: I don't think this is needed any more

portality/tasks/harvester.py
```diff
@@ -0,0 +1,128 @@
...
+
+        :param background_job: the BackgroundJob instance
+        :return:
+        """
+        background_job.save()
+        harvest.schedule(args=(background_job.id,), delay=10)
+        # fixme: schedule() could raise a huey.exceptions.HueyException and not reach redis- would that be logged?
+
+
+@main_queue.periodic_task(schedule("harvest"))
```
richard-jones, line 116: suggest that we put this on the `long_running` queue instead, as the harvester can run for a while

portality/tasks/harvester.py
```diff
@@ -0,0 +1,128 @@
...
+@write_required(script=True)
+def scheduled_harvest():
+    user = app.config.get("SYSTEM_USERNAME")
+    job = HarvesterBackgroundTask.prepare(user)
+    HarvesterBackgroundTask.submit(job)
+
+@main_queue.task()
+@write_required(script=True)
```
richard-jones, line 124: `long_running` queue here too

richard-jones (changes requested):

doajtest/unit/apiv2_tests/test_apiv2_discovery.py
```diff
@@ -1,4 +1,4 @@
-from doajtest.helpers import DoajTestCase
+`from doajtest.helpers import DoajTestCase
```
richard-jones, line 1: erroneous character?

doajtest/unit/resources/harvester_resp.json
```diff
@@ -0,0 +1 @@
...ication": 2010, "printPublicationDate": "2010-01-01", "journal": {"title": "My Journal", "medlineAbbreviation": "My Jour", "essn": "1234-5678", "issn": "9876-5432", "isoabbreviation": "My Jour", "nlmid": "123456789"}}, "pubYear": "2010", "pageInfo": "1", "affiliation": "aga@cottagelabs.com", "publicationStatus": "publish", "language": "eng", "pubModel": "Print-Electronic", "pubTypeList": {"pubType": ["Comment", "research-article", "Journal Article"]}, "commentCorrectionList": {"commentCorrection": [{"id": "12345678", "source": "MED", "reference": "My Jour. 2010 Jan;1(1):e1001011", "type": "Comment on", "orderIn": 1}]}, "isOpenAccess": "Y", "inEPMC": "Y", "inPMC": "Y", "hasPDF": "Y", "hasBook": "N", "hasSuppl": "N", "citedByCount": 0, "hasData": "N", "hasReferences": "N", "hasTextMinedTerms": "Y", "hasDbCrossReferences": "N", "hasLabsLinks": "N", "license": "cc by", "authMan": "N", "epmcAuthMan": "N", "nihAuthMan": "N", "hasTMAccessionNumbers": "N", "dateOfCompletion": "2011-08-19", "dateOfCreation": "2011-05-10", "firstIndexDate": "2011-05-11", "fullTextReceivedDate": "2020-07-09", "electronicPublicationDate": "2011-04-12", "firstPublicationDate": "2011-04-12"}]}, "version": "6.5"}
```
richard-jones, line 1: Very unimportant, but it might be best to pretty print this so we can easily inspect it by eye during development down the line

doajtest/unit/test_harvester.py
```diff
@@ -0,0 +1,97 @@
...
+from portality.tasks.harvester_helpers.workflow import HarvesterWorkflow
+from portality.models.harvester import HarvestState
+from portality.models.harvester import HarvesterPlugin
+
+
+class TestHarvester(DoajTestCase):
+
+
+    def mock_get_journals_issns(self, string1):
+        return ["1234-5678", "9876-5432"]
```
richard-jones, line 28: I think we decided that this was no longer necessary, and we should handle it by injecting the correct data into the database instead

doajtest/unit/test_harvester.py
```diff
@@ -0,0 +1,97 @@
...
+        HarvesterWorkflow.get_journals_issns = self.mock_get_journals_issns
+
+        self.publisher = models.Account(**AccountFixtureFactory.make_publisher_source())
+        self.journal = models.Journal(**JournalFixtureFactory.make_journal_source())
+        self.publisher.add_journal(self.journal.id)
+        self.publisher.generate_api_key()
+
+        self.publisher.save()
+        self.journal.save()
+        time.sleep(2)
```
richard-jones, line 50: Now that we figured out why the blocking save wasn't working, I'd go back to using that in the journal save

doajtest/unit/test_harvester.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.old_harvester_api_keys = app.config.get('HARVESTER_API_KEYS')
+        self.old_initial_harvest_date = app.config.get("INITIAL_HARVEST_DATE")
+
+
+        app.config['HARVESTER_API_KEYS'] = {self.publisher.id: self.publisher.api_key}
+
+        self.old_get = httputil.get
```
richard-jones, line 58: the httputil.get doesn't seem to get overridden by the mock

doajtest/unit/test_harvester.py
```diff
@@ -0,0 +1,97 @@
...
+        json_file.close()
+
+    def tearDown(self):
+        super(TestHarvester, self).tearDown()
+        app.config['HARVESTER_API_KEYS'] = self.old_harvester_api_keys
+        app.config["INITIAL_HARVEST_DATE"] = self.old_initial_harvest_date
+        HarvesterWorkflow.get_journals_issns = self.mock_get_journals_issns
+
+    @patch('portality.tasks.harvester_helpers.epmc.client.EuropePMC.query')
+    def test_harvest(self, mock_query):
```
richard-jones, line 77: I think we planned to only override httputil.get, so that our test also works out the `query` method

portality/tasks/harvester_helpers/workflow.py
```diff
@@ -1,42 +1,47 @@
...
 
         # now update the issn states
         HarvesterWorkflow.process_issn_states(account_id, issns)
 
         for issn in issns:
             HarvesterWorkflow.process_issn(account_id, issn)
 
+    @classmethod
+    def get_journals_issns(cls, account_id):
+        return Journal.issns_by_owner(account_id)
```
richard-jones, line 32: An observation from another thread of work - we should probably only consider journals that are `in_doaj`, so you can add `in_doaj=False` to this method call, I think.

portality/tasks/harvester_helpers/workflow.py
```diff
@@ -1,42 +1,47 @@
...
     def process_issn_states(cls, account_id, issns):
         # first check that there are state records for all the provided issns,
         # and that if they were deactivated they are now reactivated
         for issn in issns:
             state = HarvestState.find_by_issn(account_id, issn)
             if state is not None:
-                if state.suspended:
+                #TODO: HarvesterStateFixture does not provide suspended, next line for the test only
+                if state["status"] == "suspended":
+                #if state.suspended:
```
richard-jones, line 44: I'm not sure I understand what's happening here.  The HarvestState object should have a method/property `suspended`.  Having something that's for test only that seems to work around the HarvestState object API seems odd and risky.  Instead if there's a problem with what the object looks like in the test, then we should fix the test.

richard-jones (changes requested):

doajtest/unit/test_harvester.py
amdomanska: Yes, true. But I haven't found a way to get to work correctly with Mocks. I will write a separate test to test query method only to get that cover (that is a separate issue, that I feel like my test covers a way too big chunk of code, probably something to think about later).

portality/tasks/harvester_helpers/workflow.py
amdomanska: ok, but I will add that as the default parameter so it could be called differently if needed.

portality/tasks/harvester_helpers/workflow.py
amdomanska: very true. I meant to change that and forgot about that part, sorry!

doajtest/unit/test_harvester.py
amdomanska: Yes, it is unnecessary

portality/tasks/harvester_helpers/workflow.py
amdomanska: The `Journal.issns_by_owner` method does not accept `in_doaj` param:
```
@classmethod
    def issns_by_owner(cls, owner):
        q = IssnQuery(owner)
        res = cls.query(q=q.query())
        issns = [term.get("term") for term in res.get("facets", {}).get("issns", {}).get("terms", [])]
        return issns
```
I think it is actually a good idea to have it. I'll add the implementation.

portality/models/background.py
```diff
@@ -84,6 +84,7 @@ def add_audit_message(self, msg, timestamp=None):
             timestamp = dates.now_with_microseconds()
         obj = {"message": msg, "timestamp": timestamp}
         self._add_to_list_with_struct("audit", obj)
+        print(msg)
```
Steven-Eardley, line 87: do we want print statements rather than logging?

portality/settings.py
```diff
@@ -1024,30 +1025,30 @@
 
 ## Configuration options for the DOAJ API Client
 
-DOAJ_SEARCH_BASE = "https://doaj.org"
+DOAJ_SEARCH_BASE = "https://localhost:5004"
```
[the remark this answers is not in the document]
Steven-Eardley: bump for this, we can't merge as-is.

portality/tasks/harvester_helpers/workflow.py
```diff
@@ -56,7 +61,7 @@ def process_issn_states(cls, account_id, issns):
     @classmethod
     @capture_sigterm
```
Steven-Eardley, line 62: We won't have the sigterm control flow any more

Steven-Eardley (changes requested):
Main thing is removing the process name and sigterm control flow. They were to ensure we don't get more than one harvester running at the same time. I suppose it could still happen with background jobs, but none of our other tasks check whether there's a previous run in progress.

portality/models/background.py
amdomanska: no, we're logging them as we used to and I just added the print statement at the end of log to make test debugging easier.

portality/tasks/harvester_helpers/epmc/client.py
```diff
@@ -102,45 +103,54 @@ def complex_search(cls, query_builder, cursor="", page_size=25):
...
                 break
             for r in results:
                 yield r
 
     @classmethod
-    def query(cls, query_string, cursor="", page_size=25):
-        """
-        :return: (results, next_cursor)
-        """
+    def url_from_query(cls, query_string, cursor, page_size):
```
amdomanska, line 130: This diff is very weirdly presented. All that happened here is the extraction of url_from_query method outside of query method.

.gitignore
```diff
@@ -114,3 +114,6 @@ cms/fragments*
 !portality/static/doaj/css/*_widget.css
 
 portality/static/widget/fq_widget_build_info.txt
+
+#json for harvester test
+doajtest/unit/resources/harvester_resp.json
```
richard-jones, line 119: Should this file be in the gitignore?  It's also in the PR

portality/models/background.py
richard-jones, line 87: Just a reminder note to remove print statements

portality/models/harvester.py
```diff
@@ -8,7 +8,7 @@ class HarvesterPlugin(object):
     def get_name(self):
         raise NotImplementedError()
 
-    def iterate(self, issn, since, to=None):
+    def iterate(self, issn, since, job, to=None):
```
richard-jones, line 11: job has been added to the signature, but it doesn't seem to be used anywhere inside the model

portality/models/v2/journal.py
```diff
@@ -966,8 +966,25 @@ def query(self):
...
         "query": {
-            "term": {"admin.owner.exact": "<owner id here>"}
-        },
+         "filtered": {
+           "filter": {
+             "bool": {
+               "must": [
+                 {
+                   "term": {
+                     "admin.in_doaj": "True"
```
richard-jones, line 975: If this is set, it should be a boolean `True` not string `"True"`

portality/models/v2/journal.py
```diff
@@ -980,9 +997,12 @@ class IssnQuery(object):
         }
     }
 
-    def __init__(self, owner):
+    def __init__(self, owner, in_doaj=False):
         self._query = deepcopy(self.base_query)
-        self._query["query"]["term"]["admin.owner.exact"] = owner
+        self._query["query"]["filtered"]["query"]["term"]["admin.owner.exact"] = owner
+        in_doaj_str = "True" if in_doaj else "False"
```
richard-jones, line 1003: Same here, should just be the boolean

portality/models/v2/journal.py
```diff
@@ -980,9 +997,12 @@ class IssnQuery(object):
         }
     }
 
-    def __init__(self, owner):
+    def __init__(self, owner, in_doaj=False):
```
richard-jones, line 1000: Does adding this default to `False` change any other usages of this query?  The previous version didn't consider `in_doaj` at all, so returned everything.  Now we will definitely only return in_doaj=True or in_doaj=False values.

Also, if the default value is valid at all, the default to `False` seems risky.  You should probably have to intentionally ask for things that are not in DOAJ, otherwise you should just get the stuff that is.

portality/models/v2/journal.py
```diff
@@ -70,8 +70,8 @@ def find_by_issn(cls, issns, in_doaj=None, max=10):
         return records
 
     @classmethod
-    def issns_by_owner(cls, owner):
-        q = IssnQuery(owner)
+    def issns_by_owner(cls, owner, in_doaj=None):
```
richard-jones, line 73: The default to `None` here is at odds with the default to `False` of its usage with the `ISSNQuery`

portality/tasks/consumer_main_queue.py
```diff
@@ -24,3 +24,4 @@
 from portality.tasks.async_workflow_notifications import async_workflow_notifications
 from portality.tasks.check_latest_es_backup import scheduled_check_latest_es_backup, check_latest_es_backup
 from portality.tasks.request_es_backup import scheduled_request_es_backup, request_es_backup
+from portality.tasks.harvester import scheduled_harvest
```
richard-jones, line 27: As we're putting these on the long_running queue, they should be imported in `consumer_long_running.py` instead

portality/tasks/harvester_helpers/epmc/client.py
```diff
@@ -58,6 +58,7 @@ def __init__(self, message, rawstring, *args, **kwargs):
 
 
 class EuropePMC(object):
+    job = None
```
richard-jones, line 61: The EMPC client should not know about our job queue.  For example, what if we want to run the client independently of the job queue, or integrate it into some other part of the system as well?  Architecturally this is not a good approach, it creates a two-way binding between modules which should only have a one-way binding.

The client should be a generic client, and the job queue should wrangle it.  Looking into why the job is passed in, it's for the recording of logs, and for no other functional purpose.  If that's the case I'd suggest 2 alternative approaches:

1. pass in a "logger" for the client to use.  This could be provided by the job queue to route the logs to the job's audit trail
2. Have the client log to a state variable (e.g. EuropePMC.log) and have the task read the logs into its own audit logs after the event.

portality/tasks/harvester_helpers/epmc/epmc_harvester.py
```diff
@@ -1,16 +1,18 @@
...
 from datetime import datetime
 import time
 
+
 class EPMCHarvester(HarvesterPlugin):
     def get_name(self):
         return "epmc"
 
-    def iterate(self, issn, since, to=None):
+    def iterate(self, issn, since, job, to=None):
```
richard-jones, line 14: See comments in previous file - job should not be passed in here

portality/tasks/harvester_helpers/workflow.py
```diff
@@ -1,34 +1,35 @@
...
-        gen = doaj.field_search_iterator("journals", "username", account_id)
-        issns = []
-        for journal in gen:
-            issns += journal.all_issns()
-        issns = list(set(issns))
+    job = None
 
-        app.logger.info("Account:{x} has {y} issns to harvest for: {z}".format(x=account_id, y=len(issns), z=",".join(issns)))
+    @classmethod
+    def process_account(cls, account_id, job):
```
richard-jones, line 16: Another place where job should not be passed in

richard-jones (changes requested):
A few odds and ends, and one main architectural issue that we should address before merge

.gitignore
amdomanska: Well, this file is changed every single time the harvester test is run and I don't think we want to have it in every single PR later. That is why I added that to the repo and then set github to ignore it. Not sure what is the best solution here though.

portality/models/background.py
amdomanska: I actually added the print message to make debugging of the background jobs easier. I will add `if app.config.get("DOAJENV") == 'dev'` to make sure it prints only in dev.

portality/tasks/harvester_helpers/epmc/client.py
```diff
@@ -58,7 +60,16 @@ def __init__(self, message, rawstring, *args, **kwargs):
...
-    job = None
+
+    logger = None
+
+    @classmethod
+    def _write_to_logger(cls, msg):
+        try:
+            cls.logger.write(msg)
+        except:
+            app.logger.warn("Logger not provided for Harvester Background task, message: {}").format(msg)
```
richard-jones, line 71: Probably no need to issue the warning here, just write to the `app.logger` if no custom logger is supplied

portality/models/background.py
richard-jones: probably use `if app.config.get("DEBUG") == True` for more generality.

portality/tasks/harvester_helpers/workflow.py
```diff
@@ -1,34 +1,56 @@
...
-    @capture_sigterm
-    def process_account(cls, account_id):
-        app.logger.info("Harvesting for Account:{x}".format(x=account_id))
+    def create_logger(cls):
+        try:
+            cls.logger = open(BASE_FILE_PATH + "/Harvester_logger.log", "x")
+        except:
+            cls.logger = open(BASE_FILE_PATH + "/Harvester_logger.log", "wr")
+            cls.logger.truncate(0)
+        return cls.logger
```
richard-jones, line 22: For simplicity, I'd consider just keeping the log information in memory, as it's going to get written to the audit logs down the line anyway.  That saves you just writing it to disk.

If you are certain you want to write it to disk, then we should use our file store instead.  There is a Temporary File store implementation which can be used to stage files.

```
from portality import store
import io

tmpStore = store.StoreFactory.tmp()
tmpStore.store("harvester", "Harvester_logger-2021-06-15.log", source_stream=io.StringIO(""))
loggerFile = tmpStore.get("harvester", "Harvester_logger-2021-06-15.log")

loggerFile.write("log line")
...

# then later
tmpStore.delete_file("harvester", "Harvester_logger-2021-06-15.log")

File: portality/tasks/harvester_helpers/workflow.py [hunk 28]
```diff
@@ -1,34 +1,56 @@
...
-from portality.decorators import capture_sigterm
+from portality.models import Journal, Account
 from portality.models.harvester import HarvesterProgressReport as Report
 from portality.models.harvester import HarvestState
 from portality.lib.dataobj import DataObjException
+from portality.settings import BASE_FILE_PATH
 
 
 class HarvesterWorkflow(object):
+    logger = None
```
richard-jones, line 13: It doesn't feel right to create a per-run logger on a global singleton member variable (it may have side-effects which could cause runtime errors down the line).  We should either:

1. Pass the logger in wherever it is needed by a `@classmethod`
2. Change the HarvesterWorkflow class so that it is instantiated when it is run, with a logger as an instance member variable

richard-jones (changes requested):
Added a few comments on the implementation of the logging

amdomanska:
@richard-jones another run on logger implementation

richard-jones (approved):
Reviewed this with @amdomanska yesterday, changes look good to go

portality/tasks/harvester_helpers/runner.py
```diff
@@ -84,8 +84,9 @@ def run_only_once():
         flask.logging.create_logger(app)
 
     accs = list(app.config.get("HARVESTER_API_KEYS", {}).keys())
```
Steven-Eardley, line 86: An observation here - we still use `HARVESTER_API_KEYS` which is a dictionary of `{user_id: doaj_api_key}`, but we're no longer creating items via the API. the configuration structure no longer makes sense. (no problem they're just ignored) We may wish to change this to contain which harvester(s) to run per account.

We might want to consider adding a role 'harvester' so it's easier to see and manage which accounts retrieve via the harvester.

Steven-Eardley:
testing successful on the test server (using manual run)

PR #1890, opened 2021-04-23
Feature/2870 subject search

richard-jones:
Changes for: https://github.com/DOAJ/doajPM/issues/2870

This does the following things:

* adds a mechanism to generate search urls algorithmically (rather than by copy and paste of the search url with substitutions)
* Changes the ToC to search for subject specifically by LCC code rather than subject term
* Changes the save routine for journals and applications to index the full classification path into index.classification
* Adds a migration script to apply the reindex


Steven-Eardley (approved):
My compliments; this is superb.

PR #1891, opened 2021-04-23
Feature/2778 model improvements

richard-jones:
A variety of minor, non-functionality changing code tweaks:

* addition of form documentation
* Removing comments and other dead bits of code
* refactoring generated documentation layout

I don't think this needs to be deployed, it can just be merged down so it becomes part of the master branch eventually.


portality/forms/application_forms.py
```diff
@@ -2879,3 +2879,39 @@ def wtform(formulaic_context, field, wtfargs):
 
 ApplicationFormFactory = Formulaic(APPLICATION_FORMS, WTFORMS_BUILDERS, function_map=PYTHON_FUNCTIONS, javascript_functions=JAVASCRIPT_FUNCTIONS)
 JournalFormFactory = Formulaic(JOURNAL_FORMS, WTFORMS_BUILDERS, function_map=PYTHON_FUNCTIONS, javascript_functions=JAVASCRIPT_FUNCTIONS)
+
+
+if __name__ == "__main__":
```
Steven-Eardley, line 2884: Can you add a few comments to explain the documention steps here and expected use?

Steven-Eardley (approved):
Looks great, although I do think we need to explain the purpose of the documentation step somewhere

portality/lib/query_filters.py
```diff
@@ -186,11 +186,12 @@ def prune_author_emails(results, unpacked=False):
 
 
 def publisher_result_filter(results, unpacked=False):
+    allowed_admin = ["ticked", "seal", "in_doaj", "related_applications", "current_application", "current_journal", "application_status"]
```
Steven-Eardley, line 189: I've found a test which references these properties, `doajtest/unit/test_query_filters.py` - I'll hotfix that since I've merged this PR already

portality/forms/application_forms.py
richard-jones: I've opened a new issue to do that, since this one is merged: https://github.com/DOAJ/doajPM/issues/2918

PR #1892, opened 2021-04-25
Continuation for closed https://github.com/DOAJ/doaj/pull/1881

amdomanska:
Correct changes to https://github.com/DOAJ/doaj/pull/1881, 
Patch for:
DOAJ/doajPM#2882

changes added to coerce.py and application_2_journal crosswalk and new UT provided to make sure that empty string issn in an application will be ignored.

TODO:
- [x] check the data for any empty strings in journals' issns
- [x] make sure we do not have to worry about empty strings issn in crossref articles (DOAJ schema takes care of it)

### DIFF
--- a/portality/bll/services/application.py
+++ b/portality/bll/services/application.py
@@ -365,6 +365,12 @@ class ApplicationService(object):
             journal.set_owner(application.owner)
         journal.set_seal(application.has_seal())
 
+        b = application.bibjson()
+        if b.get_one_identifier("pissn") == "":
+            b.add_identifier("pissn", None)
+        if b.get_one_identifier("eissn") == "":
+            b.add_identifier("eissn", None)
+
         # no relate the journal to the application and place it in_doaj
         journal.add_related_application(application.id, dates.now())
         journal.set_in_doaj(True)
--- a/portality/bll/services/application.py
+++ b/portality/bll/services/application.py
@@ -366,9 +366,9 @@ class ApplicationService(object):
         journal.set_seal(application.has_seal())
 
         b = application.bibjson()
-        if b.get_one_identifier("pissn") == "":
+        if b.pissn == "":
             b.add_identifier("pissn", None)
-        if b.get_one_identifier("eissn") == "":
+        if b.eissn == "":
             b.add_identifier("eissn", None)
 
         # no relate the journal to the application and place it in_doaj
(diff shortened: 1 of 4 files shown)


amdomanska:
>     * make sure we do not have to worry about empty strings issn in crossref articles (DOAJ schema takes care of it)

Crossref schema does not accept empty string issn either.

portality/bll/services/application.py
```diff
@@ -365,6 +365,12 @@ def application_2_journal(self, application, manual_update=True):
             journal.set_owner(application.owner)
         journal.set_seal(application.has_seal())
 
+        b = application.bibjson()
+        if b.get_one_identifier("pissn") == "":
+            b.add_identifier("pissn", None)
+        if b.get_one_identifier("eissn") == "":
+            b.add_identifier("eissn", None)
```
richard-jones, line 372: If we're adding new code here, we should be using the current methods, not the deprecated ones.  So this would be `b.pissn` rather than `b.get_one_identifier("pissn")`, and `b.eissn` instead of `b.get_one_identifier("eissn")`

portality/scripts/2882_empty_string_issns.py
```diff
@@ -0,0 +1,52 @@
...
+                }
+            },
+            "query": {
+                "match_all": {}
+            }
+        }
+    }
+}
+
+if __name__ == "__main__":
```
richard-jones, line 24: We could probably run this on test to get an idea of the answer, without merging and deploying to live first.

richard-jones (changes requested):

portality/bll/services/application.py
amdomanska: sure!

portality/scripts/2882_empty_string_issns.py
amdomanska: Agree.

amdomanska: Locally does not return anything.

Steven-Eardley:
```

(doaj) cloo@doaj-new-test-1:~/doaj/src/doaj$ DOAJENV=test python portality/scripts/2882_empty_string_issns.py 
Running in test
Loaded environment config from /home/cloo/doaj/src/doaj/test.cfg
Loaded secrets config from /home/cloo/doaj/src/doaj/app.cfg
Compiling static content
Compiling SASS
Analyzed 15870 journals in DOAJ
```

Steven-Eardley: `out.csv` empty so nothing stopping the release and run on live.

PR #1895, opened 2021-04-28
Correct subject of email for update request

amdomanska:
patch for: https://github.com/DOAJ/doajPM/issues/2570

### DIFF
--- a/doajtest/unit/application_processors/test_application_processor_emails.py
+++ b/doajtest/unit/application_processors/test_application_processor_emails.py
@@ -915,7 +915,7 @@ class TestUpdateRequestReviewEmails(DoajTestCase):
         #   * to the journal contact, informing them of the journal's acceptance
         publisher_template = 'publisher_update_request_accepted.txt'
         publisher_to = re.escape(owner.email)
-        publisher_subject = 'journal accepted'
+        publisher_subject = 'update request accepted'
 
         publisher_email_matched = re.search(email_log_regex % (publisher_template, publisher_to, publisher_subject),
                                             info_stream_contents,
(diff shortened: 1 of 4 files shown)


amdomanska:
Tested with Mailtrap:

Screenshot from 2021-04-28 17-55-18

richard-jones (approved):

portality/formcontext/formcontext.py
```diff
@@ -492,7 +492,11 @@ def _send_application_approved_email(self, journal_title, publisher_name, email,
 
         to = [email]
         fro = app.config.get('SYSTEM_EMAIL_FROM', 'feedback@doaj.org')
-        subject = app.config.get("SERVICE_NAME", "") + " - journal accepted"
+        if update_request:
```
Steven-Eardley, line 495: We don't use formcontext.py any more, it will be removed - changes here aren't needed.

Steven-Eardley (changes requested):
This causes a genuine test failure - we need some updates in doajtest/unit/application_processors/test_application_processor_emails.py

The first test needs updating:
* It uses an update request fixture where it should be a new application
* It doesn't check both update requests and new applications

I assume the fixture was repurposed after this test was written, since it predates update requests.

doajtest/unit/application_processors/test_application_processor_emails.py
```diff
@@ -915,7 +915,7 @@ def test_01_maned_review_emails(self):
         #   * to the journal contact, informing them of the journal's acceptance
         publisher_template = 'publisher_update_request_accepted.txt'
         publisher_to = re.escape(owner.email)
-        publisher_subject = 'journal accepted'
+        publisher_subject = 'update request accepted'
```
amdomanska, line 918: test updated

PR #1897, opened 2021-04-29
prevent 'undefined' being printed where dates are not properly formatted

richard-jones:
A bit of defensive programming for this issue: https://github.com/DOAJ/doajPM/issues/2883


Steven-Eardley (approved):
looks good. now I want some naan

PR #1899, opened 2021-04-29
update a couple of csv questions

richard-jones:
For https://github.com/DOAJ/doajPM/issues/2886


Steven-Eardley (changes requested):
@richard-jones tests will fail due to expected headers in `doajtest/fixtures/v2/journals.py`

PR #1905, opened 2021-04-30
Hotfix/813 ogurl

amdomanska:
For: https://github.com/DOAJ/doajPM/issues/813

I am not sure if that is still the issue. I deployed to the server two versions of the app. One with the `<meta property="og:url" content="https://doaj.org>` and one with: `<meta property="og:url" content={{ request.url }}>`. I tested it with https://developers.facebook.com/tools/debug/ for journal toc page.
In both cases, canonical URL has been correctly scrapped as testdoaj/toc/xxxx-xxxx url (og:url set to doaj.org seemed to be ignored?). 

I'm submitting this PR with requested change as it seems reasonable but it may need more investigation. Or maybe just I need more explanation on what is expected.

### DIFF
--- a/portality/templates/layouts/base.html
+++ b/portality/templates/layouts/base.html
@@ -35,7 +35,7 @@ I'm not sure of the value of it.  Leaving the code snippet here to remind us #}
   <meta name="theme-color" content="#282624">
 
   <!-- OpenGraph -->
-  <meta property="og:url" content={{ request.url }}>
+  <meta property="og:url" content="{{ request.url }}">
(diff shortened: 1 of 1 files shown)


richard-jones (changes requested):
You need quote marks around that `{{request.url}}`

See also my comment about the actual fix for this being the other `og` fields - we'll check whether we need to do any more work here before deploying this.

amdomanska:
> You need quote marks around that `{{request.url}}`
added.

PR #1907, opened 2021-05-04
Add intro to frontmatter

sssoz:
See DOAJ/doajPM#2556


amdomanska (approved):

richard-jones (commented):
I'm a bit uncertain about this approach, though I see why you've done it this way.

We're introducing a requirement to actually include a template at any point in a static page, and perhaps what we should do is have a general approach for that, rather than wedging in an `intro` frontmatter header to fix this current situation.

Can you assign the original issue to me, and I'll look at some alternatives?

sssoz:
@richard-jones sure let’s do that. One thing though: I was hoping to use the intro in the front matter as the page’s `<meta type="description">` as well as the description content for OG & Twitter cards (as part of https://github.com/DOAJ/doajPM/issues/2921); I have seen it done that way usually, as this kind of content is meant to be short & is considered metadata, so it is appropriate for the front matter. 

Though being able to place markdown text at any point on the page would be pretty essential in combination with the above approach.

sssoz:
@richard-jones I’ve just added the tiny bit of text, think that we should just deploy it when we can and revise copy when we have newer, fresher content :)

Steven-Eardley (approved):

PR #1909, opened 2021-05-05
redirect .suggest to public application

amdomanska:
hotfix for: https://github.com/DOAJ/doajPM/issues/2926

Redirects `/suggest` to `/apply`


richard-jones (approved):

Steven-Eardley (changes requested):
I think in general, handling legacy redirects is better suited to the webserver than the app - it's cleaner to simply get rid of old code (and fix if it's been referred to elsewhere within the codebase) than maintain a series of routes that we don't use.

see `deploy/nginx/doaj-redirects.map` to add to that list

Steven-Eardley (approved):

PR #1910, opened 2021-05-05
Hotfix/2923 unlock fix

amdomanska:
hotfix for https://github.com/DOAJ/doajPM/issues/2923

Bug caused by inconsistencies between journal/application/suggestion types

### DIFF
--- a/portality/view/doajservices.py
+++ b/portality/view/doajservices.py
@@ -16,6 +16,11 @@ blueprint = Blueprint('doajservices', __name__)
 @ssl_required
 @write_required()
 def unlock(object_type, object_id):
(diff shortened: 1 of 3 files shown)


richard-jones (approved):

portality/view/doajservices.py
```diff
@@ -16,6 +16,11 @@
 @ssl_required
 @write_required()
 def unlock(object_type, object_id):
+    # change from suggestion to application after redesign, that needs refactoring to make the code consistent
```
Steven-Eardley, line 19: I'm not sure I understand your comment here - you're changing the type to the one we don't use in order to avoid changing the old code?

PR #1911, opened 2021-05-08
Data model improvements

amdomanska:
Work on https://github.com/DOAJ/doajPM/issues/2899

### DIFF
--- a/data_import_settings/article_sample.json
+++ b/data_import_settings/article_sample.json
@@ -5,7 +5,6 @@
 	"max_content_length" : 100000000,
 	"types" : {
 		"account" : {"import" : false, "limit" : -1},
-		"bulk_upload" : {"import" : false, "limit" : -1},
 		"bulk_reapplication" : {"import" : false, "limit" : -1},
 		"background_job" : {"import" : false, "limit" : -1},
 		"lock" : {"import" : false, "limit" : -1},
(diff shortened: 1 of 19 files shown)


amdomanska:
>admin.last_manual_update exists but doesn't appear to be used. There is also last_manual_update. Check that the admin one isn't valid anywhere, and then remove it from the models (I don't believe any data is stored in the index in this field)

I haven't found any applications or journals with `admin.last_manual_update` populated with data nor any usage for this field in the code. Field removed from the `application` model.

amdomanska:
>last_reapplication field is obsolete, I believe

I haven't found any records where this field exists. In the code it is only used multiple times in the `portality/migrate/20180106_1463_ongoing_updates/sync_journals_applications.py`. I'm removing this field from the model.

amdomanska:
>  In the journal CSV there are a variety of ways that we join lists, | ; , - should we try to standardise
https://github.com/DOAJ/doajPM/issues/2899#issuecomment-835829848

amdomanska:
> licence_display is a list which takes a value "Embed" as a single element. This is a hang-over from a late-stage course change on the application form that is patched at the view layer to be a straight True/False, but in the model is not. If possible we should refactor the model to be in line with how this data is entered and viewed

I refactored that in code:
```
"fields" : {
                        "license_display" : {"coerce" : "unicode", "allowed_values" : ["Embed", "Display", "No"]},
```
@Steven-Eardley does it require some changes in es configuration?

Steven-Eardley:
closing this PR as superseded (keeping the branch)

PR #1913, opened 2021-05-21
Changes to diff table

amdomanska:
Issue: https://github.com/DOAJ/doajPM/issues/2824

- [x] Display fields in the same order as it is in the application form
- [x] Display full field names instead of their IDs
Issue: https://github.com/DOAJ/doajPM/issues/2761

- [x] Languages should be written out in full    => ...and countries

Issue: https://github.com/DOAJ/doajPM/issues/2820 (the same commit as for #2824)
- [x] The change table in the UR form should not show these 4 fields (Pissn,e issn, title and alt title).

### DIFF
--- a/portality/lib/formulaic.py
+++ b/portality/lib/formulaic.py
@@ -100,6 +100,8 @@ import csv
 from copy import deepcopy
 from wtforms import Form
 from wtforms.fields.core import UnboundField, FieldList, FormField
+
+from portality import datasets
 from portality.lib import plugin
 from flask import render_template
 import json
@@ -276,6 +278,9 @@ class FormulaicContext(object):
                 else:
                     fieldlist.append(field)
         return fieldlist
+
+    def lang_to_country(self, lang):
+        return datasets.language_for(lang).name
 
     def make_wtform_class(self, fields):
         class TempForm(Form):
--- a/portality/lib/formulaic.py
+++ b/portality/lib/formulaic.py
@@ -266,6 +266,17 @@ class FormulaicContext(object):
     def default_group_template(self):
         return self._definition.get("templates", {}).get("default_group")
 
+    def list_fields_in_order(self):
+        fieldlist = []
+        for fs in self.fieldsets():
+            for field in fs.fields():
+                if field.group_subfields():
+                    for sf in field.group_subfields():
+                        fieldlist.append(sf)
+                else:
+                    fieldlist.append(field)
+        return fieldlist
+
     def make_wtform_class(self, fields):
         class TempForm(Form):
             pass
--- a/portality/lib/formulaic.py
+++ b/portality/lib/formulaic.py
@@ -101,7 +101,6 @@ from copy import deepcopy
 from wtforms import Form
 from wtforms.fields.core import UnboundField, FieldList, FormField
 
-from portality import datasets
 from portality.lib import plugin
 from flask import render_template
 import json
--- a/portality/lib/formulaic.py
+++ b/portality/lib/formulaic.py
@@ -268,12 +268,6 @@ class FormulaicContext(object):
     def default_group_template(self):
         return self._definition.get("templates", {}).get("default_group")
 
-    def language_name_from_language(self, lang):
-        return datasets.name_for_lang(lang)
-
-    def country_name_from_code(self, code):
-        return datasets.get_country_name(code)
-
     def list_fields_in_order(self):
         fieldlist = []
         for fs in self.fieldsets():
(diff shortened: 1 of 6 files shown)


portality/templates/application_form/_value.html
```diff
@@ -0,0 +1,15 @@
...
+    {% if x == None or x == "" or x == [] %}
+        <em>[Nothing]</em>
+    {% elif x is string and x.startswith("http") %}
+        <a href="{{ x }}" target="_blank" rel="noopener" style="word-break: break-all">{{ x }}</a>
+    {% elif x == "y" %}
+        Yes
+    {% elif x == "n" %}
+        No
+    {% elif field_name == "language" or field_name == "publisher_country" or field_name == "institution_country" %}
+        {{ formulaic_context.lang_to_country("EN") }}
```
richard-jones, line 11: Should `EN` be hard-coded here?

portality/templates/application_form/_value.html
```diff
@@ -0,0 +1,15 @@
+{% macro value(x, field_name, formulaic_context) %}
+    {% if x == None or x == "" or x == [] %}
+        <em>[Nothing]</em>
+    {% elif x is string and x.startswith("http") %}
+        <a href="{{ x }}" target="_blank" rel="noopener" style="word-break: break-all">{{ x }}</a>
```
richard-jones, line 5: should embedded style actually be in the SASS?

portality/lib/formulaic.py
```diff
@@ -266,6 +268,20 @@ def default_field_template(self):
...
+            for field in fs.fields():
+                if field.group_subfields():
+                    for sf in field.group_subfields():
+                        fieldlist.append(sf)
+                else:
+                    fieldlist.append(field)
+        return fieldlist
+
+    def lang_to_country(self, lang):
+        return datasets.language_for(lang).name
```
richard-jones, line 283: What happens if the `lang` is not found? Presumably `.name` causes an exception.

Also, are we converting the language to a country, or are we converting the language to the language name?

In any case, this feels to form-content-specific to be in formulaic, which ought to be generic.  As this is just a wrapper for a call to the datasets module, the calling code should just use the datasets module directly, imo.

richard-jones (changes requested):

portality/templates/application_form/_value.html
sssoz: I did that one. This is because we don’t want to `break-all` link styles necessarily. Inlining isn’t always a bad thing when you need to apply a style to a specific element (plus, in certain cases, it is better for performance).

portality/templates/application_form/_value.html
amdomanska: It shouldn't, that should be just value converted

portality/lib/formulaic.py
amdomanska: it actually should convert language to the language name for language field and country code for the country for the country field. I am just wondering whether the template really should cause datasets module directly. It is not really form-specific. It is the same conversion for every form (although used for diff table). Not creating the wrapper for datasets module in formulaic will require passing that module as a param to each form template - not ideal either... What do you think? @richard-jones

Steven-Eardley: I'm mid-way on this - my first instict was that a thin wrapper around datasets wasn't necessary, but I agree that exposing that entire lib to the template is also clunky.

Would it be better to define this as a template filter? (not sure if you can do it on blueprints instead of app) - see https://github.com/DOAJ/doaj/blob/develop/portality/app.py#L182 for examples.

amdomanska: @Steven-Eardley after call with Richard the datasets module is now exposed to jinja templates by setup in core.py. What do you think about it?

portality/lib/formulaic.py
```diff
@@ -100,6 +100,8 @@
 from copy import deepcopy
 from wtforms import Form
 from wtforms.fields.core import UnboundField, FieldList, FormField
+
+from portality import datasets
```
richard-jones, line 104: This is no longer in use, so can be removed

richard-jones (approved):
One minor code cleanup thing remaining, then good to go with me.

portality/lib/formulaic.py
richard-jones: @Steven-Eardley's approach with a template filter is the other way to do this, but I think since we need a few functions from `datasets` making it available as a global to the templating system makes the most sense now.  There are probably a few other places in the templates where we could use this, too.

portality/lib/formulaic.py
amdomanska: done

amdomanska:
@Steven-Eardley it is ready to go

Steven-Eardley (approved):

PR #1914, opened 2021-05-25
Hotfix/2956 crossref xwalk errors

amdomanska:
patch for https://github.com/DOAJ/doajPM/issues/2956
and https://github.com/DOAJ/doajPM/issues/2960

- [x] Author's given name and surname (incorrectly joint)
- [x] Make given name optional
*The comma from between the given name and surname has been removed and now we accept the author with surname only as well*

- [x] Affiliation (absent)
- [x] ORCID (absent)
*Both added.*

- [x] Publication date ( "online" date is taken instead of "print" date)
*Digested the first one provided in the xml in accordance with: https://github.com/DOAJ/doajPM/issues/2956#issuecomment-847748076*


doajtest/unit/test_crosswalks.py
```diff
@@ -117,5 +117,5 @@ def test_06_crossref_article_xml_xwalk(self):
         assert bibjson.get_one_identifier(bibjson.E_ISSN) == "9876-5432", "expected '9876-5432', received: {}".format(bibjson.get_one_identifier(bibjson.E_ISSN))
         assert bibjson.year == "2004", "expected '2004', received: {}".format(bibjson.year)
         assert bibjson.title == "Article 12292005 9:32", "expected 'Article 12292005 9:32', received: {}".format(bibjson.title)
-        assert bibjson.author == [{'name': 'Surname, Bob'}], "expected [{{'name': 'Surname, Bob'}}]', received: {}".format(bibjson.author)
+        assert bibjson.author == [{'name': 'Surname Bob'}], "expected [{{'name': 'Surname Bob'}}]', received: {}".format(bibjson.author)
```
Steven-Eardley, line 120: is this the correct representation of an author's name? "Domańska Aga" rather than "Aga Domańska"?

If we're keeping it surname then given name, the comma would be typical. But I think since our metadata form has one name field and lots of the articles I've looked at have it entered as typically western ordering, Firstname Surname, we may wish to be consistent.

I may have missed a prior discussion about this when the crossref work was originally done, though.

portality/crosswalks/article_crossref_xml.py
```diff
@@ -273,9 +273,16 @@ def crosswalk_article(self, record, journal):
         if contribs is not None:
             for ctb in contribs:
                 if ctb.attrib["contributor_role"] == 'author':
-                    name = _element(ctb, "x:surname", NS)
-                    name = name + ', ' + _element(ctb, "x:given_name", NS)
-                    bibjson.add_author(name, affiliation=None)
+                    e = _element(ctb, "x:surname", NS)
+                    name = e if e else ""
```
Steven-Eardley, line 277: I suppose the schema will prevent the case where surname is empty, so the `if e` ... might not be necessary, but there's nothing wrong with coding defensively

Steven-Eardley (commented):
I'd just like to check we have the intended name ordering before this goes live

doajtest/unit/test_crosswalks.py
amdomanska: I assumed surname first as it is the required piece of data, given name is optional. I think the ordering depends a lot on the country though ;) I am not sure whether we discussed that or not, that is true. @dommitchell I think it is your decision here to make.

portality/crosswalks/article_crossref_xml.py
amdomanska: True, unnecessary. I will clean it up.

doajtest/unit/test_crosswalks.py
dommitchell: We should display them as First name, Last name as we do for the rest of DOAJ article metadata

Steven-Eardley (approved):

PR #1917, opened 2021-05-28
Feature/script tags

amdomanska:
- Script to find any `<script>` in our article's metadata created (passed with tested article)
- Custom validator added to wt forms to check for <script> tags in metadata form.
- Check added to creating and updating the article
- No additional checks for xml uploads not needed as <script> tags are rejected by schemas
- API Unittest added
- 
- NoScriptTag allowed added to front and backend of application form
- Escaping the html chars in full content widget
- warning email send if attempt of sending <script> tag in articles metadata via api has been detected (no need for that for other ways as will be handled by front-end)

image
image
image
image


doajtest/unit/resources/create_article.json
```diff
@@ -0,0 +1,58 @@
+{
```
amdomanska, line 1: file for api tests (just to make our lives slightly easier)

doajtest/unit/resources/create_application.json
```diff
@@ -0,0 +1,120 @@
+{
```
amdomanska, line 1: file for api tests (just to make our lives slightly easier)

portality/api/v2/crud/applications.py
```diff
@@ -58,6 +61,20 @@ def create(cls, data, account, dry_run=False):
             ia = IncomingApplication(data)
         except seamless.SeamlessException as e:
             raise Api400Error(str(e))
+        except dataobj.ScriptTagFoundException as e:
+            jdata = json.dumps(data, indent=4)
```
Steven-Eardley, line 65: does the email sending need to be in a try/catch so we don't lose the response if the email doesn't send?

Steven-Eardley, line 65: at this stage do we have enough useful information in the article data itself or do we need to add more? (account or API key perhaps)

portality/api/v2/data_objects/application.py
```diff
@@ -179,6 +183,17 @@ def custom_validate(self):
         if len(self.data["bibjson"]["keywords"]) > 6:
             raise seamless.SeamlessException("bibjson.keywords may only contain a maximum of 6 keywords")
 
+    def _check_for_script(self, article):
+        for key, value in article.items():
+            if value:
+                if isinstance(value, dict):
+                    if self._check_for_script(value):
```
Steven-Eardley, line 190: :heart: recursion

Steven-Eardley: Although you could `return` the internal function call rather than another boolean check. 

Also, what happens if it's a list?

Steven-Eardley (approved):
This is absolutely superb work, very comprehensive.

portality/api/v2/data_objects/application.py
amdomanska: > Although you could return the internal function call rather than another boolean check.
not really. If the internal function call returns false we still want to proceed with the rest of the json. So we want to return it only if it is true.

amdomanska: > Also, what happens if it's a list? 

I cannot think of the data in the list that could be risky... But I can add just in case we will change the data model in the future, that's good idea.

portality/api/v2/crud/applications.py
amdomanska: I added the account information to the data in the email.

portality/api/v2/data_objects/common.py
```diff
@@ -0,0 +1,16 @@
+def _check_for_script(data):
+    print(data)
```
Steven-Eardley, line 2: I don't think we want print statements in the low level API

amdomanska: Nope. This is overlooked from development.

Steven-Eardley (approved):

PR #1920, opened 2021-06-07
add callback to animate method, animate html not scrollselector

amdomanska:
Wrong branch name, should have number 2880
patch for https://github.com/DOAJ/doajPM/issues/2880

`$('html')` should be scrolled instead of scroll selector
page should be rendered only after scroll is finished


richard-jones (changes requested):
Lets go through this together, and look at alternative ways to approach

amdomanska:
Taken care of by Richard's pull request

PR #1923, opened 2021-06-09
All menu items should be focusable

amdomanska:
patch for https://github.com/DOAJ/doajPM/issues/2971
change span to anchor to ensure focusability


portality/templates/includes/header-secondary-navigation.html
```diff
@@ -6,10 +6,10 @@
 {% endif %}
 
 <li class="dropdown {{ extraClasses }}">
-    <span class="dropdown__trigger nav-bar__link" data-toggle="dropdown" id="menu-{{ selected_nav.id }}" role="button" aria-haspopup="true" aria-expanded="false">
+    <a href="" class="dropdown__trigger nav-bar__link" data-toggle="dropdown" id="menu-{{ selected_nav.id }}" role="button" aria-haspopup="true" aria-expanded="false">
```
sssoz, line 9: We shouldn’t be using anchors that aren’t a link / don’t have an `href`. Another way to do it semantically and accessibly could be to use the `<button>` element, though some style changes would be required so that it doesn’t get displayed as a UI button (i.e. only elements using `.button` should be styled, not the HTML element). This way, you can tab through all menu items in sequential order.

sssoz: Also just remembered that I had changed the anchor to a `<span>` to avoid the empty `href` originally. 
 
This is because `<a href="...">` should be taking the user to another page / elsewhere whereas `<button>s` indicates an action. 

It’s a bit complicated though because this is still very debatable. An anchor with an empty `href` is still valid HTML but not necessarily correct. 

The `<button>` article on MDN also states:
> If your buttons are not for submitting form data to a server, **be sure to set their type attribute to button**. Otherwise they will try to submit form data and to load the (nonexistent) response, possibly destroying the current state of the document.

We can discuss this on Monday if you’d like so that we can simply agree on a common (accessible) approach going forward .

amdomanska: I did not plan to release it before talking to you. I just tried to find the way to avoid tabindex and the anchor is the way that you did this in the primary nav. But that is why I asked you to talk to you on Monday - cause I am not a big fun of that solution as well.

amdomanska: @sssoz I think we need to tackle the issue now. 

It looks like w3 actually accepts anchors with empty href as a good pattern for that use case and operates with `tabindex` property.

https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-1/menubar-1.html

amdomanska:
In line with discussion with @sssoz on Slack (start here: https://doajplus.slack.com/archives/CMS19QQHZ/p1624900023040200) all menu items are changed to anchores and appropriate aria-label and role attributes are added, and `href` attributes set to `#`. 

To meet all accessibility recommendations from the document linked in my previous comment we should additionally implement appropriate behaviour on arrows keys click (arrow down should move focus to the first/next submenu item and arrow up to the last/previous item).

portality/templates/includes/header.html
```diff
@@ -11,11 +11,10 @@ <h2 class="sr-only">The Directory of Open Access Journals</h2>
                     {% include "includes/svg/logo.svg" %}
                 </a>
             </div>
-            <nav class="col-xs-8 col-sm-6">
+            <nav class="col-xs-8 col-sm-6" aria-label="Main actions">
```
sssoz, line 14: The aria label is repeated here

portality/templates/includes/header.html
```diff
@@ -41,12 +40,12 @@ <h2 class="sr-only">Main actions</h2>
         {% set extraClass="secondary-nav--precedes-highlight" %}
     {% endif %}
 {% endif %}
-<nav class="secondary-nav {{ extraClass }}">
+<nav class="secondary-nav {{ extraClass }}" aria-label="Secondary Nav">
```
sssoz, line 43: Same here; the heading associated for the list indicates that it’s the Menu, so we should keep them that way (Secondary nav isn’t a helpful label for the user)

portality/templates/includes/header-primary-navigation.html
```diff
@@ -17,8 +17,8 @@
                 {% set urlAttrs = "" %}
             {% endif %}
 
-            <li>
-                <a {{ urlAttrs }} href="{{ url }}" class="dropdown__link">
+            <li role="none">
```
sssoz, line 20: We should keep list items as part of the accessibility tree.

portality/templates/includes/header-primary-navigation.html
```diff
@@ -1,14 +1,14 @@
 {% set selected_nav = data.nav.entries|selectattr("id", "equalto", section_id)|first %}
-<li class="dropdown primary-nav__item_desktop">
-    <a href="{{ url_for(selected_nav.route) }}" data-toggle="dropdown" class="dropdown__trigger nav-bar__link" id="action-menu-{{ selected_nav.id }}" role="button" aria-haspopup="true" aria-expanded="false">
+<li class="dropdown primary-nav__item_desktop" role="none">
```
sssoz, line 2: Idem re: a11y tree

portality/templates/includes/header-primary-navigation.html
amdomanska: `role="none"` has been added here in line with the W3 pattern. Why is it not correct?

portality/templates/includes/header.html
amdomanska: Yes, that is intended. Again: in line with the W3 pattern.

portality/templates/includes/header.html
amdomanska: Yes, that should state "Secondary Menu".

portality/templates/includes/header-primary-navigation.html
amdomanska: It is used in the nav pattern this way on a11y tree here as well:

https://www.digitala11y.com/none-role/

sssoz: Yes, I’ve seen this as well, but we might consider that keeping the unordered list item is something useful for our particular case, in terms of semantics

portality/templates/includes/header-primary-navigation.html
sssoz: (same comment re: semantics of lists)

portality/templates/includes/header.html
sssoz: I mean that there is already a visually-hidden heading (`<h2 class="sr-only">Main actions</h2>`) which has the same function as an `aria-label` -- we don’t need both. The `sr-only` class is meant to include textual content as a guide for screen readers only.

portality/templates/includes/header.html
sssoz: It’s more for consistency here; the corresponding heading (`<h2 class="secondary-nav__header">Menu</h2>`) says _Menu_ only -- I chose this as this was the main menu content (and _Main actions_ seemed more semantically correct for the top nav items, "Support" and "Apply" being action verbs). The shorter the label, the better.

amdomanska:
@sssoz do you think we should implement navigation with arrow keys as an accessibility feature for these menus?

sssoz:
> @sssoz do you think we should implement navigation with arrow keys as an accessibility feature for these menus?

I hadn’t considered it, but it could be good. I haven’t implemented anything like this before though.

sssoz (approved):

amdomanska:
Implementing it is fairly easy javascript issue, not a problem other than planning some time for it. The w3 site mentions the arrows as a default way of navigating through navbar and I think it is fairly common to have that done this way. It is really only about the convenience for our users.

amdomanska:
I resolved some conflicts between mine and @sssoz 's code. To be sure that everything is ok I would be grateful if we could put that on test for @sssoz to review

sssoz:
@amdomanska I’ve tested the branch locally, looks like the `sr-only` content and the hamburger menu are appearing on desktop when they should be visually hidden 

Screenshot 2021-07-13 at 17 51 16

sssoz:
Ahh I see there is some duplicated HTML. Do you want to remove that? The first instances of both are the correct ones.

amdomanska:
@sssoz thank you! Now it should be fine.

sssoz:
You’ve now removed the Menu label (which is a menu item on smaller viewports): 

Screenshot 2021-07-13 at 18 50 41

sssoz:
The secondary nav also now has lost its visually-hidden heading and is now an unnamed `<nav>` -- there is no need to remove the `<h2>` headings with the screen reader classes, `sr-only`. 

Screenshot 2021-07-13 at 18 52 08

portality/templates/includes/header.html
```diff
@@ -42,7 +40,7 @@ <h2 class="label">Menu</h2> <span data-feather="menu" aria-hidden="true"></span>
 {% endif %}
 <nav class="secondary-nav {{ extraClass }}">
     <div class="container">
-        <ul class="nav-bar secondary-nav__menu">
```
sssoz, line 45: Did not notice when it happened but the `<h2 class="sr-only">Secondary actions</h2>` (or something along those lines) belongs here @amdomanska

amdomanska: @sssoz Yep, that's the problem of this merge. It wasn't very straightforward that's why I needed another pair of eyes. Thank you!

portality/templates/includes/header.html
```diff
@@ -40,11 +39,7 @@ <h2 class="label">Menu</h2> <span data-feather="menu" aria-hidden="true"></span>
     {% endif %}
 {% endif %}
 <nav class="secondary-nav {{ extraClass }}">
-<h2 class="sr-only">Menu</h2>
     <div class="container">
-        <div class="secondary-nav__menu-toggle" aria-label="Click to open menu" role="button">
```
sssoz: @amdomanska This is the one that needed to stay

portality/templates/includes/header.html
amdomanska: Interestengly that is not there even on live...

amdomanska:
ok, `sr-only` headers restored and mobile menu is back. How does it look now @sssoz ?

image

portality/templates/includes/header.html
sssoz: Yes, that is why I mentioned that I did not notice when it was removed

sssoz: (Fixed to "Secondary actions") in 58700ee534e236dcd15fe5f67fb66299aca49e51

sssoz (approved):
Fixed up a few minor copy and markup issues

PR #1926, opened 2021-06-14
Feature/2957 heloise

amdomanska:
https://github.com/DOAJ/doajPM/issues/2957


portality/scripts/out.csv
```diff
@@ -0,0 +1,19 @@
...
+J,o,u,r,n,a,l, ,5,d,a,3,a,4,1,f,6,b,d,0,4,6,b,e,9,b,f,e,0,f,8,8,d,4,c,d,7,d,4,e, ,m,o,d,i,f,i,e,d, ,a,n,d, ,s,a,v,e,d,.
+J,o,u,r,n,a,l, ,e,d,5,d,0,3,2,0,a,7,f,6,4,3,c,0,a,e,9,8,8,3,1,9,c,6,8,7,1,d,d,4, ,m,o,d,i,f,i,e,d, ,a,n,d, ,s,a,v,e,d,.
+J,o,u,r,n,a,l, ,e,a,c,c,e,f,d,5,d,4,1,d,4,5,3,5,a,3,5,6,a,5,0,5,3,0,2,1,a,b,c,1, ,m,o,d,i,f,i,e,d, ,a,n,d, ,s,a,v,e,d,.
+J,o,u,r,n,a,l, ,9,7,c,6,9,b,a,5,9,d,3,7,4,8,d,f,9,b,8,6,5,2,e,2,2,5,4,0,1,d,c,5, ,m,o,d,i,f,i,e,d, ,a,n,d, ,s,a,v,e,d,.
+J,o,u,r,n,a,l, ,c,8,c,7,b,b,7,0,a,0,b,4,4,d,b,d,8,5,7,7,3,a,1,b,8,1,c,d,a,3,e,e, ,m,o,d,i,f,i,e,d, ,a,n,d, ,s,a,v,e,d,.
+J,o,u,r,n,a,l, ,6,5,b,7,d,4,4,9,f,a,7,c,4,4,0,4,9,2,6,f,8,2,d,2,a,6,2,0,6,0,a,4, ,m,o,d,i,f,i,e,d, ,a,n,d, ,s,a,v,e,d,.
+J,o,u,r,n,a,l, ,1,c,9,e,f,7,b,4,7,a,0,d,4,a,9,f,b,3,b,c,5,3,d,b,b,5,1,4,a,3,9,b, ,m,o,d,i,f,i,e,d, ,a,n,d, ,s,a,v,e,d,.
+J,o,u,r,n,a,l, ,e,a,f,4,2,8,5,8,f,4,0,1,4,f,0,1,9,5,f,1,c,1,5,8,4,1,2,b,c,a,e,9, ,m,o,d,i,f,i,e,d, ,a,n,d, ,s,a,v,e,d,.
+J,o,u,r,n,a,l, ,3,5,e,7,f,1,3,4,3,c,7,3,4,f,0,9,b,6,4,c,1,6,7,8,a,7,c,d,4,a,e,2, ,m,o,d,i,f,i,e,d, ,a,n,d, ,s,a,v,e,d,.
```
richard-jones, line 19: Something weird about this CSV format :)

portality/scripts/remove_heloise_policy.py
```diff
@@ -0,0 +1,38 @@
+import csv
```
richard-jones, line 1: Suggest that this script lives in the `migration` directory, as it's a one-time affair for a specific change.  Then the resources file related to it can be wrapped up in that directory as well, and they won't sit in the `scripts` directory never being used again

portality/scripts/resources/journalID-for-heloise-removal.txt
```diff
@@ -0,0 +1,43 @@
+85d41bcb878c419d8cb7b3387fa2987d
```
richard-jones, line 1: See previous comment about where to put this file

portality/scripts/remove_heloise_policy.py
```diff
@@ -0,0 +1,38 @@
...
+            print(j.bibjson().deposit_policy)
+            if j:
+                bib = j.bibjson()
+                deposit_policy = j.bibjson().deposit_policy
+                if 'Héloïse' in deposit_policy:
+                    deposit_policy.remove('Héloïse')
+                if len(deposit_policy) == 0:
+                    bib.has_deposit_policy = False
+                    j.save(blocking=True)
+                    writer.writerow("Journal {} modified and saved.".format(j.id))
```
richard-jones, line 37: Ah yeah, here's your problem with the output file format.  You need to put the row as a list:

```
writer.writerow(["Journal {} modified and saved.".format(j.id)])
```

portality/scripts/remove_heloise_policy.py
```diff
@@ -0,0 +1,38 @@
...
+            j = Journal.pull(i.rstrip('\n'))
+            print(j.bibjson().deposit_policy)
+            if j:
+                bib = j.bibjson()
+                deposit_policy = j.bibjson().deposit_policy
+                if 'Héloïse' in deposit_policy:
+                    deposit_policy.remove('Héloïse')
+                if len(deposit_policy) == 0:
+                    bib.has_deposit_policy = False
+                    j.save(blocking=True)
```
richard-jones, line 36: It doesn't matter particularly, but why blocking save?

richard-jones (changes requested):
Looks good to me, in terms of code.  Propose to move the scripts to their own migration directory instead of putting them in the scripts directory.

portality/scripts/out.csv
amdomanska: Oh boy.....

portality/scripts/remove_heloise_policy.py
amdomanska: Yep, I realised right after your comment :)

richard-jones (approved):
Looks good to me.  Ready for merge and then run

Steven-Eardley:
Am I right that the migration should be merged & run first, so that the models don't become invalid and give us errors? Alternatively I'll do it super quickly during lower usage period and hope we don't get many requests for those journals

Steven-Eardley:
ah, does `allow_fail=True` protecut us from that DataStructureException?

Steven-Eardley (approved):

richard-jones:
@Steven-Eardley `allow_fail=True` does allow these to fail, so it would be fine, but in fact this code is also obsolete as the DataObj is only used for Articles now, and the new data model using Seamless doesn't rely on that coerce function.  It'll probably all get cleaned up as I do my code review for that other issue.

PR #1927, opened 2021-06-16
rename the script and add applications check

amdomanska:
Updating the script html tag script to check applications too.

### DIFF
--- a/portality/scripts/scrip_tags_check.py
+++ b/portality/scripts/scrip_tags_check.py
@@ -55,7 +55,6 @@ if __name__ == "__main__":
             for key, value in results:
                 writer.writerow([app.id, app.bibjson().title, key, value])
 
-        #ALL_APPLICATION IS FINE
         for j in esprit.tasks.scroll(conn, ipt_prefix(models.Journal.__type__), q=ALL, page_size=100, keepalive='5m'):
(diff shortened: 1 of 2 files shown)


portality/scripts/scrip_tags_check.py
```diff
@@ -44,10 +41,23 @@ def find_nested(substr, d, result):
...
 
             for key, value in results:
                 writer.writerow([article.id, article.bibjson().title, key, value])
+
+        for j in esprit.tasks.scroll(conn, ipt_prefix(models.Application.__type__), q=ALL, page_size=100, keepalive='5m'):
+            app = models.Application(_source=j)
+            results = find_nested("<script>", app.__dict__, [])
+
+            for key, value in results:
+                writer.writerow([app.id, app.bibjson().title, key, value])
```
Steven-Eardley, line 56: Might need to add 'application', to each row so that we know which IDs are for which type, since they are stored in different indexes.

amdomanska: Done

Steven-Eardley (approved):

Steven-Eardley:
even with the correct target it seems that `git flow hotfix finish` will mean this branch shows **closed** rather than **merged**

PR #1928, opened 2021-06-18
support html tags for crossref upload

amdomanska:
https://github.com/DOAJ/doajPM/issues/2953


doajtest/unit/test_tasks_ingestCrossrefArticles.py
```diff
@@ -1853,4 +1858,92 @@ def test_50_3_issns(self):
...
+        bib = art.bibjson()
+        assert bib.title == "This article has <ns0:i>unescaped</ns0:i> and &lt;i&gt;escaped&lt;/i&gt; html tags", "expected: 'This article has <i>unescaped</i> and &lt;i&gt;unexcaped&lt;/i&gt; html tags', received: {}".format(bib.title)
+
+        with open(ARTICLES, "r") as f:
+            data = f.read()
+        new_data = data.replace("This article has <i>unescaped</i> and &lt;i&gt;escaped&lt;/i&gt; html tags", "{REPLACEME}")
+        with open(ARTICLES, "w") as f:
+            f.write(new_data)
+
+    def test_51_html_tags_in_title_attr(self):
```
richard-jones, line 1925: Not that it matters terribly, but your test numbers are both `51`

portality/crosswalks/article_crossref_xml.py
```diff
@@ -304,9 +305,19 @@ def crosswalk_article(self, record, journal):
...
+        text = string[start:end]
+        return text if text else None
+    else:
+        return None
+
+# def _element(xml, field, namepsace):
+#     el = xml.find(field, namepsace)
+#     if el is not None and el.text is not None and el.text != "":
+#         return el.text.strip()
+#     return None
```
richard-jones, line 323: Dead code cleanup

portality/tasks/ingestarticles.py
```diff
@@ -286,12 +286,13 @@ def _process(self, file_upload):
...
         ingest_exception = False
         result = {}
+        ids = []
         try:
             with open(path) as handle:
                 articles = xwalk.crosswalk_file(handle, add_journal_info=False) # don't import the journal info, as we haven't validated ownership of the ISSNs in the article yet
                 for article in articles:
                     article.set_upload_id(file_upload.id)
-                result = articleService.batch_create_articles(articles, account, add_journal_info=True)
+                result, ids = articleService.batch_create_articles(articles, account, add_journal_info=True)
```
richard-jones, line 295: The article IDs are surely added to each article once they are saved?  You don't need to return the ids, I don't think.  If we do have to change the return type of the `batch_create_articles` and the `create_articles` methods, then we should talk about it in a bit more detail first, as the aim of the BLL is to provide a clean, consistent interface to the core, and this response type change feels hacked in.

portality/bll/services/article.py
```diff
@@ -50,10 +50,11 @@ def batch_create_articles(self, articles, account, duplicate_check=True, merge_d
         all_unowned = set()
         all_unmatched = set()
 
+        ids = []
 
         for article in articles:
             try:
-                result = self.create_article(article, account,
+                result, id = self.create_article(article, account,
```
richard-jones, line 57: See comment on below about the `batch_create_article` method.  I don't think this change is necessary, but if it is I'd want to talk through the implementation of a change to the BLL's method signature, as I don't think this is the cleanest approach.

richard-jones (changes requested):
There's a change to the BLL method signatures, which may break other things, and will probably break tests, and we should look at whether it's necessary and if so if there's a better way to approach it.

dommitchell:
@amdomanska @richard-jones can I just confirm that what we are seeing here now is as expected? https://bit.ly/3mjLnrm

with italics tags appearing in the titles?

PR #1929, opened 2021-06-18
Feature/2677 review and map

richard-jones:
Another periodic merge down for this branch please!

Also, to run the featuremap script on this branch, you'll need to check out and install the latest version of that library https://github.com/CottageLabs/FeatureMap


Steven-Eardley (approved):
All looks good

PR #1930, opened 2021-06-21
Added 'clockss' role

RK206:
Added a role 'cloockss' for the user to enable clockss feature


richard-jones (approved):
All good.

Before merging this, let me check with @dommitchell that we're ok to merge this change before any of the other bits of the project are finished.  It won't make a difference to functionality, but it may be briefly confusing for admins.

dommitchell:
Thanks  @richard-jones It can be merged. I shall alert the Admin users.

dommitchell:
Thanks @richard-jones - it can be merged *but* can we call it something other than 'clockss'? CLOCKSS is just the first agency to be in this project; others will soon come along. Can we rename the role: 'preservation' please?

RK206:
Chaged role name from 'clockss' to 'preservation'

PR #1932, opened 2021-06-23
Hotfix/2626 next button

amdomanska:
1. Makes sure that <li> element in questions with radio buttons is focusable so that it can be focused instead of hidden input
2. fixes jquery selector for `nextBtn` in showTab

### DIFF
--- a/portality/forms/application_forms.py
+++ b/portality/forms/application_forms.py
@@ -2719,9 +2719,9 @@ class ListWidgetWithSubfields(object):
         html = ['<%s %s>' % (self.html_tag, html_params(**kwargs))]
         for subfield in field:
             if self.prefix_label:
-                html.append('<li>%s %s' % (subfield.label, subfield(**kwargs)))
+                html.append('<li tabindex=0>%s %s' % (subfield.label, subfield(**kwargs)))
             else:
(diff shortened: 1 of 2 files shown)


portality/forms/application_forms.py
```diff
@@ -2719,9 +2719,9 @@ def __call__(self, field, **kwargs):
         html = ['<%s %s>' % (self.html_tag, html_params(**kwargs))]
         for subfield in field:
             if self.prefix_label:
-                html.append('<li>%s %s' % (subfield.label, subfield(**kwargs)))
+                html.append('<li tabindex=0>%s %s' % (subfield.label, subfield(**kwargs)))
```
amdomanska, line 2722: @richard-jones I don't think this is the best place to put that but I cannot figure out what would better place be. IT does not feel like a new widget for formulaic either... What do you think?

richard-jones: I think you'll have to talk me through this one, I'm not sure what bug is being fixed, and what the trade-offs here are.

amdomanska: This line will fix https://github.com/DOAJ/doajPM/issues/2993

richard-jones (approved):
ok, I think this looks good, now I understand what the `tabindex` thing does here

PR #1933, opened 2021-06-24
Initial commit for preservation model

RK206:
Created intial model for preservation

### DIFF
--- a/portality/models/preservation.py
+++ b/portality/models/preservation.py
@@ -1,0 +1,57 @@
+from portality.dao import DomainObject
+from datetime import datetime
+
+
+class Preserve(DomainObject):
+    """Model class for preservation feature"""
+    __type__ = "preserve"
+
+    @property
+    def status(self):
+        return self.data.get("status")
+
+    @status.setter
+    def status(self, value):
+        self.data["status"] = value
+
+    @property
+    def filename(self):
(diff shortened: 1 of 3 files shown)


portality/models/preservation.py
```diff
@@ -0,0 +1,57 @@
...
+        self.status = status
+
+    def validated(self):
+        self.status = "validated"
+
+    def pending(self):
+        self.status = "pending"
+
+    def uploaded_to_ia(self):
+        self.status = "Uploaded to IA"
```
richard-jones, line 51: I suggest here we just have a simple status like "uploaded", and then when we display the status to the user the template would be responsible for displaying a meaningful message to the user.

richard-jones (changes requested):
A very minor status code change recommendation, otherwise this looks good to me.

I think we can merge this one down to develop, as it doesn't change anything about the current app.  

At some point we're going to make a change that will change the actual user experience, and then we'll probably need to start staging those changes in a branch that gets merged to develop all in one go when we're ready for release.

portality/models/preservation.py
RK206: Agree. Updated the status as suggested

richard-jones (approved):

Steven-Eardley (approved):

PR #1935, opened 2021-07-02
Feature/2996 prevent the same issns

amdomanska:
https://github.com/DOAJ/doajPM/issues/2996#issuecomment-872412684

https://docs.google.com/spreadsheets/d/1WmfwyusZa0apNYvawIWXeTrBSrAIVC4u8InRSz4xHGI/edit#gid=2143003155

UTs run locally all ok.

### DIFF
--- a/portality/ui/messages.py
+++ b/portality/ui/messages.py
@@ -55,6 +55,8 @@ class Messages(object):
     EXCEPTION_ARTICLE_OVERRIDE = "Cannot update the article. An article with this URL and DOI already exists. If you are sure you want to replace it please delete it and then re-create it."
     EXCEPTION_SCRIPT_TAG_FOUND = "Metadata contains <script> tag. Please remove any <script> tags from your metadata and try again."
 
+    EXCEPTION_IDENTICAL_PISSN_AND_EISSN = "PISSN and EISSN are identical and equal {value}. They must be different"
+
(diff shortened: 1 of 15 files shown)


portality/ui/messages.py
```diff
@@ -55,6 +55,8 @@ class Messages(object):
     EXCEPTION_ARTICLE_OVERRIDE = "Cannot update the article. An article with this URL and DOI already exists. If you are sure you want to replace it please delete it and then re-create it."
     EXCEPTION_SCRIPT_TAG_FOUND = "Metadata contains <script> tag. Please remove any <script> tags from your metadata and try again."
 
+    EXCEPTION_IDENTICAL_PISSN_AND_EISSN = "PISSN and EISSN are identical and equal {value}. They must be different"
```
richard-jones, line 58: Suggest slightly alternative wording to match up with other uses of ISSN in the UI: `The Print and Online ISSNs supplied are identical. If you supply 2 ISSNs they must be different.`

richard-jones (changes requested):
The changes are incredibly minor, just in the text for the UI.  Otherwise, all ready to go from my perspective.

We did briefly talk about the idea of embedding the issn check into the `create_article` method of the BLL, which would act as an absolute back-stop to prevent any such articles getting into the system.  Do you think it's worth putting that in at this stage?

amdomanska:
Move to: DOAJ/doaj#1939

PR #1937, opened 2021-07-05
Feature/clockss upload

RK206:
Added few methods to save the file into temp directory uploaded by the user.


richard-jones (approved):
This looks fine, it will be easier to see how it all fits together once it's more integrated into the app

PR #1940, opened 2021-07-06
Feature/2988 issn validations

amdomanska:
Fix for issues: DOAJ/doajPM#3008 (comment)

Merge 2 issues together as changes are in the same places and that prevents the merge conflicts.

Requires functional tests.

https://docs.google.com/spreadsheets/d/1WmfwyusZa0apNYvawIWXeTrBSrAIVC4u8InRSz4xHGI/edit#gid=0
script


portality/bll/services/article.py
```diff
@@ -141,6 +141,24 @@ def _prepare_update_publisher(self, article, duplicate, merge_duplicate, account
                 raise exceptions.DuplicateArticleException()
         return is_update
 
+    # here we should have the final point of validation for all incoming articles
+    def _sanity_check(self, article):
```
richard-jones, line 145: There's already a method in the `create_article` function called `is_acceptable` which does some validation on the object, so I'd say this logic belongs there instead of a new function.

Also, now we're changing the capabilities of the bll, we need to update the bll test, which I can help you with.

richard-jones (changes requested):

richard-jones (approved):
Looks good, I think this is ready to go

amdomanska:
New issue created putting together #2998 and #2996. Ready for tests, requires Functional Testing

doajtest/unit/apiv2_tests/test_api_crud_article.py
```diff
@@ -721,3 +721,38 @@ def test_15_no_redirects(self):
...
+        journal.set_owner(account.id)
+        journal.save(blocking=True)
+
+        data = ArticleFixtureFactory.make_article_source(pissn="1234-5678", eissn="1234-5678")
+
+        with self.assertRaises(Api400Error):
+            ArticlesCrudApi.create(data, account)
+
+    def test_17_3_issns(self):
+        """ Check we get an error when article with 2 the same issns is provided. """
```
Steven-Eardley, line 743: very minor - this description should stay something about 3 ISSNs

portality/api/v2/crud/articles.py
```diff
@@ -83,6 +84,9 @@ def create(cls, data, account):
             raise Api400Error("; ".join(e.errors))
         except DuplicateArticleException as e:
             raise Api403Error(str(e))
+        except IngestException as e:
+            raise Api400Error(str(e) if str(e) else e.message)
```
Steven-Eardley, line 88: this is a distressing mix of python2 and python3 exceptions - but I guess it's ok if that's where we're at.

portality/bll/services/article.py
```diff
@@ -175,12 +193,18 @@ def create_article(self, article, account, duplicate_check=True, merge_duplicate
 
         # quickly validate that the article is acceptable - it must have a DOI and/or a fulltext
         # this raises an exception if the article is not acceptable, containing all the relevant validation details
-        self.is_acceptable(article)
+
+        try:
+            self.is_acceptable(article)
+        except Exception as e:
+            raise e
```
Steven-Eardley, line 200: isn't this equivalent to what it replaced? (apart from it's clearer that an exception is expected I suppose)

portality/scripts/articles_identical_issns.py
```diff
@@ -0,0 +1,42 @@
...
+if __name__ == "__main__":
+
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-o", "--out", help="output file path")
+    args = parser.parse_args()
+
+    if not args.out:
+        print("Please specify an output file path with the -o option")
```
Steven-Eardley, line 24: if the script no longer requires an output file specified, add it as a default in `add_argument` and don't ask the user for it when it's missing

Steven-Eardley (changes requested):

portality/bll/services/article.py
amdomanska: Yes, it is. But I added that for clarity exactly. The `create_article` method is the "main" one and while looking at that we (or at least I) do not necessary check every single helper method. So I guess it is good no know from its level that it may raise an error.

portality/scripts/articles_identical_issns.py
amdomanska: oh no, it's other way round. I forgot to remove the "out.csv" output that I used for debugging and uncomment the exit.

portality/api/v2/crud/articles.py
amdomanska: I don't like this part neither. But for some reason the Ingest Exeption that is raised here returns `' '` for `str(e)` but has `message` attribute.

Steven-Eardley (approved):

PR #1941, opened 2021-07-06
Trip trailing whitespaces in journals' and articles' metadata

amdomanska:
script for: https://github.com/DOAJ/doajPM/issues/3002


richard-jones (changes requested):
Is this just building a report at this stage?  If so, that's fine.  When we do the migration, we should use the `upgrade.py` script to drive this, as it will do the batching and updating for us.

amdomanska:
As it is actually migration script I refactored it and it works with upgrade.py now. Tested locally fine.

portality/migrate/3002_remove_trailing_whitespaces/operations.py
```diff
@@ -0,0 +1,18 @@
...
+            return remove_trailing_whitespaces_rec(value)
+        if isinstance(value, str):
+            d[key] = value.strip()
+
+def remove_trailing_whitespaces(model):
+    with open("out.csv", "w", encoding="utf-8") as f:
+        writer = csv.writer(f)
+        writer.writerow(["es_type", "ID", "key", "old value", "new value"])
+
+        remove_trailing_whitespaces_rec(model.__dict__)
```
Steven-Eardley, line 16: Does this actually write anything to the the CSV?

Steven-Eardley (changes requested):

portality/migrate/3002_remove_trailing_whitespaces/operations.py
amdomanska: No, it doesn't :/ Corrected.

portality/migrate/3002_remove_trailing_whitespaces/operations.py
```diff
@@ -0,0 +1,15 @@
...
+def remove_trailing_whitespaces_rec(d):
+    for key, value in d.items():
+        if isinstance(value, dict):
+            return remove_trailing_whitespaces_rec(value)
+        if isinstance(value, str):
+            d[key] = value.strip()
+
+
+def remove_trailing_whitespaces(model):
+    remove_trailing_whitespaces_rec(model.__dict__)
```
richard-jones, line 14: Using `__dict__` isn't right here, that could have some unusual effects.

The better and faster way to do this would be to set `init_with_model:false` in `migrate.json` so that the function just gets given the raw dict of the record in the first case.  Then you can get rid of the "unwrapper" function on line 13.

You also need to `return` the dict from the function, as that's what the upgrade script is expecting.

portality/migrate/3002_remove_trailing_whitespaces/migrate.json
```diff
@@ -0,0 +1,20 @@
+{
+"types": [
+		{
+			"type" : "article",
+			"init_with_model" : true,
```
richard-jones, line 5: See note below

portality/migrate/3002_remove_trailing_whitespaces/migrate.json
```diff
@@ -0,0 +1,20 @@
...
+			"keepalive" : "20m",
+			"functions" : [
+				"portality.migrate.3002_remove_trailing_whitespaces.operations.remove_trailing_whitespaces"
+			]
+		},
+		{
+			"type": "journal",
+			"init_with_model": true,
```
richard-jones, line 13: See note below

richard-jones (changes requested):

richard-jones (approved):

Steven-Eardley (approved):

Steven-Eardley (changes requested):
Just came back to this today and saw my comments from last week's review were pending

PR #1942, opened 2021-07-14
Feature/2952 sitemap to s3

richard-jones:
For https://github.com/DOAJ/doajPM/issues/2952

This PR moves the sitemap into the standard storage layer, meaning that in dev it works over the local store, and in prod will work using S3.

When this is deployed, for a short period the sitemap will not be retrievable from the application.  So as soon as we deploy this it may be worth running

`portality/scripts/sitemap.py`

This will generate and publish the sitemap in the new storage location.  If we do this straight away then users shouldn't notice the difference if cloudflare continues to serve the cached sitemap file for a short while.


doajtest/unit/test_bll_site_sitemap.py
```diff
@@ -0,0 +1,185 @@
...
+
+        ###########################################################
+        # Execution
+
+        if raises is not None:
+            with self.assertRaises(raises):
+                self.svc.sitemap(prune)
+
+                tempFiles = self.tmpStore.list(self.container_id)
+                assert len(tempFiles) == 0
```
amdomanska, line 119: Could we please start adding the messages to test assertion:
`assert len(tempFiles) == 0, "expected 0, received {}".format(len(tempFiles))`

This makes debugging tests much more effective.

doajtest/unit/test_bll_site_sitemap.py
```diff
@@ -0,0 +1,185 @@
...
+
+            # Check the results
+            ################################
+            sitemap_info = models.cache.Cache.get_latest_sitemap()
+            assert sitemap_info.get("filename") == url
+
+            filenames = self.localStore.list(self.container_id)
+            if prune:
+                assert len(filenames) == 2
+                assert "sitemap__doaj_20180101_0000_utf8.xml" not in filenames
```
amdomanska, line 132: Here the message added to the assertion probably should be:
`assert "sitemap__doaj_20180101_0000_utf8.xml" not in filenames, "'sitemap__doaj_20180101_0000_utf8.xml' not found in filenames, filenames found: {}".format(filenames)`

amdomanska (changes requested):

Steven-Eardley (approved):

doajtest/unit/test_bll_site_sitemap.py
Steven-Eardley: I didn't like how long this was so I didn't do it. Silly reason? Maybe.

PR #1951, opened 2021-07-19
Allow scrolling in menu on mobile

sssoz:
Need to test it out with various people, not sure if this is the right approach. The big downside to doing this is that, though it allows people to "scroll" down the menu on mobile, it will also require them to go back to the top of the page to tap the "MENU" item to close it again. This isn’t clear, nor intuitive. 

We could ensure that the MENU button changes to "CLOSE" and the hamburger icon to an X, but it requires a minor change in the JS. 

See DOAJ/doajPM#3003

### DIFF
--- a/cms/sass/layout/_secondary-nav.scss
+++ b/cms/sass/layout/_secondary-nav.scss
@@ -25,7 +25,7 @@
 
   .secondary-nav > .container {
     z-index: 3000;
-    position: fixed;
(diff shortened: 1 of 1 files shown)


sssoz:
a6fee0acc6694e0e188dbe3d500b9a86

amdomanska (approved):

PR #1956, opened 2021-07-28
Feature/bg harvester fixes

richard-jones:
This should fix the issues with the harvester background task in https://github.com/DOAJ/doajPM/issues/2153.  It:

* Changes how audit messages are tracked from the harvester code, so that each line of the audit is written as a separate record.  This prevents ES balking over very large fields
* Removes all email notifications
* Prevents multiple harvester jobs running at the same time
* Removes unused files, and strips back the settings
* Adds some unit tests covering the above

There are a few key things to note about this PR:

1. The setting `HARVESTER_API_KEYS` has gone, and is replaced with `HARVEST_ACCOUNTS`.  This will need to be updated in `production.cfg` on deploy
2. I have modified the BackgroundJobs ES mapping to not `.exact` analyse the audit messages.  Until we drop and re-create the index, this won't have any effect, as the mappings will not be pushed on an existing index.  Proposing to leave this in so that when we do the ES upgrade, the new index has the new, more efficient, mappings.


doajtest/unit/test_models.py
```diff
@@ -1281,6 +1279,14 @@ def test_26_background_job(self):
         bj.add_audit_message("message")
         assert len(bj.audit) == 2
 
+    def test_26a_background_job_active(self):
+        source = BackgroundFixtureFactory.example()
+        bj = models.BackgroundJob(**source)
+        bj.save(blocking=True)
+
+        assert models.BackgroundJob.has_active(source["action"])
```
amdomanska, line 1287: test message, please

amdomanska (changes requested):

Steven-Eardley (approved):
looks great

doajtest/unit/test_tasks_harvest.py
```diff
@@ -37,14 +36,10 @@ def setUp(self):
         self.old_harvester_api_keys = app.config.get('HARVESTER_API_KEYS')
         self.old_initial_harvest_date = app.config.get("INITIAL_HARVEST_DATE")
 
-
         app.config['HARVESTER_API_KEYS'] = {self.publisher.id: self.publisher.api_key}
```
Steven-Eardley, line 39: Config has changed

portality/tasks/harvester_helpers/workflow.py
```diff
@@ -124,6 +129,5 @@ def process_article(self, account_id, article):
         self._write_to_logger("Created article in DOAJ for Account:{x} with ID: {y}".format(x=account_id, y=id))
         return True
 
-
     def get_api_key(self, account_id):
```
Steven-Eardley, line 132: This is extraneous and the config is being changed

Steven-Eardley (changes requested):
found a couple of more things relating to the config change

PR #1960, opened 2021-08-06
Tests speed up

richard-jones:
Ready for merge to es upgrade branch


doajtest/helpers.py
```diff
@@ -3,43 +3,87 @@
...
 from datetime import datetime
 from glob import glob
 import os, csv, shutil
 from portality.lib import paths
+import functools
+import time
 
-prepare_for_test()
 
+def patch_config(inst, properties):
```
Steven-Eardley, line 16: this is nifty. However, if a test has specific overrides you'd have to make sure you don't assign to the same `originals` or those will be lost

doajtest/helpers.py
```diff
@@ -50,7 +94,7 @@ def _make_and_push_test_context(self, path="/", acc=None):
         ctx = self.app_test.test_request_context(path)
         ctx.push()
         if acc is not None:
-            acc.save(blocking=True)
+            # acc.save(blocking=True)
```
Steven-Eardley, line 97: can we get away with not saving our user? 

we have this in `core.py` which touches the index on login (`LoginManager.reload_user()`
```python
@login_manager.user_loader
def load_account_for_login_manager(userid):
    """
    ~~LoginManager:Feature->Account:Model~~
    :param userid:
    :return:
    """
    from portality import models
    out = models.Account.pull(userid)
    return out
    ```

doajtest/helpers.py
richard-jones: Yes, each test will have to store its own changes in its own member variable.  It'll break if you don't, I think, so should be easy to spot.

doajtest/helpers.py
richard-jones: Not sure yet, I just took it out for the moment so that there was no elasticsearch stuff in the main test superclass.  I think it'll have to go back in somehow, but I'm not sure how yet

Steven-Eardley (approved):

Steven-Eardley:
I'm going to merge this in, and we can extract @with_es down the line if it's superfluous in addition to the indexes on demand DAO wrapper

PR #1961, opened 2021-08-09
make sure the names of the groupfields are added to the fields list

amdomanska:
hotfix for: https://github.com/DOAJ/doajPM/issues/3020#issuecomment-894180459

### DIFF
--- a/portality/lib/formulaic.py
+++ b/portality/lib/formulaic.py
@@ -274,8 +274,7 @@ class FormulaicContext(object):
                 if field.group_subfields():
                     for sf in field.group_subfields():
                         fieldlist.append(sf)
-                else:
-                    fieldlist.append(field)
+                fieldlist.append(field)
         return fieldlist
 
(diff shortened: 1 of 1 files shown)


portality/lib/formulaic.py
```diff
@@ -274,8 +274,7 @@ def list_fields_in_order(self):
                 if field.group_subfields():
                     for sf in field.group_subfields():
                         fieldlist.append(sf)
-                else:
-                    fieldlist.append(field)
+                fieldlist.append(field)
```
Steven-Eardley, line 277: if this is meant to be ordered, should this be __before__ the subfield is added?

```python
for field in fs.fields():
    fieldlist.append(field)
    if field.group_subfields():
        for sf in field.group_subfields():
            fieldlist.append(sf)

```

amdomanska: Doesn't really matter in this case but I changed it to avoid any misinterpretations.

PR #1971, opened 2021-09-06
Feature/2677 review and map

richard-jones:
Another interim merge from the featuremap/cleanup branch


portality/lib/coerce.py
```diff
@@ -28,7 +32,7 @@ def to_isolang(output_format=None):
         * name
         * fr
     Can be a list in order of preference, too
-    fixme: we could make these pycountry's keys, removing the need for so many transformations and intermediate steps
+    ~~-> Languages:Data~~
```
Steven-Eardley, line 35: This might be better described as a library - backed by pycountry which uses the Debian maintained dataset for languages.

portality/forms/validate.py
```diff
@@ -110,56 +113,14 @@ def __make_optional(self, form, field):
         raise validators.StopValidation()
 
 
-class ExtraFieldRequiredIf(OptionalIf):
```
Steven-Eardley, line 113: I'm assuming this is unused

portality/lib/coerce.py
richard-jones: I'm thinking of it as using Language Data in the abstract here, and I've used it like this elsewhere so it ties it all up together.

portality/forms/validate.py
richard-jones: Yup, the only place it appears is here where it is declared

Steven-Eardley (approved):
I'm not sure about the consequences of removing config for `asynchronous_warnings` - was this an unimplemented feature?

Steven-Eardley (approved):

PR #1975, opened 2021-09-13
Feature/2447 email templates

richard-jones:
Updated all the email templates to be better jinja2 templates, and to include the date_applied

For https://github.com/DOAJ/doajPM/issues/2447

This is ready for merge


portality/notifications/application_emails.py
```diff
@@ -135,6 +138,7 @@ def send_publisher_application_editor_assigned_email(application):
     if owner is not None:
         send_list.append(
             {
+                "owner" : owner,
                 "name" : owner.name,
```
amdomanska, line 142: I cannot see the value "name" used anywhere further, is that needed?

portality/notifications/application_emails.py
```diff
@@ -243,19 +247,20 @@ def send_editor_completed_email(application):
...
 def send_publisher_update_request_inprogress_email(application):
     """Tell the publisher the UR is underway"""
-    journal_title = application.bibjson().title
+    # journal_title = application.bibjson().title
 
     owner = models.Account.pull(application.owner)
     send_list = [
         {
+            "owner" : owner,
             "name" : owner.name,
```
amdomanska, line 264: same as above

portality/notifications/application_emails.py
```diff
@@ -273,9 +278,11 @@ def send_publisher_update_request_inprogress_email(application):
             app_email.send_mail(to=to,
                                 fro=fro,
                                 subject=subject,
-                                template_name="email/publisher_update_request_inprogress.txt",
-                                publisher_name=instructions["name"],
-                                journal_title=journal_title)
+                                template_name="email/publisher_update_request_inprogress.jinja2",
+                                owner=instructions["owner"],
+                                application=application)
+                                # publisher_name=instructions["name"],
```
amdomanska, line 284: unused code - can be cleaned up

amdomanska (commented):
Just couple of cosmetic things.

portality/forms/application_processors.py
```diff
@@ -441,42 +441,41 @@ def finalise(self, account, save_target=True, email_alert=True):
...
 
-        to = [email]
+        to = [owner.email]
         fro = app.config.get('SYSTEM_EMAIL_FROM', 'feedback@doaj.org')
         if update_request:
             subject = app.config.get("SERVICE_NAME", "") + " - update request accepted"
         else:
             subject = app.config.get("SERVICE_NAME", "") + " - journal accepted"
-        publisher_name = publisher_name if publisher_name is not None else "Journal Owner"
+        # publisher_name = publisher_name if publisher_name is not None else "Journal Owner"
```
Steven-Eardley, line 456: yes, this is a better approach to this - having the default in the template now that it can include template logic.

Steven-Eardley (approved):
Looks like a good improvement. I know we had tests for emails, I haven't checked whether they'll test the template substitutions for us

portality/notifications/application_emails.py
richard-jones: It's not, I don't think, this code needs a more general tidy up. I decided to limit myself to just adding the thing I needed rather than getting dragged into even more upstream refactoring.  Propose to leave this as is, and we'll review it again when we move to the notifications system proper.

portality/notifications/application_emails.py
richard-jones: same response :)

richard-jones:
I've cleaned up the dead, commented out code, this is now ready to go live as per https://github.com/DOAJ/doajPM/issues/2447#issuecomment-933187004

PR #1986, opened 2021-10-02
Hide apc charges from articles results

amdomanska:
Check whether the results is journal, if yes then proceed as before, if no return False and hide the field.
https://github.com/DOAJ/doajPM/issues/3078


Steven-Eardley (approved):
Yep, looks good.

richard-jones (approved):

PR #1992, opened 2021-10-05
Fix the application scrolling

amdomanska:
Application scrolls to the first invalid field only if validation is required and application has not been saved.


richard-jones (commented):
I ran this locally, and it *didn't* scroll me when I switched the application to a validatable state and then hit Save.  What behaviour with scrolling are we aiming for here?

amdomanska:
It should scroll whenever the application should be validated but is not valid - and hence is not saved. Could you please give me more details on when you see the incorrect behaviour? I've checked that locally and seems to work fine.

richard-jones (approved):

PR #1996, opened 2021-10-12
make sure the form is validated before enabling submit button

amdomanska:
for https://github.com/DOAJ/doajPM/issues/3063


amdomanska:
Merged into https://github.com/DOAJ/doaj/pull/1976

PR #2001, opened 2021-10-18
fixes a display bug with the readonly application form.  This changes…

richard-jones:
… the tab display feature to only show the single latest selected tab, to get rid of some problems arising from asynchronous calls to change tab display via jquery animate.  Also generally tidies up the 'summary table only' view

HOTFIX for https://github.com/DOAJ/doajPM/issues/3097


Steven-Eardley (approved):
works great

PR #2002, opened 2021-10-18
Display Seal logo

sssoz:
Display SVG logo for journals that have been awarded the DOAJ Seal.

Can be seen on:
- Results page: http://localhost:5004/search/journals?source={"query":{"filtered":{"filter":{"bool":{"must":[{"term":{"index.has_seal.exact":"Yes"}}]}},"query":{"match_all":{}}}},"size":50,"sort":[{"created_date":{"order":"desc"}}]} 
- Journal details page: http://localhost:5004/toc/2624-5647


richard-jones (changes requested):
There's a lot of duplication of the seal logo in this code:

* 2 copies of the actual file in cms/assets/img and templates/includes
* 3 copies in doaj.fieldrender.edges.js

We should DRY this up as much as possible.

1. The copy in cms/assets/img/logotype doesn't appear to be referenced anywhere, can we just remove it?
2. In doaj.fieldrender.edges.js, we should either load the svg as a regular image file like you would do with any binary format, OR we should add the body of the svg to a module variable and re-use it, rather than replicating it in full each time it is needed.

sssoz:
Happy to do a cleanup of the static assets, however it wasn’t part of this issue (https://github.com/DOAJ/doajPM/issues/3016) so I considered it a separate task. 

@richard-jones I would rather we stick to the SVG so that it is scalable and resizable without damaging the quality -- and I can eventually add interactivity to it.

sssoz:
@richard-jones 

> The copy in cms/assets/img/logotype doesn't appear to be referenced anywhere, can we just remove it?

That logotype has nothing to do with the Seal logo, which is what this PR is about. It’s the different variations of the DOAJ logo which we might link to from other places. I would rather we keep it since the logo should exist in both black & white, and transparent backgrounds.

richard-jones:
> @richard-jones I would rather we stick to the SVG so that it is scalable and resizable without damaging the quality -- and I can eventually add interactivity to it.

Yes, I agree, what I mean is that we could serve the svg like a regular file, rather than inlining it in the code.

richard-jones:
> The copy in cms/assets/img/logotype doesn't appear to be referenced anywhere, can we just remove it?

Ah, sorry, I didn't realise that it was a different file - that's fine!

Steven-Eardley (approved):

PR #2003, opened 2021-10-18
Make review table header’s base colour white

ssssoz:
The `.button` component uses the CSS value `currentColor` which was set to a dark grey by default for table headers. Simply changed the base colour for the `review-table` component to be white. 

See DOAJ/doajPM#3047


richard-jones (approved):
Just to confirm that I changed the merge target from develop to master for hotfix

Steven-Eardley (approved):

PR #2011, opened 2021-10-25
CSV with user accounts and OA Start date

amdomanska:
for https://github.com/DOAJ/doajPM/issues/3092
`admin_journalcsv.py` should be run with `-a' flag to NOT obcure the accounts. Without the flag the script works as it used to - it DOES obscure the accounts.

### DIFF
--- a/portality/bll/services/journal.py
+++ b/portality/bll/services/journal.py
@@ -185,8 +185,12 @@ class JournalService(object):
             else:
                 return [("Owner", o)]
 
+        def oa_start(j):
+            oa = j.bibjson().oa_start
+            return [("OA Start", oa)]
+
         with open(file_path, "w", encoding="utf-8") as f:
-            self._make_journals_csv(f, usernames)
+            self._make_journals_csv(f, [usernames, oa_start])
 
     def _make_journals_csv(self, file_object, additional_columns=None):
         """
@@ -206,8 +210,7 @@ class JournalService(object):
                 ("DOAJ Seal", YES_NO.get(journal.has_seal(), "")),
                 # ("Tick: Accepted after March 2014", YES_NO.get(journal.is_ticked(), "")),
                 ("Added on Date", journal.created_date),
-                ("Last updated Date", journal.last_manual_update),
-                ("OA Start Date", journal.bibjson().oa_start)
+                ("Last updated Date", journal.last_manual_update)
             ]
             return kvs
 
@@ -238,7 +241,8 @@ class JournalService(object):
             article_kvs = _get_article_kvs(j)
             additionals = []
             if additional_columns is not None:
-                additionals = additional_columns(j)
+                for col in additional_columns:
+                    additionals += col(j)
             cols[issn] = kvs + meta_kvs + article_kvs + additionals
 
             # Get the toc URL separately from the meta kvs because it needs to be inserted earlier in the CSV
@@ -254,6 +258,7 @@ class JournalService(object):
             if qs is None:
                 qs = [q for q, _ in cols[i]]
                 csvwriter.writerow(qs)
+                print(qs)
             vs = [v for _, v in cols[i]]
             csvwriter.writerow(vs)
 
--- a/portality/bll/services/journal.py
+++ b/portality/bll/services/journal.py
@@ -160,12 +160,13 @@ class JournalService(object):
         models.Cache.cache_csv(url)
         return url, action_register
 
-    def admin_csv(self, file_path, account_sub_length=8, obscure_accounts="true"):
+    def admin_csv(self, file_path, account_sub_length=8, obscure_accounts=True):
         """
         ~~AdminJournalCSV:Feature->JournalCSV:Feature~~
         
         :param file_path:
         :param account_sub_length:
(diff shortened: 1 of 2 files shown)


portality/bll/services/journal.py
```diff
@@ -160,7 +160,7 @@ def filter(filename):
         models.Cache.cache_csv(url)
         return url, action_register
 
-    def admin_csv(self, file_path, account_sub_length=8):
+    def admin_csv(self, file_path, account_sub_length=8, obscure_accounts="true"):
```
richard-jones, line 163: Let's not pass `"true"` as a string here, let's make it a bool

portality/bll/services/journal.py
```diff
@@ -171,18 +171,21 @@ def admin_csv(self, file_path, account_sub_length=8):
         # create a closure for substituting owners for consistently used random strings
         unmap = {}
 
-        def obscured_usernames(j):
+        def usernames(j):
             o = j.owner
-            sub = None
-            if o in unmap:
-                sub = unmap[o]
+            if (obscure_accounts == "true"):
```
richard-jones, line 176: Then this just becomes a bool check

portality/scripts/admin_journalcsv.py
```diff
@@ -12,11 +12,12 @@
 
     parser = argparse.ArgumentParser()
     parser.add_argument("-o", "--out", help="output file for admin csv")
+    parser.add_argument("-a", "--obscure_accounts", help="true if accounts should be obscured", default="true")
```
richard-jones, line 15: You can make this a flag with no argument to consume, and set `store_true` as the action.  That means you just need to set `-a` on its own, and then `args.obscure_accounts` will be boolean: https://docs.python.org/3/library/argparse.html

richard-jones (changes requested):

portality/scripts/admin_journalcsv.py
amdomanska: I set the action to `store_false` to keep the default behaviour (DO obscure the accounts when the flag is NOT set)

portality/scripts/admin_journalcsv.py
```diff
@@ -12,12 +12,12 @@
...
     parser = argparse.ArgumentParser()
     parser.add_argument("-o", "--out", help="output file for admin csv")
-    parser.add_argument("-a", "--obscure_accounts", help="true if accounts should be obscured", default="true")
+    parser.add_argument("-a", "--do_not_obscure_accs", help="set this flag if the accounts should not be obscured", action='store_false')
     args = parser.parse_args()
 
     # ~~-> Journal:Service~~
     journalService = DOAJ.journalService()
-    journalService.admin_csv(file_path=args.out, obscure_accounts=args.obscure_accounts)
+    journalService.admin_csv(file_path=args.out, obscure_accounts=args.do_not_obscure_accs)
```
richard-jones, line 20: I like the default to obscured approach, very nice.

portality/bll/services/journal.py
```diff
@@ -202,7 +206,8 @@ def _get_doaj_meta_kvs(journal):
                 ("DOAJ Seal", YES_NO.get(journal.has_seal(), "")),
                 # ("Tick: Accepted after March 2014", YES_NO.get(journal.is_ticked(), "")),
                 ("Added on Date", journal.created_date),
-                ("Last updated Date", journal.last_manual_update)
+                ("Last updated Date", journal.last_manual_update),
+                ("OA Start Date", journal.bibjson().oa_start)
```
richard-jones, line 210: This is going to make the public journal CSV expose the OA Start Date, which we don't want to do yet.  We only want that available in the admin csv for the moment, so we need to move this to the other method.

richard-jones (changes requested):

portality/bll/services/journal.py
amdomanska: I think I figured that out now!

portality/bll/services/journal.py
```diff
@@ -160,29 +160,37 @@ def filter(filename):
...
         unmap = {}
 
-        def obscured_usernames(j):
+        def usernames(j):
             o = j.owner
-            sub = None
-            if o in unmap:
-                sub = unmap[o]
+            if obscure_accounts:
+                sub = None
```
Steven-Eardley, line 178: `sub` here is never used (can't be `None` due to if /else)

portality/scripts/admin_journalcsv.py
```diff
@@ -12,11 +12,13 @@
 
     parser = argparse.ArgumentParser()
     parser.add_argument("-o", "--out", help="output file for admin csv")
+    parser.add_argument("-a", "--do_not_obscure_accs", help="set this flag if the accounts should not be obscured", action='store_false')
     args = parser.parse_args()
 
     # ~~-> Journal:Service~~
     journalService = DOAJ.journalService()
-    journalService.admin_csv(args.out)
+    cols = ["oa_start"]
```
Steven-Eardley, line 20: unused code again, the admin csv function adds oa_start itself

Steven-Eardley (approved):

PR #2015, opened 2021-11-01
Fix sticky header for safari

sssoz:
See DOAJ/doajPM#2987 

This needs to be tested, if possible, with Dom, Judith, Cenyu (who had found the issue originally).


portality/static/js/doaj.js
```diff
@@ -12,38 +12,25 @@ var doaj = {
...
-        var topBtn = document.getElementById("top");
+        // On scroll, display back-to-top button & add class to header primary menu
+        var topBtn = document.getElementById("top"),
+            topNav = document.querySelector(".primary-nav");
 
-        function displayTopBtn() {
-            if (topBtn) {
+        function displayOnScroll() {
+            if (topBtn && topNav) {
                 if (document.body.scrollTop > 100 || document.documentElement.scrollTop > 100) {
```
richard-jones, line 21: While we're changing this bit of code, it would be good to pull the magic numbers out of here into a configuration-like property (or even calculate them dynamically if that's possible?).

At the least, we should probably set at the top of this file

```
var doaj = {
   navHeight: 100,
   init: function() { .... }
}
```

Then here you can do 

```
if (document.body.scrollTop > doaj.navHeight || document.documentElement.scrollTop > doaj.navHeight) {
```

richard-jones: `navHeight` might not be the right property name, but insert whatever name makes most sense.

If it's possible to calculate that value, that would be even better, then any changes to the navigation would automatically adjust the scroll behaviour.  I don't know if that's possible - you might be able to inspect the DOM and determine the size of the nav element, but I'd have to experiment with it a bit to make sure we could do it reliably.

richard-jones (changes requested):
I haven't tested this in the browser, but code review is mostly fine, just one suggestion to improve the maintainability

portality/static/js/doaj.js
sssoz: So I did choose 100px the point at which these elements should appear / change on the page a bit randomly, but I couldn’t really think of a more dynamic way to display it. Just to be clear, this function allows the "scroll to top" button at the bottom right to appear as well as the navbar to be minimised when the user scrolls past 100px. 

We don’t want the "scroll to top" button to always be displayed too early (i.e. when the user is already at the top or close to the top); I made the navbar change at the same time just because I find it looks nicer and less distracting to have both animations at the same time. 100px seemed like a good point to start simply. Do you have an idea of a smarter / less _magical_ way to define this?

sssoz: I’ve implemented your simpler suggestion here though: 360d48a91e8c5880fb9ff558fcb3e5b5924cc2e0

richard-jones (approved):

portality/static/js/doaj.js
richard-jones: I'd have to play around a bit to see if there was a dynamic way of assigning this number.  If it's based on some physical aspect of the page, like the height of the nav or the size of the "return to the top" button we could inspect those elements on page load for size in the user's environment and use that number.  

I think given the effort/benefit, it's probably fine just to set it here in one place, then at least it's quick to update if we ever need to :)

Steven-Eardley (approved):

PR #2021, opened 2021-11-01
Add WeChat link

sssoz:
See DOAJ/doajPM#2989

Not sure about a better way to do this. WeChat’s link is a QR code that seems to be usually shown in a modal, as opposed to an external link, as with the other menu items. The WeChat modal is hardcoded by the footer now, as it can’t be conditionally displayed in the `<li>` alongside its WeChat menu item. 

@richard-jones Do you have an idea?


richard-jones:
I think the overall approach is fine, it just needs a bit more generality.  I've just pushed a commit which does that, take a look and see if that works for you.

sssoz:
@richard-jones look good and makes a lot more sense! Who knows, we might have other modals linked from the footer eventually.

richard-jones (approved):

Steven-Eardley (approved):

PR #2025, opened 2021-11-19
except some buttons from the application form from triggering the 'le…

richard-jones:
…ave' warning

For https://github.com/DOAJ/doajPM/issues/3159


Steven-Eardley (approved):
I agree with you that hardocding out specific instances to exclude is probably an unreliable approach, but i'm happy to release this.

Can we opt-in via a class rather than opt-out?

PR #1872, opened 2021-03-22
Improve & clarify disabled fields in update requests

sssoz:
See DOAJ/doajPM#2679

PR #2029, opened 2021-12-15
Added samesite security for cookies

RK206:
Added samesite configuration for cookies to be secure  #1872


richard-jones (approved):

RK206:
Updated app.py as the user cannot  login after these changes in http mode. Added ssl_context so that app will run in https mode in development

richard-jones:
Does that mean we'll need to set up our local instances to run under https now?

RK206:
Yes, app.py has been updated to run the app in  https mode. So the  app will run in  https mode by default in development.

richard-jones:
ok, before we merge this, can we go through the steps that we need to do locally to make sure we can continue to develop (I'm a bit vague on using https locally)

RK206:
Reverted the changes as per the discussion in technical meeting.
Developers should add the following configuration in dev.cfg to skip cookie security.
SESSION_COOKIE_SAMESITE=None
SESSION_COOKIE_SECURE=False
REMEMBER_COOKIE_SECURE=False

PR #2036, opened 2021-12-29
Updated script to remove licenses from article

RK206:
I have tested and the licenses are getting removed from journal.
Also the license under article is getting removed. If it is expected behavior than it should be fine. Otherwise it can be fixed.
This is happening when the article object is created


Steven-Eardley (approved):

Steven-Eardley (approved):

Steven-Eardley:
So good I approved it twice. Merging.

PR #2037, opened 2021-12-29
Separate Applications and Update Requests in Admin Interface

amdomanska:
https://github.com/DOAJ/doajPM/issues/3160


portality/settings.py
```diff
@@ -683,8 +683,16 @@
         "suggestion" : {
             "auth" : True,
             "role" : "admin",
+            "query_filters" : ["not_update_request"],
```
richard-jones, line 686: I think this is fine for the route, but the actual method behind `not_update_request` is still differentiating between update requests and new applications in the old way, it needs to look at `admin.application_type` for either new applications or update requests (see `constants.APPLICATION_TYPE_NEW_APPLICATION` and the related UR one).

portality/static/js/edges/admin.applications_and_update_requests.edge.js
```diff
@@ -59,7 +59,7 @@ $.extend(true, doaj, {
                     id: "application_status",
                     category: "facet",
                     field: "admin.application_status.exact",
-                    display: "Application Status",
+                    display: "Status",
```
richard-jones, line 62: I'm attaching this comment here, but it's more broadly to do with this new file.  What I meant was for us to produce 2 new edges files, something like `admin.new_applications.edge.js` and `admin.update_requests.edge.js` and for the common elements of them to be factored into `doaj.fieldrender.edges.js` (like, facet definitions, result display definitions, etc).  Your approach here works fine until we want to vary the behaviours of the pages independently of each other, at which point we have to do that refactoring anyway.  So, I could be convinced to leave it as it is for the moment, but maybe we should check that the team don't want to change the list of facets for each page at this point, and if they do then we should go the extra distance to refactor.

portality/templates/admin/nav.html
```diff
@@ -1,13 +1,15 @@
...
 {% set users = url_for("account.index") %}
 {% set site_search = url_for("admin.admin_site_search") %}
 {% set editor_groups = url_for("admin.editor_group_search") %}
 {% set bg = url_for("admin.background_jobs_search") %}
 
 {% set tabs = [
     (journals, "Journals"),
-    (applications, "Applications/update requests"),
+    (applications, "Applications"),
+    (update_requests, "Update Requests"),
```
richard-jones, line 12: We should check if there are any downstream consequences to this.  Like, are there any pages that link to the applications search that will be tripped up by update requests now appearing under a new link?  The featuremap doesn't currently reach to all these corners of the system, so isn't itself likely to yield that answer, probably should just do some creative searching over the code to see.

richard-jones (changes requested):
This looks pretty good, there are two major feedback bits:

1. The mechanism for selecting between update requests and new applications needs to be updated to use the new data model element that does that.  I didn't see the migration from the related branch merged in here, I guess that needs to be merged into this too to give us the complete package: https://github.com/DOAJ/doajPM/issues/1779

2. The code refactor isn't quite as I'd imagined it, and the approach taken here doesn't have any flexibility to allow us to vary the pages independently if necessary.  We should find out if we need to vary them independently and then decide whether we need to go further with the refactoring.

richard-jones:
Oh, my mistake on the migration, that was already merged, it is in `20210906_1779_urlabelling`, so we just need to update the code to use the new data model, and make sure that migration is run during deployment (and, indeed, on test during testing)

portality/static/js/edges/admin.applications_and_update_requests.edge.js
richard-jones: My intention was that the repeated bits of code would go into `doaj.fieldrender.edges.js` and be re-used by each of these components.  You can look at how the `inDOAJ` facet is handled in journals, for example

portality/lib/query_filters.py
```diff
@@ -50,13 +50,13 @@ def owner(q):
 def update_request(q):
     q.clear_match_all()
     q.add_must_filter({"range" : {"created_date" : {"gte" : app.config.get("UPDATE_REQUEST_SHOW_OLDEST")}}})
-    q.add_must_filter({"exists" : {"field" : "admin.current_journal"}})
+    q.add_must_filter({"term" : {"admin.application_type" : constants.APPLICATION_TYPE_UPDATE_REQUEST}})
```
richard-jones, line 53: this still needs `admin.application_type.exact`

portality/lib/query_filters.py
```diff
@@ -50,13 +50,13 @@ def owner(q):
...
     return q
 
 
 def not_update_request(q):
     q.clear_match_all()
-    q.add_must_not({"exists" : {"field" : "admin.current_journal"}})
+    q.add_must_not({"term" : {"admin.application_type" : constants.APPLICATION_TYPE_UPDATE_REQUEST}})
```
richard-jones, line 59: `admin.application_type.exact`

portality/settings.py
```diff
@@ -683,8 +683,16 @@
...
             "dao" : "portality.models.Suggestion"    # ~~->Application:Model~~
         },
+        # ~~->AdminUpdateRequestQuery:Endpoint~~
+        "update_requests": {
+            "auth": True,
+            "role": "admin",
+            "query_filters" : ["update_request"],
+            "dao": "portality.models.Suggestion"  # ~~->Application:Model~~
```
richard-jones, line 694: Since this is new code, can we use the new model, rather than the back-compat model.  `portality.models.Application`

portality/static/js/edges/admin.applications_and_update_requests.edge.js
richard-jones: See: https://github.com/DOAJ/doaj/blob/develop/portality/static/js/doaj.fieldrender.edges.js#L3 and https://github.com/DOAJ/doaj/blob/develop/portality/static/js/edges/admin.journals.edge.js#L78

So each interface is composed from a variety of standard components (ideally)

richard-jones (changes requested):

portality/templates/admin/nav.html
richard-jones: application_emails.py  def send_admin_ready_email(application, editor_id):

richard-jones (approved):
@amdomanska and I have worked together on this PR, and the latest commit covers the only outstanding thing that we identified, so I'm happy with this

PR #2038, opened 2021-12-29
Prevent leave warning when withdrawing the journal...

amdomanska:
.. by adding `this.submitting = true` flag for withdraw button.

https://github.com/DOAJ/doajPM/issues/3159


portality/static/js/application_form.js
```diff
@@ -691,6 +696,7 @@ doaj.af.ManEdApplicationForm = class extends doaj.af.EditorialApplicationForm {
         super(params);
 
         $("#open_quick_reject").on("click", (e) => {
+            console.log("click open");
```
richard-jones, line 699: console log for cleanup

portality/static/js/application_form.js
```diff
@@ -486,6 +489,8 @@ doaj.af.EditorialApplicationForm = class extends doaj.af.BaseApplicationForm {
     }
 
     beforeUnload(event) {
+        // This solution will withdraw the journal without the warning even if there are some changes that are not saved.
+        // should that if condition be changed to AND: if (!this.changed && this.submitting)
```
richard-jones: Changing it the way I have recommended fixes this

richard-jones (commented):
I have taken the liberty of rewriting this one, unfortunately the original implementation confused the two possible approaches, and as a result introduced a change that prevented the "don't leave" warning from appearing at all after a particular button was pressed.

Steven-Eardley (approved):

PR #2039, opened 2021-12-29
Do not allow to accept UR for deleted or withdrawn journals

amdomanska:
https://github.com/DOAJ/doajPM/issues/2764


portality/templates/application_form/editorial_form_body.html
```diff
@@ -1,3 +1,10 @@
+{% if current_journal and not current_journal.is_in_doaj() %}
```
amdomanska, line 1: This checks whether the current journal is not withdrawn. It is not defined if we open the journal form. How to check whether journal has been deleted? I guess current_journal will exist but will it be just empty string?

portality/templates/application_form/editorial_form_body.html
```diff
@@ -1,3 +1,10 @@
+{% if obj.application_status == "update_request" and (not current_journal or not current_journal.is_in_doaj()) %}
```
Steven-Eardley, line 1: I'm not sure whether `current_journal` is passed into the template, should it be `obj.current_journal`?

portality/forms/application_processors.py
```diff
@@ -303,6 +303,10 @@ def finalise(self, account, save_target=True, email_alert=True):
             raise Exception("You cannot edit a not-existent application")
         if self.source.application_status == constants.APPLICATION_STATUS_ACCEPTED:
             raise Exception("You cannot edit applications which have been accepted into DOAJ.")
+        if self.form.application_status.data == "accepted" and (
```
Steven-Eardley, line 306: Please add test case for this

Steven-Eardley (changes requested):

portality/forms/application_processors.py
richard-jones: Also, should use the value from `constants`, not the magic string

portality/forms/application_processors.py
```diff
@@ -303,6 +303,10 @@ def finalise(self, account, save_target=True, email_alert=True):
...
+                self.source.current_journal is None or not models.Journal.pull(
+            self.source.current_journal).is_in_doaj()):
```
richard-jones, line 308: This line could fail if `Journal.pull` returns None, I would be inclined to break this out into a block rather than inlining it into the if statement

```python
if self.form.application_status.data == "accepted" and self.source.current_journal is not None:
    j = models.journal.pull(self.source.current_journal)
    if j is None or not j.is_in_doaj():
        raise Exception()
```

richard-jones: Generally speaking I recommend simple if statements, and to prep all the things you need beforehand.  Inlining a load from database is a good warning sign that the if statement is too complex

portality/forms/application_processors.py
```diff
@@ -303,6 +303,10 @@ def finalise(self, account, save_target=True, email_alert=True):
...
+            raise Exception("This journal has been withdrawn or deleted, update request cannot be accepted.")
```
richard-jones, line 309: Can we add this text into the messages.py file

portality/forms/application_processors.py
```diff
@@ -511,6 +515,7 @@ def validate(self):
         valid = super(AdminApplication, self).validate()
 
         if self.form is not None:
+
```
richard-jones, line 518: extra whitespace for cleanup

portality/static/js/application_form.js
```diff
@@ -703,6 +707,9 @@ doaj.af.ManEdApplicationForm = class extends doaj.af.EditorialApplicationForm {
             }
             that.submitting = true;
         });
+
+
+
```
richard-jones, line 712: extra whitespace for cleanup

portality/templates/application_form/editorial_form_body.html
richard-jones: Also, I believe we have access to `constants` here so we should use that instead of `"update_request"`

richard-jones: We **do** have access to `current_journal` in this context

richard-jones: `obj.application_status` isn't going to be reliable here, as the status of an item changes after it goes through triage.  There are 2 ways to deal with this:

1. Ask the object if it has the **property** `current_journal`
2. Merge this code with the other PR for distinguishing URs from new applications, and then use `obj.admin.application_type=="update_reqeust"` (but using `constants`)

The first object is probably sufficient, because `current_journal` only gets removed when the item is accepted or rejected, though (2) would be a better overall longer term solution

richard-jones: So, to be clear on (1) we would do

```
{% if obj.admin.current_journal is not None and (not current_journal or not current_journal.is_in_doaj()) %}
```

portality/templates/application_form/js/_form_js.html
```diff
@@ -29,6 +29,9 @@
             {% if form_diff %}
                 formDiff: {{ form_diff|tojson }},
             {% endif %}
+            {% if obj.application_status == "update_request" and (not current_journal or not current_journal.is_in_doaj()) %}
+                withdrawn: "true"
+            {% endif %}
```
richard-jones, line 34: Same comments as above about determining if `withdrawn` is true.  Also, why is `"true"` a string here?  Should be a boolean

richard-jones (changes requested):

portality/templates/application_form/editorial_form_body.html
amdomanska: I'm going to implement the first solution now to not mix up PRs before they are ready. I guess we can change it once everything is merged and working.

richard-jones (approved):

portality/templates/application_form/js/_form_js.html
```diff
@@ -29,6 +29,9 @@
             {% if form_diff %}
                 formDiff: {{ form_diff|tojson }},
             {% endif %}
+            {% if obj.admin.current_journal is not none and (not current_journal or not current_journal.is_in_doaj()) %}
+                withdrawn: True
```
richard-jones, line 33: `True` is the python true, for javascript it needs to be `true`

richard-jones (changes requested):

portality/templates/application_form/js/_form_js.html
amdomanska: Fixed. @Steven-Eardley I think we may update that on the test server now.

amdomanska:
@Steven-Eardley @richard-jones Could I please ask you to review this?

portality/models/v2/journal.py
```diff
@@ -817,6 +819,20 @@ def propagate_in_doaj_status_to_articles(self):
...
+            from portality.models.v2.application import Application
+            ca = Application.pull(self.current_application)
+            if ca is None:
+                raise exceptions.IllegalStatusException(message="Data mismatch: current application of this journal does not exist")
+            ca.add_note("Update request was automatically rejected because the associated journal was withdrawn or "
+                        "deleted.")
+            ca.save()
+            from portality.bll.doaj import DOAJ
+            svc = DOAJ.applicationService()
+            svc.reject_application(ca, current_user._get_current_object())
```
richard-jones, line 833: I don't love the positioning of this, there's a lot of application level mixing going on.  The models should not be calling services, I don't think.

Instead there should be a service who is responsible for rejecting update requests

portality/tasks/journal_in_out_doaj.py
```diff
@@ -61,6 +61,9 @@ def run(self):
             j.set_in_doaj(in_doaj)
             j.save()
             j.propagate_in_doaj_status_to_articles()  # will save each article, could take a while
+            if not in_doaj:
+                job.add_audit_message("Rejecting all associated update requests")
+                j.reject_update_requests()
```
richard-jones, line 66: I think this needs to change.  We want this to be in the application service, where related methods are, like `reject_journal`.

So, here we would have a call to 

```
svc = DOAJ.applicationService()
svc.reject_update_requests()
```

and in `portality.bll.service.application` then we want the contents of `Journal.reject_update_requests` to be reflected as a BLL method rather than a model method.

richard-jones: Also, it would be nice if this could report on the update requests rejected, so have that `reject_update_requests` return some status saying whether it has rejected them or not.  This could be a true or false, or a count (there should be only one) or an object with some status informaiton.  Not sure which is best, probably the count.  Then this could be logged in the audit message

portality/models/v2/journal.py
richard-jones: See my comment elsewhere about where to put this functionality.

richard-jones (changes requested):

richard-jones:
In addition to the requested changes, can we also get the featuremap annotations in before this is merged

amdomanska:
PR comments addressed, feature map annotations added. @richard-jones Could you please re-review?

portality/bll/services/application.py
```diff
@@ -251,6 +252,17 @@ def accept_application(self, application, account, manual_update=True, provenanc
...
         return j
 
+    def reject_update_request_of_journal(self, journal_id):
+        ur, jlock, alock = self.update_request_for_journal(journal_id)
+        if ur:
+            ur.add_note("Update request was automatically rejected because the associated journal was withdrawn or "
+                            "deleted.")
+            ur.save()
+            self.reject_application(ur, current_user._get_current_object())
+            return ur.id
```
richard-jones, line 262: Since you have the update reqeust object, why not just return the whole thing?  This feels like the one use case you had for it required the id, so you just returned that, but the more general approach would be to return the object, then any caller can do what they want with it

richard-jones (changes requested):
I'm not sure what happened here, all it really needed was that function moved and tidied up a bit, but instead it's gone all a bit wrong!

portality/bll/services/application.py
```diff
@@ -251,6 +252,17 @@ def accept_application(self, application, account, manual_update=True, provenanc
 
         return j
 
+    def reject_update_request_of_journal(self, journal_id):
+        ur, jlock, alock = self.update_request_for_journal(journal_id)
```
richard-jones, line 256: THis function creates the update request if it does not exist, so you can't use this method here.  You just need to get the UR directly if it exists

portality/bll/services/application.py
```diff
@@ -251,6 +251,22 @@ def accept_application(self, application, account, manual_update=True, provenanc
...
 
+    def reject_update_request_of_journal(self, journal_id, account):
+        """
+            Rejects update request associated with journal
+
+            :param journal_id:
+            :param account:
+            :return: Journal object
+        """
+        ur = models.Suggestion.find_latest_by_current_journal(journal_id)  # ~~->Application:Model~~
```
richard-jones, line 262: `Suggestion` is the old model name, kept around for back-compat of old code, we use `models.Application` in new code

portality/bll/services/application.py
```diff
@@ -251,6 +251,22 @@ def accept_application(self, application, account, manual_update=True, provenanc
...
+        if ur:
+            self.reject_application(ur, account, note="Update request was automatically rejected because the associated journal was withdrawn or "
```
richard-jones, line 264: Can we put the note in `portality.ui.messages`

portality/models/v2/journal.py
```diff
@@ -1,3 +1,5 @@
+from flask_login import current_user
```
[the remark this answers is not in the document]
richard-jones: This still needs to be removed

portality/models/v2/journal.py
```diff
@@ -1,3 +1,5 @@
...
+from portality.bll import exceptions
```
richard-jones, line 2: This needs to be removed

portality/tasks/journal_in_out_doaj.py
```diff
@@ -53,10 +54,18 @@ def run(self):
...
+            # ~~->Account:Model~~
+            account = models.Account.pull(job.user)
             if j is None:
                 raise RuntimeError("Journal with id {} does not exist".format(journal_id))
+            if not in_doaj:
+                job.add_audit_message("Rejecting all associated update requests")
+                # ~~->Application:Service~~
+                svc = DOAJ.applicationService()
+                ur = svc.reject_update_request_of_journal(j.id, account)
+                job.add_audit_message("Update request {x} automatically rejected".format(x=ur))
```
richard-jones, line 68: `ur` can be none, that needs to be caught, and the audit messages need to make sense.  Either don't add the audit messages at all if there's no UR, or make the second message say that (I'd lean towards the former)

portality/bll/services/application.py
```diff
@@ -251,6 +251,22 @@ def accept_application(self, application, account, manual_update=True, provenanc
 
         return j
 
+    def reject_update_request_of_journal(self, journal_id, account):
```
richard-jones, line 254: Can we have a unit test or two for this function please

richard-jones (changes requested):

portality/bll/services/application.py
amdomanska: Already implemented.

portality/bll/services/application.py
```diff
@@ -80,7 +80,7 @@ def reject_application(self, application, account, provenance=True, note=None, m
         if manual_update:
             application.set_last_manual_update()
 
-        saved = application.save()
+        saved = application.save(blocking=True)
```
richard-jones, line 83: We try to avoid using `blocking` saves in the live application, as it can result in long hangs in the response times.  I suspect you've put this in here to aid a test, in which case you need to put the block or the `time.sleep` in the test instead.

portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -3127,6 +3127,9 @@ $.extend(true, doaj, {
                 }
 
                 var status = "";
+                if (!resultobj.admin.current_journal || !resultobj.admin.current_journal.in_doaj){
```
richard-jones, line 3130: `resultobj.admin.current_journal.in_doaj` isn't a field that exists

richard-jones: As this is the search interface all you have is `resultobj.admin.current_journal` on an application, which gives the ID of the current journal.  It won't have a property `in_doaj` that you can query here.

richard-jones (changes requested):
A couple of minor things, one to do with testing I think

portality/bll/services/application.py
amdomanska: It is not for the test. It's the race caused somehow by the background job. I discussed that with @Steven-Eardley, still better than putting `time.sleep` on the background job.

portality/static/js/doaj.fieldrender.edges.js
amdomanska: I think this is not needed at all, it is the leftover from the previous version.

PR #2045, opened 2022-01-10
Adjust apc display

sssoz:
See DOAJ/doajPM#3077


richard-jones (changes requested):
This looks fine to me, in all functional ways I approve the PR.

The only thing is that as a hotfix it appears to have acquired some changes to `setup.py`.  These are changes that we have at least pushed to `develop` and I'm not sure whether they are in `master` or not, but in any case this PR shouldn't have them as they are unrelated to the change.  @Steven-Eardley maybe you can tell what's going on, as you merged my PR with these changes to `setup.py` before?

So, before we merge this we just need to sort out what's going on with that change being included.

sssoz:
@richard-jones I didn’t understand either – this branch was forked from `master` and I just re-based it off `master` but your commit b7a30f26fade536dceca6aa26425e4995207a081 still appears

Steven-Eardley:
Cherrypicked and merged.

PR #2049, opened 2022-01-24
New dashboard layout 

sssoz:
Integrating dashboard UI @richard-jones’s work on the Editorial TODO list. 

See DOAJ/doajPM#3175 for new dashboard layout


portality/templates/layouts/base_dashboard.html
```diff
@@ -0,0 +1,176 @@
+<!DOCTYPE html>
```
richard-jones, line 1: I'm not so keen to have a second full base template, as it is going to create some duplication between the other base template.  Can we just adjust the balance of what is in the one main base template, and what the new admin area template looks like.

Sorry, I realise I misunderstood when we had our meeting, I thought the base_dashboard would extend the base template, and be for use by the dashboard page.

sssoz: I thought so :) I will reorganise it to have sub-base templates for the public site vs the dashboard. Didn’t make much sense to repeat standard <head>  content

sssoz: Done in https://github.com/DOAJ/doaj/pull/2049/commits/f4a49f20659885a17b145376454d8224b8e154ae

tHANK you!

portality/constants.py
```diff
@@ -38,8 +38,8 @@
...
-TODO_MANED_STALLED = "todo_maned_stalled"
-TODO_MANED_FOLLOW_UP_OLD = "todo_maned_follow_up_old"
-TODO_MANED_READY = "todo_maned_ready"
-TODO_MANED_COMPLETED = "todo_maned_completed"
-TODO_MANED_ASSIGN_PENDING = "todo_maned_assign_pending"
+TODO_MANED_STALLED = "Chase <strong>stalled</strong> application"
+TODO_MANED_FOLLOW_UP_OLD = "Follow up with <strong>old<strong> application"
+TODO_MANED_READY = "Make a decision on <strong>ready<strong> application"
+TODO_MANED_COMPLETED = "Review <strong>completed<strong> application"
+TODO_MANED_ASSIGN_PENDING = "Assign <strong>pending<strong> application"
```
richard-jones: A bit of a misunderstanding here for the meaning of this file.  `constants` is for internal references within the code to the same thing, so that if we change the value of the variable, everything still hangs together.  It isn't "safe" to use this for display, we'd need to define a mapping from those values to the display value somewhere else (probably `ui.messages`).  

I complicated matters by displaying it, but that was only because there wasn't a UI for the todo list yet :)

richard-jones: I'm pushing a change to show what I mean

sssoz: No problem! I had originally done it that way but then I saw that some of the variables had the "human-readable" names (like `INDEX_RECORD_TYPE_UPDATE_REQUEST_UNFINISHED`), so I thought perhaps the template approach was wrong! 

Thanks for fixing.  

Is there anything else I should do, besides correcting the two conflicts?

sssoz: Is there anything else you needed me to do here?

sssoz:
@richard-jones I’ve resolved the conflicts and now have your latest changes made on `feature/3176_editorial_todo--maned`. Pushed on this branch and tested, everything looks good.

PR #2053, opened 2022-02-01
Hotfix/2022 02 01 single thread boto

Steven-Eardley:
https://github.com/DOAJ/doajPM/issues/3223


Steven-Eardley:
Since the change was via @richard-jones I reckon I'll just go ahead and merge this, already been reviewed and tested by both of us.

PR #2054, opened 2022-02-08
Quick review of ManEd UI design

sssoz:
Prettifying the UI a bit more before deploying it on test for ManEds to view. 

See DOAJ/doajPM#3175 and DOAJ/doajPM#3217

### DIFF
--- a/portality/static/js/doaj.fieldrender.edges.js
+++ b/portality/static/js/doaj.fieldrender.edges.js
@@ -3552,7 +3552,8 @@ $.extend(true, doaj, {
                     var result = '<p><a class="edit_suggestion_link button" href="';
                     result += params.editUrl;
                     result += resultobj['id'];
-                    result += '">' + linkName + '</a></p>';
+                    result += '" target="_blank"';
+                    result += '>' + linkName + '</a></p>';
                     return result;
                 }
                 return false;
@@ -3565,7 +3566,8 @@ $.extend(true, doaj, {
                     var result = '<br/><p><a class="readonly_journal_link button" href="';
                     result += params.readOnlyJournalUrl;
                     result += resultobj.admin.current_journal;
-                    result += '">View journal being updated</a></p>';
+                    result += '" target="_blank"';
+                    result += '>View journal being updated</a></p>';
                     return result;
                 }
                 return false;
@@ -3581,7 +3583,8 @@ $.extend(true, doaj, {
                     var result = '<p><a class="edit_journal_link button button--secondary" href="';
                     result += params.editUrl;
                     result += resultobj['id'];
-                    result += '">Edit this journal</a></p>';
+                    result += '" target="_blank"';
+                    result += '>Edit this journal</a></p>';
                     return result;
                 }
                 return false;
--- a/portality/static/js/doaj.fieldrender.edges.js
+++ b/portality/static/js/doaj.fieldrender.edges.js
@@ -3552,8 +3552,7 @@ $.extend(true, doaj, {
                     var result = '<p><a class="edit_suggestion_link button" href="';
                     result += params.editUrl;
                     result += resultobj['id'];
-                    result += '" target="_blank"';
-                    result += '>' + linkName + '</a></p>';
+                    result += '">' + linkName + '</a></p>';
                     return result;
                 }
                 return false;
@@ -3566,8 +3565,7 @@ $.extend(true, doaj, {
                     var result = '<br/><p><a class="readonly_journal_link button" href="';
                     result += params.readOnlyJournalUrl;
                     result += resultobj.admin.current_journal;
-                    result += '" target="_blank"';
-                    result += '>View journal being updated</a></p>';
+                    result += '">View journal being updated</a></p>';
                     return result;
                 }
                 return false;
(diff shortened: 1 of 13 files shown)


portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -3566,8 +3565,7 @@ $.extend(true, doaj, {
                     var result = '<br/><p><a class="readonly_journal_link button" href="';
                     result += params.readOnlyJournalUrl;
                     result += resultobj.admin.current_journal;
-                    result += '" target="_blank"';
-                    result += '>View journal being updated</a></p>';
+                    result += '">View journal being updated</a></p>';
```
richard-jones, line 3568: Are we sure about not opening these pages in new windows?  I personally think it's great to get rid of this, but it was there because it was wanted that way, are editorial happy with the change?

portality/templates/dashboard/index.html
```diff
@@ -1,41 +1,95 @@
...
       {# TODO only show tasks for this user’s groups #}
       {# TODO integrated priority in list display
       {{ todo.boost }}
        #}
-      <li class="card card--compact">
+       {% set app_url =  url_for('admin.application', application_id=todo.object_id) %}
+      <li
+        class="card card--compact todo-list__item"
+        data-app-url="{{ app_url }}"
+        {# TODO there’s probably a better way to do this... #}
```
richard-jones, line 30: oh yes, for sure :)  I'll push a change in a moment...

portality/templates/layouts/dashboard_base.html
```diff
@@ -51,10 +51,22 @@ <h2 class="sr-only">DOAJ Dashboard</h2>
             <span data-feather="smile" aria-hidden="true"></span>
           </a>
         </li>
+        <li>
+          <a href="/" class="tag tag--secondary">
+            <span class="sr-only">Home</span>
+            <span data-feather="home" aria-hidden="true"></span>
+          </a>
+        </li>
+        <li>
```
richard-jones, line 60: Is there any mileage in maybe including the entire front-end navigation system in a pull-down menu.  At least the parts that are built from configuration?  That way if I wanted to jump directly to the API docs, or the Journal search, I could go straight there rather than in 2 steps?

richard-jones (commented):
Mostly this is fine, I just attached a couple of questions, and I'll push a code refactor in a moment

portality/templates/layouts/dashboard_base.html
sssoz: We’ve already established this through the mockups — we want to diminish the noise that the current interface presents to users (all the public site menus etc are not relevant to work on the dashboard), so I want to go ahead with this and ensure that there is a clear distinction between public interface (searching journals and articles) vs dashboard interface (processing and managing applications); it’s central to the idea of the the editorial workflow overhaul. This may not be final, we’ll be testing navigation as soon as this is on the test server though.

richard-jones: I'm probably a niche user, but it's very very common for me to be switching regularly between the admin, editorial, publisher and public views, especially when debugging.

portality/templates/dashboard/index.html
richard-jones: I've pushed this change now, so if you pull the branch again you will get my approach

portality/templates/layouts/dashboard_base.html
sssoz: Ah that’s fair. Can we just remove it for now but I’ll carefully consider a design around that instead of just re-introducing the menu as it was? It’s possible that we just do like you suggested — collapse it in a dropdown — but I’d like to see exactly what parts of the menu would be useful. For example, having the journals and articles search sees repetitive. 

I’ll also see with the editorial team when we test.

portality/templates/dashboard/index.html
sssoz: Looks great! Thank you! I was trying to find a way to do that but really couldn’t figure it out :)

portality/templates/layouts/dashboard_base.html
richard-jones: Yes, sure, no worries

sssoz: Resolving because I’ve noted this comment in testing documents

portality/static/js/doaj.fieldrender.edges.js
sssoz: Idem

portality/templates/dashboard/index.html
```diff
@@ -1,41 +1,78 @@
...
 
 {% block content %}
 <main>
-
   <h1>Hi, {{ current_user.name }}!</h1>
 
   <section>
-    <h2 class="label">Here are your priorities:</h2>
-    <ol class="unstyled-list">
+    <h2 class="type-04">Here are your priorities ({{ todos|length }}):</h2>
```
richard-jones, line 37: I'm not sure you can use `todos|length` here, as it's not that meaningful.  It doesn't tell you how many todos you have, it just tells you how many are in the list on the page, which is a function of the number we've configured it to show.

sssoz: Oh I didn’t realise that! Is there a way to show the number of TODOs?

richard-jones: There is not, see some information about why buried in my comments here: https://github.com/DOAJ/doajPM/issues/3176#issuecomment-1018440621 and https://github.com/DOAJ/doajPM/issues/3176#issuecomment-1023070140

PR #2055, opened 2022-02-08
OG URLs need to be absolute to be displayed

sssoz:
See DOAJ/doajPM#3227


richard-jones (approved):

Steven-Eardley (changes requested):
these being relative paths was an indication that the variable wasn't being passed into a template, the fix is to pass it in with     

```
render_template('doaj/index.html', news=news, recent_journals=recent_journals, base_url=app.config.get('BASE_URL'))
```

sssoz:
Fixing `base_url` on another branch instead, as suggested by @Steven-Eardley

sssoz:
See new PR for this here: https://github.com/DOAJ/doaj/pull/2059

PR #2057, opened 2022-02-11
Feature/3177 editorial group info

richard-jones:
NOT FOR MERGE

but please do review, as it is now feature complete


portality/lib/edges.py
```diff
@@ -27,19 +27,15 @@ def query(self):
...
         if len(musts) > 0:
             query = {
                 "bool" : {
-                    "filter" : {
-                        "bool": {
-                            "must": musts
-                        }
-                    },
-                    "must" : query
+                    "must": musts
```
richard-jones, line 38: I cherry picked this change as a hotfix, which as since been modified, so this change will need to be resolved when we merge everything down to develop eventually

portality/static/js/doaj.js
```diff
@@ -9,9 +9,11 @@ var doaj = {
...
         var nav = document.querySelector(".secondary-nav__menu");
 
-        openMenu.addEventListener('click', function() {
-            nav.classList.toggle("secondary-nav__menu-toggle--active");
-        }, false);
+        if (openMenu) {
+            openMenu.addEventListener('click', function () {
+                nav.classList.toggle("secondary-nav__menu-toggle--active");
+            }, false);
+        }
```
richard-jones, line 16: This might actually be a general problem with the new admin layout, @sssoz you may want to pull this change into your UI dev branch

portality/templates/editor/index.html
```diff
@@ -1,45 +1,52 @@
...
+                <p class="label label--tertiary">Other associate editors:</p>
+                <ul class="tags">
+                    {% for ass in group.get_associate_accounts() %}
+                        <li class="tag"><a href="mailto:{{ass.email}}"><span data-feather="mail" aria-hidden="true"></span> {{ass.id}}</a></li>
+                    {% endfor %}
+                </ul>
+            {% endif %}
+        {% endfor %}
+    {% endif %}
+    <p><a href="mailto:{{managing_editor}}" class="button">Contact the Managing Editors</a></p>
```
sssoz, line 51: Confirming that page will be replaced with the to-do list. When can we do this? Should I do it on a separate branch? I would need some of the changes from `[feature/3175-review-man-ed-design`](https://github.com/DOAJ/doaj/pull/2054) to do this

sssoz:
@richard-jones Everything looks good, I’ve pulled in this branch onto my UI one, `feature/3175-review-man-ed-design` (https://github.com/DOAJ/doaj/pull/2054). My latest design changes also work well with the man. ed. role: 

Screenshot 2022-02-14 at 17 50 48

portality/templates/editor/index.html
richard-jones: We aren't planning to do the editors dashboard page for a while now, I understand, so we can just come back around to this after the dashboard has been released

richard-jones:
I've merged this PR as we agreed it was necessary before releasing the first version of the todo list

PR #2058, opened 2022-02-14
Update 'no results' message

sssoz:
See https://github.com/DOAJ/doajPM/issues/2924

### DIFF
--- a/portality/static/js/doaj.fieldrender.edges.js
+++ b/portality/static/js/doaj.fieldrender.edges.js
@@ -2195,9 +2195,15 @@ $.extend(true, doaj, {
                         this.currentQueryString = qs.queryString || "";
                     }
                 }
-                var frag = "<li class='alert'><p>You searched for <i>'";
+                var frag = "<li class='alert'><p>You searched for <i>";
                 frag += edges.escapeHtml(this.currentQueryString);
-                frag +="'</i> and we found no results.</p><p>Search terms must be in <strong>English</strong>.</p> <p>Try removing some of the filters you have set, modifying the text in the search box, or using less specific search terms.</p></li>";;
+                frag += "</i> and we found no results.</p>";
+                frag += "<p>Please check the following:</p><ul>";
+                frag += "<li>Check the spelling and make sure that there are no missing characters.</li>";
+                frag += "<li>Use fewer words in your search to make the search less specific.</li>";
+                frag += "<li>Try removing some of the filters you have set.</li>";
+                frag += "<li>Retry your search in English as much of the index uses English terms.</li>";
+                frag += "</ul></li>";
 
                 if (this.component.results === false) {
                     frag = "";
--- a/portality/static/widget/fq_widget_depends_compiled.js
+++ b/portality/static/widget/fq_widget_depends_compiled.js
@@ -6675,9 +6675,15 @@ $.extend(true, doaj, {
                         this.currentQueryString = qs.queryString || "";
                     }
                 }
-                var frag = "<li class='alert'><p>You searched for <i>'";
+                var frag = "<li class='alert'><p>You searched for <i>";
                 frag += edges.escapeHtml(this.currentQueryString);
-                frag +="'</i> and we found no results.</p><p>Search terms must be in <strong>English</strong>.</p> <p>Try removing some of the filters you have set, modifying the text in the search box, or using less specific search terms.</p></li>";;
+                frag += "</i> and we found no results.</p>";
+                frag += "<p>Please check the following:</p><ul>";
+                frag += "<li>Check the spelling and make sure that there are no missing characters.</li>";
+                frag += "<li>Use fewer words in your search to make the search less specific.</li>";
+                frag += "<li>Try removing some of the filters you have set.</li>";
+                frag += "<li>Retry your search in English as much of the index uses English terms.</li>";
+                frag += "</ul></li>";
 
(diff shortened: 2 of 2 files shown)


portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -2195,9 +2195,15 @@ $.extend(true, doaj, {
                         this.currentQueryString = qs.queryString || "";
                     }
                 }
-                var frag = "<li class='alert'><p>You searched for <i>'";
+                var frag = "<li class='alert'><p>You searched for <i>";
                 frag += edges.escapeHtml(this.currentQueryString);
-                frag +="'</i> and we found no results.</p><p>Search terms must be in <strong>English</strong>.</p> <p>Try removing some of the filters you have set, modifying the text in the search box, or using less specific search terms.</p></li>";;
+                frag += "</i> and we found no results.</p>";
+                frag += "<p>Please check the following:</p><ul>";
```
Steven-Eardley, line 2201: Check the following / check the spelling is a bit too repetetive for my tastes, but that's not a blocker!

portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -2195,9 +2195,15 @@ $.extend(true, doaj, {
                         this.currentQueryString = qs.queryString || "";
                     }
                 }
-                var frag = "<li class='alert'><p>You searched for <i>'";
+                var frag = "<li class='alert'><p>You searched for <i>";
```
Steven-Eardley, line 2198: I'm in 2 minds, the apostrophes do add incorrect semantics to the search, looking like an exact phrase query, but it now looks a bit odd if you produce a search with only whitespace, e.g. a space.

Steven-Eardley (changes requested):
The message on the publisher journal search is now a bit misleading, since the message appears there when we haven't made a search.

portality/static/js/doaj.fieldrender.edges.js
sssoz: Ah I wasn’t sure why there was an extra typewriter apostrophe `'` — figured it was an error. Happy with either as long as the markup comes out nicely!

sssoz:
> The message on the publisher journal search is now a bit misleading, since the message appears there when we haven't made a search.

@Steven-Eardley Are you referring to cases like the "Articles" section of a journal’s TOC when there are no articles indexed? 

Screenshot 2022-02-14 at 16 49 14

portality/static/js/doaj.fieldrender.edges.js
sssoz: OH Nevermind my comment — I understand now. Did not mean to remove it.

sssoz: Done in https://github.com/DOAJ/doaj/pull/2058/commits/19694a14b029024a6a6a9ec65ad71d307bff3bc5

portality/static/js/doaj.fieldrender.edges.js
dommitchell: Change to

'Please *try* the following:' and then 
'Remove some of the filters you have set' and 
'Do your search again in English as...'

Steven-Eardley:
I was referring to the search at /publisher/journal - suppose we won't have many publishers without journals so it's probably not a necessary consideration (and if they've done a search there the message makes sense)

portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -2195,9 +2195,15 @@ $.extend(true, doaj, {
...
+                var frag = "<li class='alert'><p>You searched for ‘<i>";
                 frag += edges.escapeHtml(this.currentQueryString);
-                frag +="'</i> and we found no results.</p><p>Search terms must be in <strong>English</strong>.</p> <p>Try removing some of the filters you have set, modifying the text in the search box, or using less specific search terms.</p></li>";;
+                frag += "</i>’ and we found no results.</p>";
+                frag += "<p>Please check the following:</p><ul>";
+                frag += "<li>Check the spelling and make sure that there are no missing characters.</li>";
+                frag += "<li>Use fewer words in your search to make the search less specific.</li>";
+                frag += "<li>Try removing some of the filters you have set.</li>";
+                frag += "<li>Retry your search in English as much of the index uses English terms.</li>";
+                frag += "</ul></li>";
```
richard-jones, line 2206: Also not a blocker, but for such a large bit of text, we'd be better off using a multi-line string, as it is a bit more efficient in the client

```
frag += "<p>Please check the following:</p><ul>\
    <li>Check the spelling and make sure that there are no missing characters.</li>\
    <li>Use fewer words in your search to make the search less specific.</li>\
";

etc.  That way we can maintain the html indentation too, which makes it easier to read.

richard-jones (approved):
Minor code style comment, otherwise technically fine

portality/static/js/doaj.fieldrender.edges.js
sssoz: Done!

portality/static/js/doaj.fieldrender.edges.js
sssoz: Done!

sssoz:
Copy has been adjusted

Screenshot 2022-02-22 at 11 45 14

Steven-Eardley (approved):

PR #2061, opened 2022-02-24
Changed feedback@ to helpdesk@

RK206:
Change support email address from feedback@ to helpdesk@ #3226


Steven-Eardley (approved):
Good find-and-replace skills ;) 

On release, must check the .cfg files to see if the key is overridden there.

PR #2063, opened 2022-03-07
first pass at a working update of the anon export

richard-jones:


Steven-Eardley (approved):
This is successfully running on the test server, sending it live to run overnight.

Steven-Eardley:
also, tests passed running locally

PR #2075, opened 2022-04-13
Feature/editorial workflow consumers rama

RK206:


portality/events/consumers/bg_job_finished_notify.py
```diff
@@ -0,0 +1,48 @@
...
+            return
+
+        acc = models.Account.pull(job.user)
+        if acc is None or not acc.has_role("admin"):
+            return
+
+        svc = DOAJ.notificationsService()
+
+        query = Facetview2.make_query(job.id)
+        url = "admin/background_jobs?source=" + Facetview2.url_encode_query(query)
```
richard-jones, line 39: Here we should use the new `edges` url builder, you should be able to do something like

`string_id_query = edges.make_url_query(query_string=job.id)`

instead of using the Facetview2 library

richard-jones: Also, we should build these with `url_for` so someting like

`url_for("admin.background_jobs", source=string_id_query)`

portality/background.py
```diff
@@ -79,31 +81,37 @@ def execute(self, background_task):
...
+        #
+        #         app_email.send_mail(to=to,
+        #                             fro=fro,
+        #                             subject=subject,
+        #                             template_name=template,
+        #                             job_id=job.id,
+        #                             action=job.action,
+        #                             status=job.status,
+        #                             background_job_url=url
+        #                             )
```
richard-jones, line 114: I think we can just delete this commented out code

richard-jones (changes requested):
Really minor changes, otherwise all good

portality/events/consumers/bg_job_finished_notify.py
RK206: Updated to edges

portality/background.py
RK206: Removed

richard-jones (approved):

PR #2080, opened 2022-04-21
Feature/consumer publisher created

RK206:


cms/data/notifications.yml
```diff
@@ -21,3 +21,22 @@ application:maned:ready:notify:
...
+    We receive hundreds of applications every month. We will review yours as soon as we can. You will receive an email from us when the review is complete or if we have questions.
+    If you need to make a change to your application, send an email to our Help desk: helpdesk@doaj.org. Please include your ISSN.
+    We cannot reply to emails requesting a status update for applications which are less than three months old.
+    If you write to us, check first that there is nothing in your Spam folder from us or one of our volunteers. Our volunteers may not be emailing from a DOAJ email address so you can check that their name is there in the url mentioned below {volunteers_url}
+
+    Yours faithfully,
+    The DOAJ Editorial Team
+
+    ***
+    This is an automated message."
```
richard-jones, line 40: No need to add the sign off here, that's added by the notification emailer.

As an aside, we may want to look at shortening the text here to feel more like a notification, will raise that separately.

cms/data/notifications.yml
```diff
@@ -21,3 +21,22 @@ application:maned:ready:notify:
...
+  email_subject:
+    "Directory of Open Access Journals - your application to DOAJ has been received"
```
richard-jones, line 42: No need to prefix with Directory of Open Access Journals - this is done by the notifier

portality/constants.py
```diff
@@ -18,7 +19,8 @@
     APPLICATION_STATUS_IN_PROGRESS,
     APPLICATION_STATUS_COMPLETED,
     APPLICATION_STATUS_ON_HOLD,
-    APPLICATION_STATUS_READY
+    APPLICATION_STATUS_READY,
+    APPLICATION_STATUS_CREATED
```
richard-jones, line 23: Not a new status, so shouldn't be here

portality/constants.py
```diff
@@ -8,6 +8,7 @@
 APPLICATION_STATUS_COMPLETED = "completed"
 APPLICATION_STATUS_ON_HOLD = "on hold"
 APPLICATION_STATUS_READY = "ready"
+APPLICATION_STATUS_CREATED = "created"
```
richard-jones, line 11: This isn't a new status.  I think the confusion here is that this notification has been wired to a "created" status, when in fact what we want is an event for the creation of a new application.  The initial application status will be "pending" or whatever it is set to.

portality/events/consumers/application_publisher_created_notify.py
```diff
@@ -0,0 +1,45 @@
...
+from portality.lib.seamless import SeamlessException
+
+
+class ApplicationPublisherCreatedNotify(EventConsumer):
+    ID = "application:publisher:created:notify"
+
+    @classmethod
+    def consumes(cls, event):
+        return event.id == constants.EVENT_APPLICATION_STATUS and \
+               event.context.get("status") == constants.APPLICATION_STATUS_CREATED
```
richard-jones, line 17: I think we should be cosnuming an `application:created` event here

portality/forms/application_processors.py
```diff
@@ -258,12 +258,12 @@ def finalise(self, account, save_target=True, email_alert=True, id=None):
...
-                emails.send_received_email(self.target)
-            except app_email.EmailException as e:
-                self.add_alert(Messages.FORMS__APPLICATION_PROCESSORS__NEW_APPLICATION__FINALISE__USER_EMAIL_ERROR)
-                app.logger.exception(Messages.FORMS__APPLICATION_PROCESSORS__NEW_APPLICATION__FINALISE__LOG_EMAIL_ERROR)
+        # trigger a status change event
+        eventsSvc = DOAJ.eventsService()
+        eventsSvc.trigger(models.Event(constants.EVENT_APPLICATION_STATUS, None, {
+            "application": self.target.data,
+            "status": constants.APPLICATION_STATUS_CREATED
+        }))
```
richard-jones, line 266: This should be

```
eventsSvc.trigger(models.Event(constants.EVENT_APPLICATION_CREATE, current_user.id, {
            "application": self.target.data
}))
```

richard-jones (changes requested):
A bit of a mix up here on the event type.  I know that Rama is away so I may come back and fix this one up myself shortly

richard-jones:
I have now made these changes, so @Steven-Eardley @amdomanska perhaps you could give it a review now?

Steven-Eardley (approved):

PR #2081, opened 2022-04-21
Notification to the publisher when the update request requires revisions

amdomanska:
Notification to the publisher when the update request requires revisions (row 11 in https://docs.google.com/spreadsheets/d/1oIeG5vg-blm2MZCE-7YhwulUlSz6TOUeY8jAftdP9JE/edit#gid=1277065021)


portality/events/consumers/application_publisher_revision_notify.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class ApplicationPublisherRevisionNotify(EventConsumer):
+    ID = "application:publisher:revision:notify"
+
+    @classmethod
+    def consumes(cls, event):
+        return event.id == constants.EVENT_APPLICATION_STATUS and \
+                event.context.get("old_status") != constants.APPLICATION_STATUS_REVISIONS_REQUIRED and \
+                event.context.get("new_status") == constants.APPLICATION_STATUS_REVISIONS_REQUIRED
```
richard-jones, line 18: I would check for the presence of an `application` context key here too

portality/events/consumers/application_publisher_revision_notify.py
```diff
@@ -0,0 +1,52 @@
...
+    @classmethod
+    def consume(cls, event):
+        context = event.context
+        app_id = context.get("application")
+        if app_id is None:
+            return
+        try:
+            application = models.Application.pull(app_id)
+        except SeamlessException:
+            raise exceptions.NoSuchObjectException("Unable to locate application with id {x}".format(x=app_id))
```
richard-jones, line 29: This is using the old approach, we should now be pulling the application source from the `application` context key, like

```
app_source = event.context.get("application")

        try:
            application = models.Application(**app_source)
        except Exception as e:
            raise exceptions.NoSuchObjectException("Unable to construct Application from supplied source - data structure validation error, {x}".format(x=e))
```

richard-jones: Actually, there's a bit of a mix here - you have the SeamlessException but are not constructing the object from the passed data.

portality/events/consumers/application_publisher_revision_notify.py
```diff
@@ -0,0 +1,52 @@
...
+        if owner is None:
+            raise exceptions.NoSuchPropertyException("No owner found for application with id {x}".format(x=app_id))
+
+        svc = DOAJ.notificationsService()
+
+        notification = models.Notification()
+        notification.who = owner
+        notification.created_by = cls.ID
+        notification.classification = constants.NOTIFICATION_CLASSIFICATION_STATUS_CHANGE
+        from datetime import datetime
```
richard-jones, line 44: Inline/delayed import isn't necessary here, just import at the top of the file as usual

doajtest/unit/event_consumers/test_application_publisher_revision_notify.py
```diff
@@ -0,0 +1,65 @@
...
+        source = ApplicationFixtureFactory.make_application_source()
+        app = models.Application(**source)
+        app.save()
+
+        acc = models.Account()
+        acc.set_id("publisher")
+        acc.set_email("test@example.com")
+        acc.save()
+
+        event = models.Event(constants.EVENT_APPLICATION_STATUS, context={"application": "abcdefghijk", "old_status": "in progress", "new_status": "revisions_required"})
```
richard-jones, line 46: the `application` context argument needs to be the core data structure for an actual application to which the event occurred.  You can do

```
source = ApplicationFixtureFactory.make_application_source()
app = models.Application(**source)
event = models.Event(constants.EVENT_APPLICATION_STATUS, context={"application": app.data, "old_status": "in progress", "new_status": "revisions_required"})
```

richard-jones (changes requested):
Need to move to using the "pass objects to which events happened in full" approach, this is a bit of a mix between passing by ID and passing by value.

portality/events/consumers/application_publisher_revision_notify.py
amdomanska: I'm checking for that in a `consume` method consistently with other notification consumers' implementation.

amdomanska:
@richard-jones Ready for re-review, please.

richard-jones (approved):

Steven-Eardley (approved):

PR #2084, opened 2022-04-27
Feature/3245 ur app tabs

amdomanska:
https://github.com/DOAJ/doajPM/issues/3245

Migration rules:

    Look for applications which are rejected
    If they have a related_journal then they must be an update request


portality/migrate/3245_ur_vs_app/operations.py
```diff
@@ -0,0 +1,7 @@
+from portality import constants
+def make_update_request(application):
+    if application.related_journal:
```
richard-jones, line 3: I'd add a `and not application.application_type == constants.APPLICATION_TYPE_UPDATE_REQUEST` check here too, just to make sure we're not calling save on applications that don't need it

richard-jones (changes requested):
Minor change to improve performance and unnecessary work on the index, otherwise good to go

Steven-Eardley (approved):

richard-jones (approved):

PR #2087, opened 2022-04-29
Feature/application publisher accepted rejected notifications

amdomanska:
~~application:publisher:accepted:notify~~
~~Create a notification for the owner of the journal when the status is switched to Accepted~~
_removed as duplicated_

application:publisher:rejected:notify
Create a notification to the publisher when the application is rejected

https://docs.google.com/spreadsheets/d/1oIeG5vg-blm2MZCE-7YhwulUlSz6TOUeY8jAftdP9JE/edit#gid=1277065021 (rows 12 & 14)


richard-jones (commented):
It turns out that two of the consumers turned out to be the same, and we may have both implemeted the same thing.  I think we'll need to unpick that one together during the week next week.

portality/bll/services/events.py
```diff
@@ -22,8 +24,7 @@ class EventsService(object):
         ApplicationEditorCompletedNotify,
         ApplicationEditorInProgressNotify,
         ApplicationEditorGroupAssignedNotify,
-        BGJobFinishedNotify,
-        ApplicationManedReadyNotify
```
amdomanska, line 26: This is duplicated with line 21

amdomanska, line 26: This is duplicated with line 21

doajtest/unit/event_consumers/test_application_publisher_rejected_notify.py
```diff
@@ -0,0 +1,60 @@
...
+
+        n = ns[0]
+        assert n.who == "publisher"
+        assert n.created_by == ApplicationPublisherRejectedNotify.ID
+        assert n.classification == constants.NOTIFICATION_CLASSIFICATION_STATUS_CHANGE
+        assert n.message is not None
+        assert not n.is_seen()
+
+    def test_consume_fail(self):
+        event = models.Event(constants.EVENT_APPLICATION_ASSED_ASSIGNED, context={"application": {"key" : "value"}})
```
richard-jones, line 57: Technically the wrong event, though it doesn't actually impact on the test execution and outcome here

portality/forms/application_processors.py
```diff
@@ -373,7 +376,12 @@ def finalise(self, account, save_target=True, email_alert=True):
                 sent = False
                 send_report = []
                 try:
-                    send_report = emails.send_publisher_reject_email(self.target, update_request=is_update_request)
+                    if self.source.application_status != self.target.application_status:
+                        eventsSvc.trigger(models.Event(constants.EVENT_APPLICATION_STATUS, account.id, {
+                            "application": self.target.data,
+                            "old_status": self.source.application_status,
+                            "new_status": self.target.application_status
+                        }))
```
richard-jones, line 384: This trigger isn't required here, as the application status change event is triggered at the end of the function anyway

portality/view/admin.py
```diff
@@ -435,7 +435,12 @@ def application_quick_reject(application_id):
     sent = False
     send_report = []
     try:
-        send_report = emails.send_publisher_reject_email(application, note=reason, update_request=update_request)
+        if self.source.application_status != self.target.application_status:
+            eventsSvc.trigger(models.Event(constants.EVENT_APPLICATION_STATUS, account.id, {
+                "application": self.target.data,
+                "old_status": self.source.application_status,
+                "new_status": self.target.application_status
```
richard-jones, line 442: self.target and self.source don't exist here, right?  You'll need to pass `application.data` and then get the before application_status from the application when you first load it.  Then the new status is obviously rejected, which we can take from the `constants` module

richard-jones (changes requested):

PR #2088, opened 2022-04-29
Prefer alpha 2 codes when looking up languages & update pycountry

Steven-Eardley:
Prerequisite for updating `pycountry`


portality/lib/isolang.py
```diff
@@ -1,15 +1,18 @@
...
-        return {}
+def find(lang, as_dict=True):
+    """" Look up a language preferably by alpha2 code, or any of its attributes """
+    language = pycountry.languages.get(alpha_2=lang)
+    if language is None:
+        try:
+            language = pycountry.languages.lookup(lang)
+        except LookupError:
+            language = None
+    return _as_dict(language) if as_dict else language
```
richard-jones, line 12: I am not a massive fan of this variable return-type approach, if it can be avoided.

One option would be for the caller to make it a dict if that's what they want.  Another would be to wrap the function with other behaviours, like

```python
def find(lang):
   _as_dict(find_raw(lang))

def find_raw(lang):
    language = pycountry.languages.get(alpha_2=lang)
    #...
```

Then anyone who wants the dict uses the original `find` method (which is presumably everywhere except the instances you are changing), and those that don't want the dict can use the `find_raw` method

richard-jones (approved):
One suggested change, not necessarily blocking

PR #2094, opened 2022-05-06
Feature/3242 anon export task

philipkcl:
For [issue 3242](https://github.com/DOAJ/doajPM/issues/3242)


portality/background_helper.py
```diff
@@ -0,0 +1,48 @@
+""" collections of wrapper function for helping you to create BackgroundTask
```
richard-jones, line 1: Can I suggest that we put this somewhere other than the root of the repo.  I propose in either `portality.lib` or `portality.tasks`.  I'm leaning towards the latter, but open to persuasion.

portality/tasks/anon_export.py
```diff
@@ -0,0 +1,202 @@
...
+                                            action=cls.__action__,
+                                            params=params)
+
+    @classmethod
+    def submit(cls, background_job):
+        background_job.save()
+        on_submit_anon_export.schedule(args=(background_job.id,), delay=10)
+
+
+@main_queue.periodic_task(schedule(AnonExportBackgroundTask.__action__))
```
richard-jones, line 190: as per other comment, suggest this goes on the long running queue

portality/tasks/anon_export.py
```diff
@@ -0,0 +1,202 @@
...
+@write_required(script=True)
+def scheduled_anon_export():
+    background_helper.submit_by_bg_task_type(AnonExportBackgroundTask,
+                                             clean=app.config.get("TASKS_ANON_EXPORT_CLEAN", False),
+                                             limit=app.config.get("TASKS_ANON_EXPORT_LIMIT", None),
+                                             batch_size=app.config.get("TASKS_ANON_EXPORT_BATCH_SIZE", 100000))
+
+
+@main_queue.task()
```
richard-jones, line 199: as per other comment, suggest this goes on the long running queue

test.cfg
```diff
@@ -52,10 +52,17 @@ HUEY_SCHEDULE = {
...
 # =======================
 # Account settings
 PUBLIC_REGISTER = True
 LOGIN_VIA_ACCOUNT_ID = True
+
+#########################################################
+# Background tasks --- anon export
+TASKS_ANON_EXPORT_CLEAN = False
+TASKS_ANON_EXPORT_LIMIT = None
+TASKS_ANON_EXPORT_BATCH_SIZE = 100000
```
richard-jones, line 68: Since these properties don't change the defaults in `settings.py` I think they can be omitted

richard-jones (changes requested):
Very minor structural changes from me, otherwise this looks good to go.

@Steven-Eardley we'll want to run this up on test and give it a bit more of a workout there

doajtest/unit/test_tasks_anon_export.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def test_execute(self):
+
+        # prepare test data
+        for _ in range(3):
+            BackgroundJob().save(blocking=True)
+        new_background_jobs = list(BackgroundJob.scroll())
+
+        for _ in range(2):
+            Account().save(blocking=True)
```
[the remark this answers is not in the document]
philipkcl: so, elasticsearch process all save request by signal queue? I thought they save all record with mult-thread.

portality/background_helper.py
philipkcl: `portality.lib` seem contain set of function that can be reuse / share to other project. but common pattern of background task could be only apply in DOAJ.
`portality.tasks` seem only contain file of `BackgroundTask`, I will move it to `portality.tasks.helpers`

philipkcl:
@richard-jones update PR for feedback, please feel free to review

doajtest/unit/test_tasks_anon_export.py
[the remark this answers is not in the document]
richard-jones: Hm, I don't know how it queues them :) All I can say is that we've never had a problem assuming that once the last to save is done, then everything before it is also done.  You can keep it or change it, I don't mind, it just makes the test slow is all.

richard-jones (approved):

philipkcl:
@richard-jones and @Steven-Eardley  Some update for this branch. just let you known in case it already in testing stage.
just find out I should use `long_running` for `consumer_long_running.py`

richard-jones:
Looks good to me.  This is now just queued for test

PR #2097, opened 2022-05-11
Feature/3261 analytics plausible

philipkcl:
For [Switch our analytics to Plausible #3261](https://github.com/DOAJ/doajPM/issues/3261)

* We can use Plausible docker for testing Plausible features.  
  * https://plausible.io/docs/self-hosting
  * https://github.com/plausible/hosting
* seem like Plausible only support target sites that using `https`
for testing, change code in app.py and install  `pip install pyopenssl`
```app.py 
app.run(host=app.config['HOST'], debug=app.config['DEBUG'], port=app.config['PORT'],
            ssl_context='adhoc')
```
* there are two environment variables added for Plausible
```
# root url for download plausible js
PLAUSIBLE_URL = "http://localhost:8000"
# site name / domain name that used to register in plausible
PLAUSIBLE_SITE_NAME = BASE_DOMAIN
```


richard-jones:
This looks very good.  I have not yet tested it with my own local Plausible instance, and we need to talk about migration from GA, so I'm going to leave this open for the moment and we'll make the plan for migration.

portality/templates/_js_includes.html
```diff
@@ -8,16 +8,11 @@
...
-    <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-
-      gtag('config', 'UA-46560124-1');
-    </script>
+    <!-- Plausible Analytics -->
+    <script defer data-domain="{{ config.get('PLAUSIBLE_SITE_NAME')}}"
+            src="{{ config.get('PLAUSIBLE_URL') }}/js/plausible.js"></script>
```
Steven-Eardley, line 15: Maybe consider providing the full URI (including path) in the settings, rather than hard-coding in the template

Steven-Eardley (approved):
This looks good, one small comment about hard-coding the path

philipkcl:
update for plausible env variable

PR #2098, opened 2022-05-11
Hotfix/3273 and 3285  journal withdrawn deleted banner

richard-jones:
This hotfix deals with two issues:
* https://github.com/DOAJ/doajPM/issues/3273
* https://github.com/DOAJ/doajPM/issues/3285

It is the result of merging hotfix/3273_warning_message_withdrawn_or_deleted which was used as the base branch to create this one and hotfix/20220509_2133_issue3285 which was manually merged (i.e. its version history is not included here)

### DIFF
--- a/portality/forms/application_processors.py
+++ b/portality/forms/application_processors.py
@@ -8,7 +8,7 @@ from portality.lib.formulaic import FormProcessor
 from portality.ui.messages import Messages
 from portality.crosswalks.application_form import ApplicationFormXWalk
 from portality.crosswalks.journal_form import JournalFormXWalk
-from portality.bll import exceptions, DOAJ
+from portality.bll import exceptions
 
 from flask import url_for, request, has_request_context
 from flask_login import current_user
@@ -301,13 +301,15 @@ class AdminApplication(ApplicationProcessor):
 
         if self.source is None:
             raise Exception(Messages.EXCEPTION_EDITING_NON_EXISTING_APPLICATION)
-        if self.source.current_journal == constants.APPLICATION_STATUS_ACCEPTED:
+
+        if self.source.application_status == constants.APPLICATION_STATUS_ACCEPTED:
             raise Exception(Messages.EXCEPTION_EDITING_ACCEPTED_JOURNAL)
+
         if self.source.current_journal is not None:
-            journal_status = DOAJ.journalService().exist_status(journal_id=self.source.current_journal)
-            if journal_status.is_deleted():
+            j = models.Journal.pull(self.source.current_journal)
+            if j is None:
                 raise Exception(Messages.EXCEPTION_EDITING_DELETED_JOURNAL)
-            elif journal_status.is_withdrawn():
+            elif not j.is_in_doaj():
                 raise Exception(Messages.EXCEPTION_EDITING_WITHDRAWN_JOURNAL)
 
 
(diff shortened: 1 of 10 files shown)


portality/templates/application_form/editorial_form_body.html
```diff
@@ -3,12 +3,10 @@ <h1>{{ obj.bibjson().title }}</h1>
 {% include "application_form/_backend_validation.html" %}
 {% include "application_form/_contact.html" %}
 
-{% if obj["es_type"] == "application" and obj.application_type == "update_request" and (not current_journal or not current_journal.is_in_doaj()) %}
-    <h2 class="alert">
-        <span data-feather="alert-triangle" aria-hidden="true"></span>
-        Journal has been withdrawn or deleted. This update request was automatically rejected.
-    </h2>
-{% endif %}
+{% set related_journal = obj.related_journal_object %}
```
philipkcl, line 6: this variable `related_journal` seem like never used

portality/templates/application_form/_application_warning_msg.html
```diff
@@ -0,0 +1,19 @@
...
+        and obj.application_type == "update_request"
+        and obj.application_status == constants.APPLICATION_STATUS_REJECTED %}
+
+        {% set related_journal = obj.related_journal_object %}
+
+        <h2 class="alert">
+            <span data-feather="alert-triangle" aria-hidden="true"></span>
+            {% if not related_journal %}
+                Journal has been deleted. This update request was automatically rejected.
+            {% else if not related_journal.is_in_doaj() %}
```
philipkcl, line 13: syntax error use `elif` instead of `else if`

portality/forms/application_processors.py
```diff
@@ -301,12 +301,16 @@ def finalise(self, account, save_target=True, email_alert=True):
 
         if self.source is None:
             raise Exception(Messages.EXCEPTION_EDITING_NON_EXISTING_APPLICATION)
-        if self.source.current_journal == constants.APPLICATION_STATUS_ACCEPTED:
+
+        if self.source.application_status == constants.APPLICATION_STATUS_ACCEPTED:
             raise Exception(Messages.EXCEPTION_EDITING_ACCEPTED_JOURNAL)
+
         if self.source.current_journal is not None:
```
philipkcl, line 308: * what is difference between `related_journal_object` and `current_journal`? 
* should we following logic in frontend `portality/templates/application_form/_application_warning_msg.html` use `related_journal_object`?

richard-jones: > what is difference between related_journal_object and current_journal?

`current_journal` is a field which contains the identifier of a journal *if the application is an active update request that's being reviewed by editorial*

`related_journal` is a field which contains the identifier of a journal *if the application is a completed update request that has been accepted or rejected*

`related_journal_object` gets the `models.Journal` object for the journal identified in `related_journal`.  I've added it as a function because it is necessary to access that data sometimes via the jinja2 templates, and that makes it easy.

> should we following logic in frontend portality/templates/application_form/_application_warning_msg.html use related_journal_object?

In this case this is the correct thing to do, because the update request is not accepted, so it will either be rejected or have a current_journal, and we don't need to show an error message if the journal is rejected, because editing it makes no difference.  So I think this is fine.

portality/templates/application_form/editorial_form_body.html
richard-jones: yup, good spot, must have been left over from the merge

richard-jones:
I've made a few changes from @philipkcl's review, I think this is ready to go @Steven-Eardley

PR #2101, opened 2022-05-23
Style editorial status section

sssoz:
See DOAJ/doajPM#3178

### DIFF
--- a/cms/sass/components/_progress-bar.scss
+++ b/cms/sass/components/_progress-bar.scss
@@ -1,0 +1,40 @@
+.progress-bar {
+  width: 100%;
+}
+
+.progress-bar__bar {
+  display: inline-flex;
+
+  .progress-bar__link {
+    display: flex;
+    align-items: center;
+    padding: 0 $spacing-03;
+    height: 50px;
+    text-decoration: none;
+    color: $warm-black;
+  }
+}
+
+.progress-bar__bar--pending {
+  background-color: $yellow;
+}
+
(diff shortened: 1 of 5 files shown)


sssoz:
@richard-jones 
As part of https://github.com/DOAJ/doajPM/issues/3178, I’ve styled the editorial information group and took out the UR information so that it can be separated.

Just so we can continue to advance on this, I’m submitting this PR that *only* displays the editorial group information and will be continuing to work on displaying the UR info in a separate PR. I hope this is OK.

There are also a few details about this PR that you’ll probably notice:
1. I wasn’t sure how to display the total number of applications in the tab header of an editor group (right now there’s just a static number next to each group — "Spanish (10)"). This is so that people can get an immediate glance at the total numbers for each group without having to click around. 
2. There are a few a11y issues that are a bit longer to explain, but I’d rather have this go out sooner and then continue to work on accessibility later on. Some things I am unsure how to work out here, e.g. dynamically adding ARIA attributes to the currently open/active tab.
3. Users should be able to see the content of the first tab by default without having to click anything. In my screenshot, it would show the contents of Southeast Asia’s application info by default. 
4. (Optional — just a question): is it possible to order these tabs in a certain order? For example, ordering them by number of applications, with the most ongoing apps first?

Screenshot 2022-06-13 at 00 09 00

Thanks!

sssoz:
Another screen recording for @dommitchell and @jbarnsby to see: 

63a07831621f1c3a97c9c738b5100e4b

PR #2115, opened 2022-06-17
Style notifications

sssoz:
See DOAJ/doajPM#3282 and DOAJ/doajPM#3283

### DIFF
--- a/cms/sass/components/_dropdown.scss
+++ b/cms/sass/components/_dropdown.scss
@@ -35,26 +35,6 @@
   }
 }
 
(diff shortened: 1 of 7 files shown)


sssoz:
@richard-jones there are a few dates that are not in a human-readable format. Could we make sure to implement that (in the notification item itself, and the full list of notifications... and if possible everywhere else in the search interfaces?). It should be in the format: `[numeric day] [long month] [numeric year]`, e.g. 2 January, 2022.

PR #2120, opened 2022-07-07
Feature/2727 enhance unittest v2

philipkcl:
For https://github.com/DOAJ/doajPM/issues/2727,

* added `StoreLocalPatcher` to avoid used same folder in difference unit test.
* added `create_tmp_dir` for create new folder  with random name


richard-jones (commented):
I have tried to run this locally, and it generated a large amount of test failures and errors, and then hung about half way through.  I'll pull some diagnostics out and put them in the main issue.

PR #2121, opened 2022-07-08
Search for rejected update requests of journals that are no longer in doaj

amdomanska:
Script for https://github.com/DOAJ/doajPM/issues/3285#issuecomment-1168500345


portality/scripts/3285_non_rejected_ur_of_journal_not_in_doaj.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    urs = []
+    for app in models.Application.iterate():
+        if app.application_type != constants.APPLICATION_TYPE_UPDATE_REQUEST:
+            continue
+
+        if app.application_status != constants.APPLICATION_STATUS_REJECTED:
+            if app.current_journal:
+                j = models.Journal.pull(app.current_journal)
+                if j.is_in_doaj():
```
richard-jones, line 18: This should also check for where `j is None`

richard-jones (changes requested):
Also catch cases where the journal does not exist at all, otherwise good

Steven-Eardley:
@richard-jones I've added the requested extra info to the script: sample run on my system:

```
Application ID,Reason
ea5288782ec1480598f46269f87b1166,Journal a5a148d9dc5b47feaa7726fbfa14505e not in DOAJ
4ac287595cff4313878373f8913408e3,Journal 59a2f57c22d54d00a572e37c22520b26 not in DOAJ
be78dd4f9d344944a9c48c484ced6439,Journal ca9eff28c8fd485ab548d4a95b29925f not in DOAJ
fe777edbbe934b44af2e06583934aeba,Journal 502f32af34784239845c9690cbc7b2b2 not in DOAJ
```

PR #2129, opened 2022-07-26
Tag should be a class, not an id

sssoz:
See DOAJ/doajPM#3335


richard-jones (approved):
This looks good to me, ready to go live

Steven-Eardley (approved):

PR #2140, opened 2022-08-23
script to reject update request id 5742fbfb237f4b38a7b58d54951727d6

amdomanska:
Script to reject update request id `5742fbfb237f4b38a7b58d54951727d6` for journal `56d3896d58c7400eb1fae581ecd2e8d6` (withdrawn from DOAJ), as per https://github.com/DOAJ/doajPM/issues/3285#issuecomment-1205443986

### DIFF
--- a/portality/scripts/reject_update_requst.py
+++ b/portality/scripts/reject_update_requst.py
@@ -1,0 +1,9 @@
+from portality.models import Application
+from portality.constants import APPLICATION_STATUS_REJECTED
+
+UR_TO_REJECT = "5742fbfb237f4b38a7b58d54951727d6"
+
+if __name__ == "__main__":
+    ur = Application.pull(UR_TO_REJECT)
+    ur.set_application_status(APPLICATION_STATUS_REJECTED)
+    ur.save()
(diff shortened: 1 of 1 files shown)


portality/scripts/reject_update_requst.py
```diff
@@ -0,0 +1,9 @@
+from portality.models import Application
+from portality.constants import APPLICATION_STATUS_REJECTED
+
+UR_TO_REJECT = "5742fbfb237f4b38a7b58d54951727d6"
+
+if __name__ == "__main__":
+    ur = Application.pull(UR_TO_REJECT)
+    ur.set_application_status(APPLICATION_STATUS_REJECTED)
+    ur.save()
```
richard-jones, line 9: We should make sure to run through all the other checks that happen during application rejection in case they need to be done too.  See https://github.com/DOAJ/doaj/blob/develop/portality/bll/services/application.py#L18

I'd consider at least the following:

* we should add a note to say what was done and why
* check the state of the current_journal value, and set it to related_journal if needed
* check the related journal and see if current_application also needs to be moved to related_application (we can probably just do this manually on the record, rather than adding the code)

richard-jones (changes requested):
At least add a note on the reason for rejection, and then check if there are any other rejection-related changes that need to be made

portality/scripts/reject_update_requst.py
amdomanska: Sorry, I assigned you too soon. Done.

Steven-Eardley (approved):

PR #2152, opened 2022-09-13
pass remote_addr through to plausible for custom events

richard-jones:
# Send user's IP to Plausible

See https://github.com/DOAJ/doajPM/issues/3369

This is a fix for the issue **IF** nginx is providing the user's IP to the application.  ProxyFix (which is already enabled on the app) should set the `request.remote_addr` correctly so we just pass this on in the X-Forwarded-For header.

## Categorisation

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [x] affects the public site

## Basic PR Checklist

- [x] **N/A** FeatureMap annotations have been added
- [x] **N/A**Unit tests have been added/modified
- [x] **N/A**Functional tests have been added/modified
- [x] **N/A** ES queries are wrapped in a Query object rather than inlined in the code
- [x] **N/A** Documentation updates - if needed - have been identified and prepared for inclusion into main documentation (e.g. added and highlighted/commented as appropriate to this PR)
    - [x] Core model documentation: https://docs.google.com/spreadsheets/d/1lun2S9vwGbyfy3WjIjgXBm05D-3wWDZ4bp8xiIYfImM/edit
    - [x] Events and consumers documentation: https://docs.google.com/spreadsheets/d/1oIeG5vg-blm2MZCE-7YhwulUlSz6TOUeY8jAftdP9JE/edit
- [x] There has been a recent merge up from `develop` (or other base branch)

## Testing

I don't know how to test this, unfortunately.  We probably need to:

* Deploy to test server
* Connect to a test Plausible instance (do we have one?)
* See if it works!

## Deployment

### Configuration changes

We need to ensure that the nginx config is set to pass the user's IP to the application in a way that ProxyFix can read.

### Scripts
**N/A**

### Migrations
**N/A**

### Monitoring
**N/A**

### New Infrastructure
**N/A**

### Continuous Integration
**N/A**

### DIFF
--- a/portality/lib/plausible.py
+++ b/portality/lib/plausible.py
@@ -9,6 +9,7 @@ from threading import Thread
 import requests
 
 from portality.core import app
+from flask import request
 
 logger = logging.getLogger(__name__)
 
@@ -40,8 +41,12 @@ def send_event(goal: str, on_completed=None, **props_kwargs):
     if props_kwargs:
         payload['props'] = json.dumps(props_kwargs)
 
+    headers = {}
+    if request:
+        headers = {"X-Forwarded-For": request.remote_addr}  # this works because we have ProxyFix on the app
+
     def _send():
-        resp = requests.post(plausible_api_url, json=payload, )
+        resp = requests.post(plausible_api_url, json=payload, headers=headers)
         if on_completed:
             if resp.status_code >= 300:
                 logger.warning(f'send plausible event api fail. [{resp.status_code}][{resp.text}]')


portality/lib/plausible.py
```diff
@@ -40,8 +41,12 @@ def send_event(goal: str, on_completed=None, **props_kwargs):
     if props_kwargs:
         payload['props'] = json.dumps(props_kwargs)
 
+    headers = {}
+    if request:
+        headers = {"X-Forwarded-For": request.remote_addr}  # this works because we have ProxyFix on the app
```
Steven-Eardley, line 46: Our current ProxyFix setup is `ProxyFix(app.wsgi_app, x_proto=1, x_host=1)`, which corresponds to our nginx directives:

```
proxy_set_header Host $host;
proxy_set_header X-Real-IP $remote_addr;   [adds cloudflare's IP]
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;    [adds Cloudflare's IP]
proxy_set_header X-Forwarded-Proto $scheme;
```

We have to be careful here to make these custom events equivalent to the frontend - will `request.remote_addr` be the real IP if we change the nginx request to include `$http_cf_connecting_ip` - particularly if we're appending rather than replacing as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax

So what I'm asking here is:
* does ProxyFix / flask make any assumptions to calculate request.remote_addr that will differ from the main Plausible client?
* do we need to do anyting with X-Real-IP in addition to forwarded_for
* can cloudflare announce its existence to complate the chain in x-forwarded_for or do we need to do that ourselves in nginx?

richard-jones: I don't know the answers to any of that, I think we'll basically need to test it for real and see what happens.  Perhaps we could set up a mock Plausible endpoint for testing, and log what gets sent.

From the documentation, I got the impression that that the point of ProxyFix was to set `request.remote_addr` to the correct value whatever the situation, although presumably as you say the configs need to match.

Steven-Eardley (approved):
I believe this is safe to merge; we're currently going to be getting the cloudflare IP through to Plausible in the custom events, which is the same as the js one does at the moment.

nginx changes to get the real connecting IP is future work.

PR #2162, opened 2022-10-12
Feature/3215 set up a monitoring task  3

philipkcl:
# Set up a monitoring task on the Background jobs query endpoint

See # https://github.com/DOAJ/doajPM/issues/3215

implement for https://github.com/DOAJ/doajPM/issues/3215#issuecomment-1258624415

## Categorisation

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [ ] affects the public site
- [x] affects the monitoring

## Basic PR Checklist

- [x] FeatureMap annotations have been added
- [x] Unit tests have been added/modified
- [x] **N/A** Functional tests have been added/modified
- [x] ES queries are wrapped in a Query object rather than inlined in the code
- [x] **N/A - no release sheet required for this one** Release sheet has been created, and completed as far as is possible https://docs.google.com/spreadsheets/d/1Bqx23J1MwXzjrmAygbqlU3YHxN1Wf7zkkRv14eTVLZQ/edit
- [x] **N/A** Documentation updates - if needed - have been identified and prepared for inclusion into main documentation (e.g. added and highlighted/commented as appropriate to this PR)
    - [x] Core model documentation: https://docs.google.com/spreadsheets/d/1lun2S9vwGbyfy3WjIjgXBm05D-3wWDZ4bp8xiIYfImM/edit
    - [x] Events and consumers documentation: https://docs.google.com/spreadsheets/d/1oIeG5vg-blm2MZCE-7YhwulUlSz6TOUeY8jAftdP9JE/edit

## Testing

No functional testing required.

Before deployment we should confirm that the status endpoint can be suitably read and acted upon by Uptime Robot

## Deployment

### Configuration changes
new environment variable 
```
BG_MONITOR_LAST_COMPLETED = {
    'main_queue': 7200,
    'long_running': 93600,
}

BG_MONITOR_ERRORS_CONFIG = {
    'journal_csv': {
        'check_sec': 3600,
        'allowed_num_err': 0,
    }
}

BG_MONITOR_QUEUED_CONFIG = {
    'journal_csv': {
        'total': 2,
        'oldest': 1200,
    }
}
```

### Scripts

N/A

### Migrations

N/A

### Monitoring

Uptime Robot will need to monitor the new status information and alert accordingly.

### New Infrastructure

N/A

### Continuous Integration

N/A


portality/view/status.py
```diff
@@ -1,10 +1,14 @@
+import json
+import math
+import os
+import time
+
+import requests
 from flask import Blueprint, make_response, url_for
+
 from portality import util
+from portality.bll import background_task_status
```
richard-jones, line 10: We have a pattern for accessing the BLL, we should access it like

```
from portality.bll import DOAJ
background_service = DOAJ.backgroundService()
```

portality/bll/background_task_status.py
```diff
@@ -0,0 +1,138 @@
+"""
```
richard-jones, line 1: This file should be in `portality.bll.services`

richard-jones: Also, if it's going in as a core service, it needs to follow the same structure, so to be an object with methods.

richard-jones (changes requested):
A couple of review comments from an incomplete review

portality/settings.py
```diff
@@ -1259,3 +1259,27 @@
...
+        'allowed_num_err': 0,
+    }
+}
+
+BG_MONITOR_QUEUED_CONFIG = {
+    'journal_csv': {
+        'total': 2,
+        'oldest': 1200,
+    }
+}
```
richard-jones, line 1284: We need to add configuration for all the other background tasks here too

portality/view/status.py
philipkcl: for more information, 
* any other benefit to create services factory, except avoid Circular import problem? 
* is it because `pre-loading` service is  too slow, if yes, how slow could it be?
* if I should add `background_task_status` to bll.doaj.py, could I avoid to create a class as below ?
```
@classmethod
def backgroundTaskStatus(cls):
    from portality.bll.services import background_task_status
    return background_task_status
```

richard-jones: Avoiding circular imports and convenience for mocking during testing are the primary motivations.  It's not the only pattern we could use for that, but for now it's the one that we landed on.  I'm not averse to doing it differently, but if we change it I want it changed for everything so that we have a consistent pattern.

philipkcl:
@richard-jones updated for feedback

PR #2166, opened 2022-10-31
python and shell scripts for dependencies analysis

amdomanska:
See: https://github.com/DOAJ/doajPM/issues/3325

This PR contains the scripts for dependencies analysis. To run make sure the dependencies.sh is executable and run the dependencies.py script. The script result is saved in dependencies.json and dependencies.csv in the root of the project.

The csv  result of the script run on fresh dev environment is attached.

[dependencies.csv](https://github.com/DOAJ/doaj/files/9901911/dependencies.csv)

### DIFF
--- a/dependencies.py
+++ b/dependencies.py
@@ -23,5 +23,5 @@ if __name__ == "__main__":
 
     all = df_dep.merge(right=df_upgrade, how="left", left_on="package.package_name", right_on="name")
     all_short = all.drop(columns=["name","version", "latest_filetype", "editable_project_location"])
-    all_short = all_short[["package.key","package.package_name","package.installed_version","latest_version","dependencies"]]
+    all_short = all_short[["package.key","package.package_name","package.installed_version","package.required_version","latest_version","dependencies"]]
     all_short.to_csv("dependencies_analysis/dependencies.csv", index=False)
--- a/dependencies.py
+++ b/dependencies.py
@@ -1,0 +1,20 @@
+import subprocess
+import pandas as pd
+import json
+
+if __name__ == "__main__":
+    rc = subprocess.call("./dependencies.sh")
+
+    with open("dependencies.json", encoding='utf-8') as file:
+        lst = json.load(file)
+
+    out = []
+    for d in lst:
+        out.append({})
+        for k, v in d["package"].items():
+            out[-1][f"package.{k}"] = v
+        out[-1]["dependencies"] = "; ".join(v["key"] for v in d["dependencies"])
+
+    df = pd.DataFrame(out)
+    print(df)
+    df.to_csv("dependencies.csv", index=False)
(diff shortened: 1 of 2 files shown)


amdomanska:
Added the check of the latest package version. Updated example of the output file.
[dependencies.csv](https://github.com/DOAJ/doaj/files/9920328/dependencies.csv)

dependencies.py
```diff
@@ -0,0 +1,27 @@
...
+        upg = json.load(file)
+
+    out = []
+    for d in lst:
+        out.append({})
+        for k, v in d["package"].items():
+            out[-1][f"package.{k}"] = v
+        out[-1]["dependencies"] = "; ".join(v["key"] for v in d["dependencies"])
+
+    df_dep = pd.DataFrame(out)
```
Steven-Eardley, line 21: pandas is a pretty heavyweight way of converting json to a csv, and I don't really want it in the project deps since it's a massive library.

dependencies.py
```diff
@@ -0,0 +1,27 @@
...
+    df_upgrade = pd.DataFrame(upg)
+
+    all = df_dep.merge(right=df_upgrade, how="left", left_on="package.package_name", right_on="name")
```
Steven-Eardley, line 24: ooh, this is very data science-y, where'd you learn to do this?

dependencies.py
```diff
@@ -0,0 +1,27 @@
+import subprocess
+import pandas as pd
+import json
+
+if __name__ == "__main__":
+    rc = subprocess.call("./dependencies.sh")
```
Steven-Eardley, line 6: bit dangerous to use subprocess with just an arbitrary shell script, all sorts of nasty things could be put in there.

dependencies.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+rm -rf dependencies_analysis
```
Steven-Eardley, line 3: probably better to do file handling in python; this script will just fail if the directory doesn't exist already

Steven-Eardley (changes requested):
We should add `pipdeptree` to `setup.py`, probably as an option in `extras_require` or add some documentation to that effect.

dependencies.py
richard-jones: Yeah, from our conversation earlier today, we should just run dependencies.sh separately, and then feed the json outputs into the analysis script

dependencies.py
richard-jones: Resolved by moving this out of the standard DOAJ install, as discussed this morning, and just running it as an external script to munge the data from the dependencies.sh

richard-jones:
@amdomanska your updated example of the outputs file is missing the column which tells us which version or version range the package wanted us to install, can we have that back in too.

richard-jones (changes requested):
Aside from Steve's comments, we could do with having the column in the output which tells us which versions were specified by the packages that installed them

richard-jones (approved):
looks like that adds the column that I wanted back in, we should get this run on the various branches of interest

PR #2177, opened 2022-11-24
API failing test case

amdomanska:
Add failing test case in the API

[See #3396](https://github.com/DOAJ/doajPM/issues/3396)

Add the failing test case to avoid KeyError exception

## Categorisation

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [x] affects the public site
- [ ] affects the monitoring

## Basic PR Checklist

- [ ] Unit tests have been added/modified
- [ ] Functional tests have been added/modified
- [ ] No deprecated methods are used
- [ ] No magic strings/numbers - all strings are in `constants` or `messages` files
- [ ] ES queries are wrapped in a Query object rather than inlined in the code
- [ ] Where possible our common library functions have been used (e.g. dates manipulated via `dates`)
- [ ] If needed, migration has been created and tested locally
    - [ ] Events and consumers documentation: https://docs.google.com/spreadsheets/d/1oIeG5vg-blm2MZCE-7YhwulUlSz6TOUeY8jAftdP9JE/edit

## Testing

1. ...
2. ...

## Deployment

### Configuration changes

### Scripts

### Migrations

### Monitoring

### New Infrastructure

### Continuous Integration

### DIFF
--- a/portality/api/current/data_objects/article.py
+++ b/portality/api/current/data_objects/article.py
@@ -173,7 +173,7 @@ class IncomingArticleDO(dataobj.DataObj, swagger.SwaggerSupport):
             if bibjson_element in bibjson:
                 for i in range(len(bibjson[bibjson_element]) - 1, -1, -1):
                     ide = bibjson[bibjson_element][i]
-                    if field is not None:
+                    if field is not None and field in ide:
                         if ide[field] == "":
                             bibjson[bibjson_element].remove(ide)
                     else:
--- a/portality/api/current/data_objects/article.py
+++ b/portality/api/current/data_objects/article.py
@@ -173,8 +173,8 @@ class IncomingArticleDO(dataobj.DataObj, swagger.SwaggerSupport):
             if bibjson_element in bibjson:
                 for i in range(len(bibjson[bibjson_element]) - 1, -1, -1):
                     ide = bibjson[bibjson_element][i]
-                    if field is not None and field in ide:
-                        if ide[field] == "":
+                    if field is not None:
+                        if ide.get(field,"") == "":
                             bibjson[bibjson_element].remove(ide)
                     else:
                         if ide == "":


portality/api/current/data_objects/article.py
```diff
@@ -173,7 +173,7 @@ def _remove_from_the_list_if_empty_data(bibjson_element, field=None):
             if bibjson_element in bibjson:
                 for i in range(len(bibjson[bibjson_element]) - 1, -1, -1):
                     ide = bibjson[bibjson_element][i]
-                    if field is not None:
+                    if field is not None and field in ide:
```
richard-jones, line 176: This gets rid of the exception, but does it have the correct result?

It looks like the objective of the function is to remove an object OR string (represented by `ide`) from the list `bibjson_element`

The old code got the list entry `ide` and then if a field was passed to the function went on to remove that that entry from the list if the value of the field inside that entry is empty.

Now it also only tries to remove that entry if the field is also present in the list element.

But isn't the empty string as the value, and "not present", effectively equivalent, given the function is called `_remove_from_the_list_if_empty_data`?

This update would mean that a list entry would NOT be removed from the list if the key wasn't present, but the object still was.

So, this would get cleaned up

```
[
    { "author": {"name": ""}}
]
```

But this would not

```
[
    {"author": {}}
]
```

I think a better way to do this would be to continue to use the if/else as before and instead:

```
if ide.get(field, "") == ""
```

?

richard-jones (changes requested):
I think this changes the logic in a slightly unintended way

portality/api/current/data_objects/article.py
amdomanska: Yep, you're right. It could also be changed to:
```
if field is not None:
     if field not in ide or ide[field] == "":
```

richard-jones (approved):

PR #2187, opened 2023-01-04
Add new facet to public search: author retains copyright

amdomanska:
Add new facet to public search: author retains copyright

[See #2546](https://github.com/DOAJ/doajPM/issues/2546)

## Categorisation

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [x] affects the public site
- [ ] affects the monitoring

## Basic PR Checklist

- [x] **N/A** FeatureMap annotations have been added
- [x] **N/A** Unit tests have been added/modified
- [x] Functional tests have been added/modified
- [x] Code has been run manually in development, and functional tests followed locally
- [x] **N/A** No deprecated methods are used
- [x] **N/A** No magic strings/numbers - all strings are in `constants` or `messages` files
- [x] **N/A** ES queries are wrapped in a Query object rather than inlined in the code
- [x] Have you done a recent merge up from `develop`
- [x] **N/A** Cleaned up commented out code, etc
- [x] Release sheet has been created, and completed as far as is possible https://docs.google.com/spreadsheets/d/1Bqx23J1MwXzjrmAygbqlU3YHxN1Wf7zkkRv14eTVLZQ/edit#gid=580842123&range=A2
- [x] Documentation updates - if needed - have been identified and prepared for inclusion into main documentation (e.g. added and highlighted/commented as appropriate to this PR)
    - [x] **N/A** Core model documentation: https://docs.google.com/spreadsheets/d/1lun2S9vwGbyfy3WjIjgXBm05D-3wWDZ4bp8xiIYfImM/edit
    - [x] **N/A** Events and consumers documentation: https://docs.google.com/spreadsheets/d/1oIeG5vg-blm2MZCE-7YhwulUlSz6TOUeY8jAftdP9JE/edit
- [x] **N/A** There has been a recent merge up from `develop` (or other base branch)
- [x] The docs for this branch have been generated and pushed to the doc site (see docs/README.md for details)

## Testing

1. public site/public search

### DIFF
--- a/portality/static/js/doaj.fieldrender.edges.js
+++ b/portality/static/js/doaj.fieldrender.edges.js
@@ -2615,6 +2615,15 @@ $.extend(true, doaj, {
                 }
                 apcs += '</li>';
 
+                var rights = "";
+                if (resultobj.bibjson.copyright) {
+                    var copyright_url = resultobj.bibjson.copyright.url;
+                    rights += '<a href="' + copyright_url + '" target="_blank" rel="noopener">';
+                    rights += resultobj.bibjson.copyright.author_retains ? 'Author <strong> retains </strong> all rights' : 'Author <strong> doesn\'t retain </strong> all rights';
+                    rights += '</a>';
+                }
+
+
                 var licenses = "";
                 if (resultobj.bibjson.license && resultobj.bibjson.license.length > 0) {
                     var terms_url = resultobj.bibjson.ref.license_terms;
@@ -2716,6 +2725,9 @@ $.extend(true, doaj, {
                           ' + externalLink + '\
                         <li>\
                             ' + apcs + '\
+                          </li>\
+                          <li>\
+                            ' + rights + '\
                           </li>\
                           <li>\
                             ' + licenses + '\


portality/static/js/edges/public.journal.edge.js
```diff
@@ -203,6 +213,23 @@ $.extend(true, doaj, {
...
+                //     lifecycle: "update",
+                //     updateType: "fresh",
+                //     renderer : doaj.renderers.newORTermSelectorRenderer({
+                //         showCount: true,
+                //         hideEmpty: false,
+                //         open: false,
+                //         togglable: true
+                //     })
+                // }),
+
```
richard-jones, line 232: This can just be removed now

richard-jones (changes requested):
Just last bit of code cleanup before merge, otherwise good

portality/static/js/edges/public.journal.edge.js
amdomanska: done.

richard-jones (changes requested):
I've just run this up locally, and I am unable to remove the new filter from the "X" on the filter lozenge.

Clear All works and removing the filter via the checkbox in the facet works.

amdomanska:
> I've just run this up locally, and I am unable to remove the new filter from the "X" on the filter lozenge.
> 
Fixed! A small change in the doaj.fieldrenderer.edges is needed to interpret the true/false boolean values correctly.

portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -2320,7 +2320,8 @@ $.extend(true, doaj, {
 
                 var value = false;
                 if (ft === "terms" || ft === "term") {
-                    value = el.attr("data-value");
+                    val = el.attr("data-value");
+                    value = val == "true" ? true : (val == "false" ? false : val);
```
richard-jones, line 2325: These equivalences need to be `===`, but also I want to ponder on the type casting nature of this.  I realise this has a specific use for which this line of code might be the solution, but I'd want to consider making the cast from string to actual value something that is controllable.  Like, what if the string "true" is what we actually want for this filter, or what happens with `0` and `1`, etc.

richard-jones: @amdomanska ^ - let me know if you want to talk over it.

amdomanska: @richard-jones It doesn't work for only that particular specific case. This is needed because the "data-value" is a string, so we need a cast from string to bool. If the expected value is "Yes", then (just as for the "has the seal" filter), it will keep the string value and use it. So technically speaking, this is the more controllable version of what was previously implemented - as now the bool value is also possible.

You are right, though, that this doesn't take into consideration the `int` value. I've changed that. I double-checked - we use `int` everywhere where it is possible, so there is no risk of converting the string to int when the string is required.

Now the filter can be used for `string`, `int` and `bool` values (tested locally for all the options).

richard-jones: Yes, ok, we can go with that for now.  What I was getting at, though, is that the renderer doesn't really know what data type you want to convert that string to, so maybe what we could do is allow the implementer to pass in a coerce function, and then for this one pass in a function which converts "true" and "false" to their boolean equivalents.  That would give you complete generality rather than having to guess the format and all the possible variations.

portality/static/js/doaj.fieldrender.edges.js
```diff
@@ -2320,7 +2320,20 @@ $.extend(true, doaj, {
 
                 var value = false;
                 if (ft === "terms" || ft === "term") {
-                    value = el.attr("data-value");
+                    val = el.attr("data-value");
```
Steven-Eardley, line 2323: Do we need this because of inconsistencies in our index, e.g. storing `'false'` as a string vs `false` in `bibjson.copyright.author_retains`?

We've previously had a function for Yes / No depending on what's come from the index, that's handled at a different layer to this.

Steven-Eardley:
Wording mismatch on facet and the 'lozenge' 

image

portality/static/js/doaj.fieldrender.edges.js
amdomanska: The "Has Seal" filter works with "Yes" because we store the same information in two places: `admin.seal` with bool (`true/false`) and `index.has_seal` with string (`"Yes"/"No"`).
We only have `bibjson.preservation.has_preservation` field storing bool. We need to compare it to the string from the HTML data-value attr.

To treat the new filter the same as the existing one, we would need to add the field to `index` (?) that translates the bool value from bibjson to "Yes"/"No" and then uses this one. That, of course, requires migration. And later the same treatment for any new filter.

PR #2191, opened 2023-01-16
Feature/3367 missing oaipmh records

amdomanska:
# "Missing OAIPMH records

[See  3367](https://github.com/DOAJ/doajPM/issues/3367)

## Categorisation

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [x] affects the public site (**oaipmh**)
- [ ] affects the monitoring

## Basic PR Checklist

- [x] Unit tests have been added/modified
- [x] **N/A** Functional tests have been added/modified
- [x] Code has been run manually in development, and functional tests followed locally
- [x] ES queries are wrapped in a Query object rather than inlined in the code
- [x] Cleaned up commented out code, etc
    - [x] **N/A** Events and consumers documentation: https://docs.google.com/spreadsheets/d/1oIeG5vg-blm2MZCE-7YhwulUlSz6TOUeY8jAftdP9JE/edit
- [x] There has been a recent merge up from `develop` (or other base branch)

## Testing

Worth re-running the user's original request to check that they get the results they expect

Harvest the Architecture set: https://doaj.org/oai?verb=ListRecords&set=TENDOkFyY2hpdGVjdHVyZQ~~&metadataPrefix=oai_dc

Harvest the Fine Art set: https://doaj.org/oai?verb=ListRecords&set=TENDOkZpbmUgQXJ0cw~~&metadataPrefix=oai_dc

Look for "Urbanism. Arhitectura. Constructii".

### DIFF
--- a/portality/models/oaipmh.py
+++ b/portality/models/oaipmh.py
@@ -1,5 +1,6 @@
 from copy import deepcopy
 from portality.models import Journal, Article
+from portality import constants
 
 class OAIPMHRecord(object):
     earliest = {
@@ -70,7 +71,7 @@ class OAIPMHRecord(object):
         if start_after is not None or from_date is not None or until_date is not None or oai_set is not None:
 
             if oai_set is not None:
-                a = oai_set.replace("LCC:","")
+                a = oai_set.replace(constants.SUBJECTS_SCHEMA,"")
                 q["query"]["bool"]["should"] = {"match":{"index.classification": a}}
 
             if until_date is not None or from_date is not None or start_after is not None:
--- a/portality/models/oaipmh.py
+++ b/portality/models/oaipmh.py
@@ -50,6 +50,7 @@ class OAIPMHRecord(object):
         "size": 25
     }
 
+    set_limit = {"term": {"index.classification.exact": "<set name>"}}
     range_limit = { "range" : { "last_updated" : {"gte" : "<from date>", "lte" : "<until date>"} } }
     created_sort = [{"last_updated" : {"order" : "desc"}}, {"id.exact" : "desc"}]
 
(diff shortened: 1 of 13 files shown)


portality/models/oaipmh.py
```diff
@@ -71,9 +71,8 @@ def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=
         if start_after is not None or from_date is not None or until_date is not None or oai_set is not None:
 
             if oai_set is not None:
-                s = deepcopy(self.set_limit)
-                s["term"]["index.schema_subject.exact"] = oai_set
-                q["query"]["bool"]["must"].append(s)
+                a = oai_set.replace(constants.SUBJECTS_SCHEMA,"")
+                q["query"]["bool"]["should"] = {"match":{"index.classification": a}}
```
richard-jones, line 75: I think this should be on `index.classification.exact` ? Or does changing it from a `term` query to a `match` fix that?

Also, are we sure it should be a `should` query?  Seems like is should be a `must` query like the one it is replacing?

richard-jones: Yes, I'm confident that this should be a `term` query and a `must`.

Can we please lay it out the way way as it was before, so use `self.set_limit` and just modify it to use `index.classification.exact` instead.

richard-jones (changes requested):

portality/models/oaipmh.py
amdomanska: Yes, you're right. The `should` and `match` is the only way to make it work without the `exact` filter. When `exact` is applied, then it is `must` and `term` that works.

portality/models/oaipmh.py
```diff
@@ -71,8 +72,9 @@ def list_records(self, from_date=None, until_date=None, oai_set=None, list_size=
         if start_after is not None or from_date is not None or until_date is not None or oai_set is not None:
 
             if oai_set is not None:
+                a = oai_set.replace(constants.SUBJECTS_SCHEMA,"")
```
philipkcl, line 75: variable `a` never used. I think is forgot to remove this line.

richard-jones: @amdomanska can you confirm if this line should be used somewhere?

richard-jones: This was a bug, that line is required, but the result of it was not used.  I have fixed and pushed.

richard-jones (approved):
@amdomanska there was one bug which I have fixed for the sake of getting this live sooner

PR #2193, opened 2023-01-17
Feature/3406 add outcome status to bkjob

philipkcl:
# Add outcome_status to background jobs

See https://github.com/DOAJ/doajPM/issues/3406

* add field `outcome_status` to `BackgroundJob` 
* search page support `outcome_status`
* update `IngestArticlesBackgroundTask` to support `outcome_status`

## Categorisation

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [ ] affects the public site
- [ ] affects the monitoring

## Basic PR Checklist

- [ ] Unit tests have been added/modified
- [ ] Functional tests have been added/modified
- [ ] No deprecated methods are used
- [ ] No magic strings/numbers - all strings are in `constants` or `messages` files
- [ ] ES queries are wrapped in a Query object rather than inlined in the code
- [ ] Where possible our common library functions have been used (e.g. dates manipulated via `dates`)
- [ ] If needed, migration has been created and tested locally
- [ ] Have you done a recent merge up from `develop`
- [ ] Cleaned up commented out code, etc
    - [ ] Events and consumers documentation: https://docs.google.com/spreadsheets/d/1oIeG5vg-blm2MZCE-7YhwulUlSz6TOUeY8jAftdP9JE/edit

## Testing

1. ...
2. ...

## Deployment

### Configuration changes

### Scripts

### Migrations

### Monitoring

### New Infrastructure

### Continuous Integration

### DIFF
--- a/doajtest/unit/test_tasks_ingestCrossref531Articles.py
+++ b/doajtest/unit/test_tasks_ingestCrossref531Articles.py
@@ -27,9 +27,9 @@ from lxml import etree
 
 from portality.ui.messages import Messages
 
-
 RESOURCES = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "unit", "resources")
 ARTICLES = os.path.join(RESOURCES, "crossref531_article_uploads.xml")
+
 
 class TestIngestArticlesCrossref531XML(DoajTestCase):
 
@@ -328,7 +328,8 @@ class TestIngestArticlesCrossref531XML(DoajTestCase):
 
         previous = []
 
-        job = ingestarticles.IngestArticlesBackgroundTask.prepare("testuser", url=url, schema="crossref531", previous=previous)
+        job = ingestarticles.IngestArticlesBackgroundTask.prepare("testuser", url=url, schema="crossref531",
+                                                                  previous=previous)
 
         assert job is not None
         assert "ingest_articles__file_upload_id" in job.params
@@ -374,7 +375,8 @@ class TestIngestArticlesCrossref531XML(DoajTestCase):
         # upload dir not configured
         del app.config["UPLOAD_DIR"]
         with self.assertRaises(BackgroundException):
-            job = ingestarticles.IngestArticlesBackgroundTask.prepare("testuser", url="http://whatever", schema="doaj", previous=[])
+            job = ingestarticles.IngestArticlesBackgroundTask.prepare("testuser", url="http://whatever", schema="doaj",
+                                                                      previous=[])
 
     def test_13_ftp_upload_success(self):
 
@@ -388,7 +390,7 @@ class TestIngestArticlesCrossref531XML(DoajTestCase):
         path = os.path.join(upload_dir, file_upload.local_filename)
         self.cleanup_paths.append(path)
 
-        url= "ftp://upload"
+        url = "ftp://upload"
         parsed_url = urlparse(url)
 
         job = models.BackgroundJob()
@@ -431,7 +433,7 @@ class TestIngestArticlesCrossref531XML(DoajTestCase):
         requests.head = ResponseMockFactory.head_fail
         requests.get = ResponseMockFactory.crossref531_get_success
 
-        url= "http://upload"
+        url = "http://upload"
 
         file_upload = models.FileUpload()
         file_upload.set_id()
@@ -679,7 +681,7 @@ class TestIngestArticlesCrossref531XML(DoajTestCase):
 
         stream = Crossref531ArticleFixtureFactory.invalid_schema_xml()
         with open(path, "w") as f:
-           f.write(stream.read())
+            f.write(stream.read())
 
         task = ingestarticles.IngestArticlesBackgroundTask(job)
         task._process(file_upload)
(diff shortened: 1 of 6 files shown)


portality/models/background.py
```diff
@@ -25,6 +34,9 @@ def __init__(self, **kwargs):
 
         super(BackgroundJob, self).__init__(raw=kwargs)
 
+        if not self.outcome_status:
+            self.outcome_status = OutcomeStatus.Success
```
richard-jones, line 38: I would leave this unset if it is not provided

richard-jones (changes requested):
I haven't done a full review yet, but in the mean time some initial change requests, mostly structural

portality/models/background.py
```diff
@@ -1,6 +1,15 @@
...
-from portality.lib import dataobj, dates, es_data_mapping
+from portality.lib import dataobj, dates, es_data_mapping, model_utils
 from portality import dao
+from enum import Enum
+
+
+class OutcomeStatus(Enum):
+    Pending = 'pending'
+    Success = 'success'
+    Fail = 'fail'
```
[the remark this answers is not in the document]
philipkcl: let's move to `portality.constants` but keep `OutcomeStatus`. to see any problem when using Enum class

portality/models/background.py
philipkcl: any suggestion for where to mark as Success ?
if not on __init__ ,  `BackgrounApi.execute` could be another choice. better

richard-jones: Ah yes, I see the problem.  Perhaps on `save` if it has not already been set?  I suppose it doesn't make much difference, perhaps easiest to leave it here in `init`

richard-jones (commented):
@philipkcl I've had a bit of a think about setting the outcome_status value initially and then as the job progresses, and suggest the following:

1. On creating a Background Job model object, if there is no outcome_status, set it to "pending"
2. When `job.fail` is called, that should then also call `job.outcome_fail` - since a failure of the job to complete is by definition a failure to get the right outcome
3. When the `run` method in `ingestarticles.py` completes, set the `outcome_status` to "success"

richard-jones (commented):
@philipkcl could we also just add a test to `doajtest.unit.test_models.test_26_background_job` to test the new model methods

philipkcl:
> When the run method in ingestarticles.py completes, set the outcome_status to "success"

I will set the outcome_status to "success", in `BackgroundApi.execute` in order to avoid developer forgot to set success when implement new `BackgroundTask.run`

philipkcl:
updated for feedback

RK206:
@philipkcl minor correction. 'value' attribute missing at multiple places. 
For example     self.outcome_status = BgjobOutcomeStatus.Pending

doajtest/unit/test_tasks_ingestCrossref531Articles.py
```diff
@@ -814,6 +824,7 @@ def test_27_run_exists(self):
         fu = models.FileUpload.pull(id)
         assert fu is not None
         assert fu.status == "processed"
+        assert job.outcome_status == BgjobOutcomeStatus.Pending
```
RK206, line 827: should we use 'value' attribute for Pending?

portality/background.py
```diff
@@ -59,6 +65,8 @@ def execute(self, background_task):
 
         try:
             background_task.run()
+            if job.outcome_status == BgjobOutcomeStatus.Pending.value:
+                job.outcome_status = BgjobOutcomeStatus.Success
```
RK206, line 69: should we use 'value' attribute for Success?

portality/models/background.py
```diff
@@ -25,6 +28,9 @@ def __init__(self, **kwargs):
 
         super(BackgroundJob, self).__init__(raw=kwargs)
 
+        if not self.outcome_status:
+            self.outcome_status = BgjobOutcomeStatus.Pending
```
RK206, line 32: should we use 'value' attribute for Pending?

RK206 (changes requested):

doajtest/unit/test_tasks_ingestCrossref531Articles.py
philipkcl: yes, missing `.value`

portality/background.py
philipkcl: setter of outcome_status will auto call `.value` if input value is enum
```
    @outcome_status.setter
    def outcome_status(self, outcome_status: Union[str, BgjobOutcomeStatus]):
        if isinstance(outcome_status, BgjobOutcomeStatus):
            outcome_status = outcome_status.value
        self._set_with_struct('outcome_status', outcome_status)
```

portality/models/background.py
philipkcl: reason same as above

philipkcl:
updated for feedback

richard-jones (changes requested):
Code wise this looks good, but I'd like us to update the functional tests to record what we expect the outcomes should be.  I propose that in the DOAJ XML upload tests we add the expected outcome status as the final check in each test.

So in every relevant test in here https://github.com/DOAJ/doaj/blob/feature/3406_add_outcome_status_to_bkjob/doajtest/testbook/article_xml_upload/article_doaj_xml_upload.yml

Add a final `step` with the `results` of what the background job outcome status should be (you can do this with some rough text, and I'll tidy it up on final review)

philipkcl:
@richard-jones , updated testbook for outcome status, please review and let me know if any problem

richard-jones (approved):

Steven-Eardley (approved):

PR #2195, opened 2023-01-18
Remove an unnecessary test expression

openrefactory:
In file: article_form.py, method: form2obj, a logical expression uses the identity operator. A new object is created inside the identity check operation and then used for matching identity. Since this is a distinct, new object, it will not match with anything else. As a result, the identity check will have a logical short circuit. The first part of the test expression already tests for it. The second part is unnecessary.

I suggested that the logical operation should be reviewed.

### DIFF
--- a/portality/crosswalks/article_form.py
+++ b/portality/crosswalks/article_form.py
@@ -43,7 +43,7 @@ class ArticleFormXWalk(object):
 
         # keywords
         keywords = form.keywords.data
-        if keywords is not None:
+        if keywords is not None and len(keywords) > 0:
             if isinstance(keywords, str):
                 keywords = keywords.split(",")
             ks = [k.strip() for k in keywords]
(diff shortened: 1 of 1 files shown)


portality/crosswalks/article_form.py
```diff
@@ -43,7 +43,7 @@ def form2obj(cls, form, add_journal_info=True):
 
         # keywords
         keywords = form.keywords.data
-        if keywords is not None and keywords is not []:
+        if keywords is not None:
```
Steven-Eardley, line 46:
```suggestion
        if keywords is not None and len(keywords) > 0:
```

It's true that the expression didn't have the required effect, but it's better to replace with a working check than remove it entirely.

Steven-Eardley (changes requested):
This is an extremely minor thing, but it's correct that there's no effect:

```pycon
>>> k = []
>>> k is []
False
>>> k is k
True
>>> [] is []
False
```

However, I suggest we change it to the intended check rather than removing it altogether.

openrefactory:
Modified based on suggestion.

Steven-Eardley (approved):

PR #2213, opened 2023-02-28
setup for selenium

philipkcl:
# Setup for selenium test cases

See
* https://github.com/DOAJ/doajPM/issues/3508
* https://github.com/DOAJ/doajPM/issues/3507

Describe the scope/purpose of the PR here in as much detail as you like:
The basic structure for writing Selenium test cases.
some selenium test added in PR as a example.
about how to run the selenium test cases please refer README.md

## Categorisation

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [ ] affects the public site
- [ ] affects the monitoring

## Basic PR Checklist

Instructions for developers:

Instructions for reviewers:

### Code Style

  - [ ] N/A
  - [x] Developer
  - [x] Reviewer

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [x] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [x] Reviewer

### Testing

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [x] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

### Documentation

  - [ ] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [x] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [x] Reviewer

  - [x] N/A
  - [ ] Developer
  - [x] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [x] Reviewer

### Release Readiness

  - [x] N/A
  - [ ] Developer
  - [x] Reviewer

  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

  - 2023-03-28

## Testing

1. ...
2. ...

## Deployment

### Configuration changes

### Scripts

### Migrations

### Monitoring

### New Infrastructure

### Continuous Integration


amdomanska:
Looks good to me.

doajtest/fixtures/article_doajxml.py
```diff
@@ -1,11 +1,14 @@
...
+from pathlib import Path
+
 from lxml import etree
-from io import BytesIO,StringIO
-from copy import deepcopy
 
 RESOURCES = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "unit", "resources")
 ARTICLES = os.path.join(RESOURCES, "doajxml_article_uploads.xml")
 
+ARTICLE_UPLOAD_SUCCESSFUL = Path(os.path.join(RESOURCES, "article_upload_successful.xml")).resolve().as_posix()
```
richard-jones, line 10: We already have a library for path management `portality.lib.paths`, so I suggest that we do the same here as we have with dates, and ensure all path manipulation is done in there.  We can still import `pathlib` there, but lets keep all that stuff encapsulated

doajtest/helpers.py
```diff
@@ -199,6 +200,16 @@ def _make_and_push_test_context(self, path="/", acc=None):
...
 
+    @staticmethod
+    def fix_es_mapping():
+        """
+        you need to call this method if you get some errors like:
+        ESMappingMissingError - 'reason': 'No mapping found for [field]
+
+        :return:
+        """
+        models.Article(**ArticleFixtureFactory.make_article_source()).save(blocking=True)
```
richard-jones, line 211: If the mappings aren't properly initialised we have a problem elsewhere, I would have thought.  It seems you are only seeing this on Articles, what is the issue that caused you to need to write this bit of code?

doajtest/selenium_helpers.py
```diff
@@ -0,0 +1,102 @@
...
+from doajtest.helpers import DoajTestCase
+from portality import app, models
+
+if TYPE_CHECKING:
+    from selenium.webdriver.remote.webdriver import WebDriver
+
+
+class SeleniumTestCase(DoajTestCase):
+    DOAJ_HOST = '172.17.0.1'
+    DOAJ_PORT = 5014
```
richard-jones, line 17: These probably need to come from configuration

doajtest/selenium_helpers.py
```diff
@@ -0,0 +1,102 @@
...
+        super().tearDown()
+
+        self.doaj_process.terminate()
+        print('doaj process terminated')
+
+        self.selenium.quit()
+
+    @classmethod
+    def get_doaj_url(cls) -> str:
+        return f'http://{cls.DOAJ_HOST}:{cls.DOAJ_PORT}'
```
richard-jones, line 66: Likewise, we may want to run this on a service running over `https`, so I'd move the whole `doaj_url` to configuration.

In fact, we have this in `BASE_URL` in settings.py

doajtest/unit/resources/article_upload_successful.xml
richard-jones, line 1: Do we need another copy of this file?  There is one used for functional testing in `doajtest/xml_upload_test_package/DOAJ/successful.xml` and we'll want to make sure that the human tests and the selenium tests are equivalent, so worth using the same files

portality/settings.py
```diff
@@ -1377,3 +1377,6 @@
 # Pages under maintenance
 
 PRESERVATION_PAGE_UNDER_MAINTENANCE = False
+
+# set driver path to run selenium with local browser
+SELENIUM_CHROME_DRIVER_PATH = ''
```
richard-jones, line 1382: I'm not sure what to put in here, can you include an example?

richard-jones (changes requested):
Some minor changes proposed, and a question about mapping initialisation

doajtest/helpers.py
philipkcl: you can see the exception by adding following test cases in any unit test file 
```
    def test_fail(self):
        # from doajtest.fixtures import ArticleFixtureFactory
        # from portality.models import Article
        # models.Article(**ArticleFixtureFactory.make_article_source()).save(blocking=True)
        q = {'query': {'bool': {'must': [{'terms': {'admin.application_status.exact': ['revisions_required', 'pending', 'in progress', 'completed', 'on hold', 'ready', 'draft']}}], 'filter': [{'term': {'admin.owner.exact': 'publisher'}}, {'term': {'admin.application_type.exact': 'new_application'}}]}}, 
             'size': 50, 'sort': [{'last_updated': {'order': 'desc'}}], '_source': {}, 'track_total_hits': True}
        for a in Article.query(q=q):
            print(a)
```


I think `initialise_index` is not prefect and the reason why no one found out this problem is because no one will drop all index in your production environment. 
every time you drop all index in your local, you will run import data and I think the import data fix the mapping.

doajtest/selenium_helpers.py
philipkcl: is different with `BASE_URL` 

for every test case (`setUp`) in current selenium structure, 

it run standalone doaj web server in background.
```
# run doaj server in a background process
def _run():
    app.run_server(host=self.DOAJ_HOST, port=self.DOAJ_PORT)
```
therefore, `get_doaj_url` must be `f'http://{cls.DOAJ_HOST}:{cls.DOAJ_PORT}'` instead of configurable 

We might need to run `cls.originals = patch_config(app, {` to patch `BASE_URL` if any test case need `BASE_URL`



-------
about `https`, I would suggest return `f'https://{cls.DOAJ_HOST}:{cls.DOAJ_PORT}'`
```
app.run_server(host=self.DOAJ_HOST, port=self.DOAJ_PORT, fake_https=True)
```

doajtest/unit/resources/article_upload_successful.xml
philipkcl: we can use same file for this case.

-------
but I'm not sure we are make `all` selenium tests and human tests are equivalent.
because environment in `testdoaj.cottagelabs.com` and selenium is different. 
In selenium tests, it recreate all index for each test case. but `testdoaj.cottagelabs.com` contain all test data and well setup for background job.

doajtest/fixtures/article_doajxml.py
richard-jones: Yeah, I was just imagining adding a function to `portality.lib.paths` like this

```python
def posix_path(source_file):
    return Path(os.path.join(source_dir)).resolve().as_posix()
```

But you make a good point that we should refactor this across the board.  Could you open a refactoring issue for that, then this is fine to leave as is

doajtest/selenium_helpers.py
richard-jones: ok, all I suggest is that all of that stuff should come out of configuration.  So when you set `DOAJ_HOST` and `DOAJ_PORT` and also whether it uses `http` or `https` needs to come from some settings, so that it can be run on anyone's setup without modifying the source of the tests.

doajtest/helpers.py
richard-jones: Ok, I see the problem.  All of the types should push a mapping, but I guess the error here is that the query is being done on an empty index with none of the queried fields already populated?

I've seen this before, and there isn't a great fix for it, that's why we started to move the objects over to `seamless` which supports the generation of the full mapping, but we didn't do that yet for Articles.

Your fix is the best alternative that I've used, though you may want to push and then delete the article straight after so that it doesn't confuse any down-stream tests.  You could also do this in `initialise_index` so that it's true of all empty systems.

philipkcl: ### `solution A`
```
models.Article(**ArticleFixtureFactory.make_article_source()).save(blocking=True)
```
is temporary solution for testing environment, 
I don't know too many about Elasticsearch,  not sure is that good for `initialise_index` and is it fixing the root cause.

let's study this issues in new branch, if we can't found other better solution than `solution A` will be used.

doajtest/selenium_helpers.py
richard-jones: But we may run these tests on the test system, which does use https, or in CI which may or may not use https, so I'd prefer we don't hard code the settings from the start

doajtest/helpers.py
richard-jones: ok, agreed, we shouldn't do this in `initialise_index` now, and we should actually just move Articles over to use seamless, then this problem will be properly dealt with.

Only thing I'd suggest now then, is for you to do

```python
article = models.Article(**ArticleFixtureFactory.make_article_source())
article.save(blocking=True)
article.delete()
```

Then you have an empty index with the mappings initialised

doajtest/selenium_helpers.py
philipkcl: by current design, It can't point to other server. because:
* it run doaj web server in background
* all index will be **drop** for each test case
* some file could be already exist / not found in remote server or local store .

is it point to other server is requirement?
if yes I need take sometime to modify the structure.

richard-jones: Ah, I see, we can't just run these tests against a running system, they operate like unit tests.

For now we don't need to restructure to do that, we can leave the code as-is.  My main point was purely about having clean code, and hard-coding "http" feels wrong, as we put all other references to urls into configuration.

philipkcl: let's hard-coding 'http' for now unit some testcase need to test with https

richard-jones (changes requested):
In getting this set up locally, I found some dependency conflicts that we need to resolve before we merge this.

Specifically:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
requests 2.22.0 requires urllib3!=1.25.0,!=1.25.1,<1.26,>=1.21.1, but you have urllib3 1.26.15 which is incompatible.
```

To fix this I updated requests from 2.22.0 to 2.28.2.  I've reviewed the requests version history, and this seems like a safe change, but we should make sure all the other unit tests are working before we sign that off.

I have not yet added that change to this PR, but we will need to update the requests version before we can merge this.

richard-jones:
I'm using 3.8 which is the same as production

philipkcl:
* added test cases for https://doaj.github.io/doaj-docs/develop/testbook/index.html#article_xml_upload/doaj_xml
* updated dependency and  `requests~=2.29.0` 
* selenium is difficult to write a stable test case, there are 5% chance throw random exception.

philipkcl:
Please review and comment on following PR:
* https://github.com/DOAJ/doaj/pull/2237

PR #2218, opened 2023-03-10
Prevent duplication of update requests in rapid re-submission

richard-jones:
# Prevent duplication of update requests in rapid re-submission

See https://github.com/DOAJ/doajPM/issues/3550

The aim of this is to prevent update request duplication by recording the creation of an update request in Redis and using that as a blocker when a subsequent update request comes in before the index has had a chance to refresh itself.

## Categorisation

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [ ] affects the public site
- [x] affects the publisher area (the API specifically)
- [ ] affects the monitoring

## Basic PR Checklist

Instructions for developers:

Instructions for reviewers:

### Code Style

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

### Testing

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

### Documentation

  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

### Release Readiness

  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

- Release sheet has been created, and completed as far as is possible https://docs.google.com/spreadsheets/d/1Bqx23J1MwXzjrmAygbqlU3YHxN1Wf7zkkRv14eTVLZQ/edit#gid=981308702
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

  - 2024-01-22

## Testing

To test this in a test environment, take a publisher account with at least one journal that does not have an update request.  

you can get this by running the testdrive `publisher_with_journal` like http://localhost:5004/testdrive/publisher_with_journal.

Note their API Key and the ID of the journal.  Go to this script and provide the parameters

```
doajtest/functional/rapid_ur_resubmission.py -u [base url] -k [api key] -j [journal]
```

where [base url] is the domain and port, such as `http://localhost:5004`

This will submit two URs for the same journal in rapid succession, and should give you a successful response output, followed by a failed response output.

### Configuration changes

The existing variables `HUEY_REDIS_HOST` and `HUEY_REDIS_PORT` have been renamed `REDIS_HOST` and `REDIS_PORT` respectively.

A new config `UR_CONCURRENCY_TIMEOUT` has been added, and defaults to 10 seconds.  Recommend to leave this untouched.

### Scripts

N/A

### Migrations

N/A

### Monitoring

N/A

### New Infrastructure
N/A

### Continuous Integration
N/A


portality/bll/services/concurrency_prevention.py
```diff
@@ -0,0 +1,14 @@
+from portality.core import app
+from portality.bll.exceptions import ConcurrentUpdateRequestException
+from portality.ui.messages import Messages
+import redis
+
+class ConcurrencyPreventionService:
+    def __init__(self):
+        self.rc = redis.Redis(host=app.config.get("HUEY_REDIS_HOST"), port=app.config.get("HUEY_REDIS_PORT"))
+
+    def prevent_concurrency(self, journal, id):
```
richard-jones, line 10: Since we're going to do this for articles, I'd make this method name specific to what kind of concurrency

portality/models/v2/application.py
```diff
@@ -210,11 +206,8 @@ def save(self, sync_owner=True, **kwargs):
             self.set_id(self.makeid())
 
         if self.application_type == constants.APPLICATION_TYPE_UPDATE_REQUEST:
-            rc = redis.Redis(host=app.config.get("HUEY_REDIS_HOST"), port=app.config.get("HUEY_REDIS_PORT"))
-            aid = rc.get(self.current_journal)
-            if aid is not None and aid != self.id:
-                raise ConcurrentUpdateRequestException()
-            rc.set(self.current_journal, self.id, ex=10)
+            cs = ConcurrencyPreventionService()
```
richard-jones, line 209: Service should be loaded via the `DOAJ` entry point, same way as the other services

richard-jones (commented):
This came up in my notifications and I made a couple of comments, hope that's helpful!

portality/bll/services/concurrency_prevention.py
amdomanska: I was thinking about it, but wouldn't that be possible (better) to make this service very general and then specify details on another level?

portality/models/v2/application.py
amdomanska: Why do we do this?

portality/bll/services/concurrency_prevention.py
richard-jones: If that's the case you need to think a bit more about the generality of this function.  It takes a `journal` and an `id` but that won't be true of other concurrency requests.  Similarly they won't all raise `ConcurrentUpdateRequestException`s.  So either the function needs to be more general, or you need custom functions for other use cases.

Because we haven't explored the other use cases yet, I'd fall back to having a more specific function for this particular use case, and then we may be able to refactor later when we know about articles in more detail.

portality/models/v2/application.py
richard-jones: It follows a model from more strongly typed languages where the `DOAJ` entry point is effectively a factory for the services, allowing them to (theoretically) be swapped out for other implementations (e.g. mocked ones for testing).  In practice we haven't actually used it that way, so there's an argument for refactoring.  But: if you want to propose a change to the way we do it, it has to be done everywhere, what I don't want is a hybrid set of patterns for this.

amdomanska:
@richard-jones I made all the changes you requested. I have a hard time figuring out other unit tests that could be useful here.

portality/bll/doaj.py
```diff
@@ -117,3 +117,13 @@ def backgroundTaskStatusService(cls):
         """
         from portality.bll.services import background_task_status
         return background_task_status.BackgroundTaskStatusService()
+
+    @classmethod
+    def UpdateRequestConcurrencyPreventionService(cls):
```
richard-jones, line 122: Can we follow the camelcase naming convention please.

portality/bll/services/concurrency_prevention.py
```diff
@@ -0,0 +1,14 @@
+from portality.core import app
+from portality.bll.exceptions import ConcurrentUpdateRequestException
+from portality.ui.messages import Messages
+import redis
+
+class UpdateRequestConcurrencyPreventionService:
```
richard-jones, line 6: I think the name of this class has changed since my last review, which makes this specific to Update Requests.

I'm in two minds about this:

In one way, this is good because it makes the class specific to its usage, and we can then make others for `ArticleConcurencyProventionService` or whatever.

On the other hand, this entire service really only has one function, and every time we want to do something like this for another type or in another context, we need another BLL service.

So, I would do this instead in one of two ways:

1. Have a generic concurrency service with generic capabilities
2. Add the concurrency prevention capabilities to the appropriate services for the type.  So, preventing concurrency for update would either go on the `ApplicationService` (since URs are a kind of application), OR we start a new `UpdateRequestsService` and put it there.

Then, later, when we do this for articles, we add the `prevent_concurrency` method to the `ArticleService`.

My feeling is that the former approach (General Concurrency Service) is better, because it needs to manage a Redis connection, and it would be best to have that all in one place.

There is a third path, which is to do both options: have a generic concurrency service, that in turn is wrapped by specific services on the the type-specific services.  That gives you both benefits: a concurrency method that can be specific to the needs, and a centrally managed access to Redis.

amdomanska:
@richard-jones back to you

portality/bll/services/concurrency_prevention.py
```diff
@@ -3,7 +3,7 @@
 from portality.ui.messages import Messages
 import redis
 
-class UpdateRequestConcurrencyPreventionService:
+class updateRequestConcurrencyPreventionService:
```
richard-jones: Class and method naming conventions are different:

1. Classes - TitleCase
2. Methods: camelCase

The point being to easily distinguish between class and method names and the things they represent.

Therefore, this class name change isn't correct, it was correct before.  It was the method call to obtain the service which was incorrectly cased.

portality/settings.py
```diff
@@ -20,7 +20,7 @@
 PORT = 5004
 SSL = True
 VALID_ENVIRONMENTS = ['dev', 'test', 'staging', 'production', 'harvester']
-CMS_BUILD_ASSETS_ON_STARTUP = False
+CMS_BUILD_ASSETS_ON_STARTUP = True
```
richard-jones, line 23: What's the reason for changing this, it seems unrelated?

portality/core.py
```diff
@@ -177,7 +177,7 @@ def create_es_connection(app):
     # else:
     #     conn = esprit.raw.Connection(app.config['ELASTIC_SEARCH_HOST'], app.config['ELASTIC_SEARCH_DB'])
 
-    conn = elasticsearch.Elasticsearch(app.config['ELASTICSEARCH_HOSTS'], verify_certs=app.config.get("ELASTIC_SEARCH_VERIFY_CERTS", True))
+    conn = elasticsearch.Elasticsearch(app.config['ELASTICSEARCH_HOSTS'], verify_certs=app.config.get("ELASTIC_SEARCH_VERIFY_CERTS", True), http_auth=('admin', 'admin'))
```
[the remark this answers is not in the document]
amdomanska: This shouldn't be committed, sorry!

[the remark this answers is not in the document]
richard-jones: If you put my configuration line above into your `dev.cfg` you can get rid of this code change altogether, and everything should work fine

amdomanska:
@richard-jones I've resolved all the comments and cleaned up. Could you please have a look?

richard-jones (commented):
I think my main summary comment here is that you need to think a bit more about how the code is layered.  You tend to write tightly coupled code which is difficult to imagine extending without refactor, and which leads to greater complexity in the implementation than is necessary.

Instead, try to think about building the behaviour up layer-by-layer, where the layers at the bottom do the most generic and simple things, and the layers above get more specific, and which only speak directly to the layers below them, and respond to the layers directly above them.

portality/bll/doaj.py
```diff
@@ -117,3 +117,13 @@ def backgroundTaskStatusService(cls):
         """
         from portality.bll.services import background_task_status
         return background_task_status.BackgroundTaskStatusService()
+
+    @classmethod
+    def ConcurrencyPreventionService(cls, **kwargs):
```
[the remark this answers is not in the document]
richard-jones: Also, in python we can also use "_" in method names instead.  I think technically that's more pythonic, but I tend to vary between styles without much consistency myself!

portality/models/v2/application.py
```diff
@@ -199,8 +202,17 @@ def prep(self, is_update=True):
             self.set_last_updated()
 
     def save(self, sync_owner=True, **kwargs):
+        if self.id is None:
+            self.set_id(self.makeid())
+
+        if self.application_type == constants.APPLICATION_TYPE_UPDATE_REQUEST:
+            # ~~-> Concurrency_Prevention:Service ~~
+            cs = DOAJ.ConcurrencyPreventionService(journal=self.current_journal, id=self.id, context="update_request")
```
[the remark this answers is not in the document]
richard-jones: The comments aren't above, they are below, because of the ordering of this thread :)

richard-jones:
@amdomanska I have reviewed the latest version, and this time I just pushed a commit with the changes I would make for you to review:

https://github.com/DOAJ/doaj/pull/2218/commits/2d4ca9ff5269b42226f4c1aef0e85a70b0ad2b62

The summary of these changes is:

1. I have generalised the test so that the wait time for the redis record to expire is pulled from configuration
2. HUEY_HOST and HUEY_PORT changed to REDIS_HOST and REDIS_PORT
3. Changed the method name in the service to camelCase
4. Made the storage of concurrency information optional and a method argument, for wider applicability
5. Moved the UR_CONCURRENCY_TIMEOUT config usage to the correct layer of the code

Steven-Eardley (changes requested):
Just a few mixed up config keys

portality/bll/services/concurrency_prevention.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def checkConcurrency(self, key, id):
+        """
+        Checks whether concurrent request has been submitted
+        Returns true if clash is detected
+        """
+        value = self.rs.get(key)
+        return value is not None and value != id
+
+    def storeConcurrency(self, key, id, timeout=10):
```
Steven-Eardley, line 17: Should be the new config key `UR_CONCURRENCY_TIMEOUT` rather than hardcoded to 10s

Set this variable at instantiation so we can define timeout outside the class (since the timeout is UR-specific)

Steven-Eardley (changes requested):
I've run the script to test this and it's working very well, I just have minor comments on the code and message before release

portality/ui/messages.py
```diff
@@ -18,6 +18,8 @@ class Messages(object):
     ARTICLE_METADATA_UPDATE_CONFLICT = ("""Article could not be updated, as it matches another existing article.
         Please check your metadata, and contact us if you cannot resolve the issue yourself.""", "error")
 
+    CONCURRENT_UPDATE_REQUEST = """Concurrent update request has been submitted, try again in a moment"""
```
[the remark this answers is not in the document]
richard-jones: I think this is the right advice, the issue should resolve itself in about 10 seconds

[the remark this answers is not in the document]
richard-jones: Ah, but I suppose this could be an erroneous re-try; I'll add a bit of extra text

PR #2264, opened 2023-07-04
Feature/3575 make notes searchable

RK206:
# Make notes searchable only for admin

journal and application indexes has to be re-indexed for this deployment

See https://github.com/DOAJ/doajPM/issues/3575

## Categorisation

This PR...
- [ ] has scripts to run
- [x] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [x] affects the public site
- [x] affects the editorial area
- [x] affects the publisher area
- [ ] affects the monitoring

## Basic PR Checklist

Instructions for developers:

Instructions for reviewers:

### Code Style

  - [ ] N/A
  - [x] Developer
  - [x] Reviewer

  - [ ] N/A
  - [x] Developer
  - [x] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [x] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [x] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

### Testing

  - [ ] N/A
  - [ ] Developer
  - [x] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [x] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

### Documentation

  - [ ] N/A
  - [x] Developer
  - [x] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [x] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [x] Reviewer

  - [x] N/A
  - [ ] Developer
  - [x] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [x] Reviewer

### Release Readiness

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

- Release sheet has been created, and completed as far as is possible https://docs.google.com/spreadsheets/d/1Bqx23J1MwXzjrmAygbqlU3YHxN1Wf7zkkRv14eTVLZQ/edit#gid=2082243850
  - [ ] N/A
  - [x] Developer
  - [x] Reviewer

  - 2023-10-16
  - 2023-09-01
  - 2023-08-13

## Testing

(https://doaj.github.io/doaj-docs/feature/3575_make_notes_searchable/testbook/index.html#administrative_search/journals/test_notes_searchable_for_admin_and_not_for_public)

## Deployment

journal and application indexes has to be re-indexed for this deployment

### Configuration changes

N/A

### Scripts

N/A

### Migrations

Journal and application indexes has to be reindexed

### Monitoring

N/A

### New Infrastructure
N/A

### Continuous Integration

N/A


doajtest/testbook/administrative_search/journals.yml
```diff
@@ -134,3 +134,22 @@ tests:
...
+      - The note is added to the journal record
+  - step: Go to the admin journal search again and search for the journal you added the note with the words given in the note
+    results:
+      - The journal record is displayed in the search results
+  - step: Search for the journal from public search page with same words given in the note
+    results:
+      - The journal record is not displayed in the search results
+  - step: Search for the journal from api search with same words given in the note
+    results:
+      - The journal record is not displayed in the search results
```
richard-jones, line 155: Can we add here, or in the other functional tests, all the other cases that we want to make sure work:

1. That you *cannot* find the record via a keyword from the note in the public search
2. That you *can* find the record via a search in the admin searches (that's all contexts, admin, editor and associate editor)
3. That searching specifically on the "note" field will work in all those admin searches
4. That searching by a specific field other than "note" will not yield a result in all those admin searches

portality/settings.py
```diff
@@ -484,7 +484,8 @@
 #                "index": False,
                 "store": True
             }
-        }
+        },
+        "copy_to": ["all_meta"]
```
richard-jones, line 488: Doing this on the type means that *all* objects which have unicode types now have a `copy_to` for `all_meta`, and I don't think we want that.  For example, this rule now appears in BackgroundJob records too.

We should look for an approach that adds this only when necessary, probably a type-specific override to these defaults for Journals and Applications

portality/settings.py
```diff
@@ -681,7 +687,7 @@
             "auth" : False,
             "role" : None,
             "query_validator" : "public_query_validator",
-            "query_filters" : ["only_in_doaj", "last_update_fallback"],
+            "query_filters" : ["only_in_doaj", "last_update_fallback", "search_all_meta"],
```
richard-jones, line 690: We also need to apply this constraint to the `journal,article` query route, otherwise that will allow you to search through notes via that endpoint (which is public).  Not sure on the implications of that, as articles do not have an `all_meta` field

portality/settings.py
```diff
@@ -713,15 +719,15 @@
...
+            "query_filters" : ["owner", "only_in_doaj", "search_all_meta"],
             "result_filters" : ["publisher_result_filter"],
             "dao" : "portality.models.Journal"  # ~~->Journal:Model~~
         },
         # ~~->PublisherApplicationQuery:Endpoint~~
         "applications" : {
             "auth" : True,
             "role" : "publisher",
-            "query_filters" : ["owner", "not_update_request"],
+            "query_filters" : ["owner", "not_update_request", "search_all_meta"],
```
richard-jones, line 730: Publisher `update_request` search also requires this constraint

portality/static/js/edges/admin.applications.edge.js
```diff
@@ -61,7 +61,8 @@ $.extend(true, doaj, {
                         {'display':'Country of publisher','field':'index.country'},
                         {'display':'Journal language','field':'index.language'},
                         {'display':'Publisher','field':'bibjson.publisher.name'},
-                        {'display':'Journal: alternative title','field':'bibjson.alternative_title'}
+                        {'display':'Journal: alternative title','field':'bibjson.alternative_title'},
+                        {'display':'Notes','field':'admin.notes'},
```
richard-jones, line 65: Does this need to be `admin.notes.note`?

portality/static/js/edges/admin.update_requests.edge.js
```diff
@@ -71,7 +71,8 @@ $.extend(true, doaj, {
                         {'display':'Country of publisher','field':'index.country'},
                         {'display':'Journal language','field':'index.language'},
                         {'display':'Publisher','field':'bibjson.publisher.name'},
-                        {'display':'Journal: Alternative Title','field':'bibjson.alternative_title'}
+                        {'display':'Journal: Alternative Title','field':'bibjson.alternative_title'},
+                        {'display':'Notes','field':'admin.notes'}
```
richard-jones, line 75: Needs to be `admin.notes.note`

RK206:
All the issues have been addressed. Here is the summary.
- Modifying the url to query the notes from public page. This is rare case as the user should know the field name and notes to query. Even though user queries by updating the url, he will not see the notes as notes is not displayed in the public search results. This is fixed by adding a validator. Now it will return error message as the validator will fail the query.
- 'all_meta' has been moved to MAPPING_OPTS of journal and application and removed from shared struct
- Updated add_mapping_extensions method to add 'copy_to' for required mapping only
- Fixed all other issues mentioned

I have tested the feature locally by deleting the indexes and doing an anonymous import of the indexes journal and application.

portality/settings.py
richard-jones: Hm, that could be a problem, because this would allow the notes data to "leak" through the fixed-query-widget's endpoint, so we need to think of a solution.  I don't really want to add an `all_meta` to articles unnecessarily, but it might be the only way.  Thoughts?

doajtest/unit/test_query.py
```diff
@@ -30,7 +31,12 @@
             "auth" : True,
             "role" : "admin",
             "dao" : "portality.models.Journal"
-        }
+        },
+        "suggestion" : {
```
richard-jones, line 35: Can we call this `application` instead of `suggestion` - we're trying to slowly remove that terminology from the codebase.

portality/settings.py
```diff
@@ -900,6 +920,52 @@
     "private_source": "portality.lib.query_filters.private_source",
     "public_source": "portality.lib.query_filters.public_source",
 }
+# Exclude the fields that doesn't want to be searched by public queries
+# This is part of non_public_fields_validator.
+EXCLUDED_FIELDS = [
```
richard-jones, line 925: Can we give this a more specific name, like `PUBLIC_QUERY_VALIDATOR__EXCLUDED_FIELDS`, just so it's clear what it is out of context

portality/settings.py
```diff
@@ -626,6 +626,19 @@
     }
 }
 
+# Extension Map from dataobj coercion declarations to ES mappings.
+# This is useful when some extensions required for some objects additional to defaults.
+# ~~->DataObj:Library~~
+# ~~->Seamless:Library~~
+DATAOBJ_TO_MAPPING_EXTENSIONS = {
```
richard-jones, line 633: Can we give this a more descriptive name like `DATAOBJ_TO_MAPPING_COPY_TO_EXTENSIONS`

portality/settings.py
```diff
@@ -900,6 +920,52 @@
...
 }
+# Exclude the fields that doesn't want to be searched by public queries
+# This is part of non_public_fields_validator.
+EXCLUDED_FIELDS = [
+    "admin.notes.note",
+    "admin.notes.id",
+    "admin.notes.author_id"
+]
+
+ADMIN_NOTES_SEARCH_MAPPING = {
```
richard-jones, line 931: In the `settings.py` file can we group this with the other MAPPINGS keys, a bit above in the file

portality/settings.py
```diff
@@ -706,7 +719,7 @@
         "article" : {
             "auth" : False,
             "role" : None,
-            "query_validator" : "public_query_validator",
+            "query_validators" : ["non_public_fields_validator", "public_query_validator"],
```
richard-jones, line 722: I'm happy to have two separate validators here chained together, OR for us to update the `public_query_validator` so that it can do both jobs.  If we're having 2 separate validators, maybe we should rename the `public_query_validator` to something more descriptive of what it does (which is to check paging parameters).

richard-jones (changes requested):
All looking good. A couple of very minor nitpicks about variable names, nothing of any great significance.  We do need to resolve the question of how we limit access to notes via the `journal,article` public endpoint, as I think at the moment it would be possible to query for notes via the route.  Possibly need to add this to our technical call next week.

portality/settings.py
RK206: `non_public_fields_validator` is used at the places where `public_query_validator` not used. For example for Publisher application query (Line no 755) So I made them separate validators.

portality/settings.py
RK206: As per my understanding, above mappings are general mappings which apply for all. ADMIN_NOTES_SEARCH_MAPPING applies only for application and journal. So I have separated them. Please correct me if wrong.

doajtest/unit/test_query.py
RK206: I think we need to change at all places if we want to change suggestion to application as this is index_type. Let me know if I need to change at all places.

portality/settings.py
RK206: Updated

portality/settings.py
RK206: updated

doajtest/unit/test_query.py
richard-jones: You are quite right, leave as is for now

Steven-Eardley (approved):

richard-jones (approved):
I've now done a deep review of the security implications, and made some adjustments for absolute certainty, including one fix to how the mappings are generated.  I've also updated the functional tests a bit for ease of use.

From my perspective this is good to go

PR #2283, opened 2023-08-15
Updated the code to exclude Update Request applications

RK206:
# Exclude UR applications from the search

See https://github.com/DOAJ/doajPM/issues/3568

## Categorisation

This PR...
- [x] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [ ] affects the public site
- [ ] affects the monitoring

## Basic PR Checklist

Instructions for developers:

Instructions for reviewers:

### Code Style

  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

### Testing

  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

### Documentation

  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

### Release Readiness

  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

  - [ ] N/A
  - [ ] Developer
  - [ ] Reviewer

  - [date of merge up]

## Testing

1. ...
2. ...

## Deployment

### Configuration changes

### Scripts

### Migrations

### Monitoring

### New Infrastructure

### Continuous Integration

### DIFF
--- a/portality/scripts/application_status_report.py
+++ b/portality/scripts/application_status_report.py
@@ -1,9 +1,20 @@
 from portality import models
+from portality.lib import dates
 import csv
+import os
 
 """This script generates a report of the status of applications in the DOAJ. The output is a CSV file with number
     of applications in each status(new, accepted, rejected) for each year."""
 
+# constants
+SUBMITTED = "submitted"
+ACCEPTED = "accepted"
+REJECTED = "rejected"
+STATUS_ACCEPTED = "status:accepted"
+STATUS_REJECTED = "status:rejected"
+
+DATA = dict()
+DEFAULT_COUNTERS = [(SUBMITTED, 0), (ACCEPTED, 0), (REJECTED, 0)]
 
 def date_applied_query(date_year_from, date_year_to,):
     return {
@@ -25,31 +36,24 @@ def date_applied_query(date_year_from, date_year_to,):
                 }
               ]
             }
-          }
+          },
+          "track_total_hits": True
         }
 
 
-def status_query(date_year_from, date_year_to, resource_id, status):
+def status_query(resource_id):
     return {
           "query": {
             "bool": {
               "must": [
                   {
-                      "term": {
-                          "action": "status:" + status
+                      "terms": {
+                          "action": [STATUS_ACCEPTED, STATUS_REJECTED]
                       }
                   },
                   {
                   "term": {
                       "resource_id": resource_id
-                  }
-                },
-                {
-                  "range": {
-                    "created_date": {
-                      "gte": str(date_year_from) + "-01-01",
-                      "lte": str(date_year_to) + "-12-31"
-                    }
                   }
                 }
               ]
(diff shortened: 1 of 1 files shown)


portality/scripts/application_status_report.py
```diff
@@ -59,27 +72,48 @@ def status_query(date_year, status):
 
     parser = argparse.ArgumentParser()
     parser.add_argument("-o", "--out", help="output file path")
-    parser.add_argument("-y", "--year", help="year to filter by")
+    parser.add_argument("-fy", "--from_year", help="from year to filter by")
```
Steven-Eardley, line 75: definitely time to add the `required=true` arg rather than checking manually below

portality/scripts/application_status_report.py
```diff
@@ -5,7 +5,7 @@
     of applications in each status(new, accepted, rejected) for each year."""
 
 
-def date_applied_query(date_year):
+def date_applied_query(date_year_from, date_year_to,):
```
Steven-Eardley, line 8: probably worth adding the `track_total_hits` param in case the count reaches 10,000 (unlikely as that is)

portality/scripts/application_status_report.py
```diff
@@ -29,27 +29,40 @@ def date_applied_query(date_year):
         }
 
 
-def status_query(date_year, status):
+def status_query(date_year_from, date_year_to, resource_id, status):
```
Steven-Eardley, line 32: likewise, `track_total_hits`

portality/scripts/application_status_report.py
```diff
@@ -59,27 +72,48 @@ def status_query(date_year, status):
...
+
+    if not args.out or not args.from_year or not args.to_year:
         parser.print_help()
         exit()
 
     with open(args.out, "w", encoding="utf-8") as f:
         writer = csv.writer(f)
 
-        res = models.Application.query(q=date_applied_query(args.year), size=0)
+        writer.writerow(["Applications data years", args.from_year + "-" + args.to_year])
```
Steven-Eardley, line 95: I think the task is to find the counts per year from 2018 to 2023, so shouldn't we be incrementing through and adding a row for each calendar year for the date range provided?

portality/scripts/application_status_report.py
```diff
@@ -59,27 +72,48 @@ def status_query(date_year, status):
...
 
-        res = models.Provenance.query(q=status_query(args.year, "accepted"), size=0)
-        writer.writerow(["Accepted", res.get("hits", {}).get("total", {}).get("value", 0)])
+        accepted_counter = 0
+        rejected_counter = 0
+
+        # Application id is saved as resource_id in Provenance model
+        # Iterate through all the applications and query Provenance with application id and status to retrieve the
+        # status of the applications
+        for record in models.Application.iterate(q=date_applied_query(args.from_year, args.to_year), page_size=20):
```
Steven-Eardley, line 107: hmm, this might be an inefficient use of the index - we're querying Applications in a date range with size 0 (fine) then we're iterating through all them (a second query), then we're performing 2 Provenance queries per application.

And here we miss an application that was submitted in one year and changed state in the next, which I think might happen frequently. That suggests we should remove the date range from the provenance queries entirely i.e.:

**for application X that applied in $year, was it ultimately accepted or rejected?**

On the other hand, this isn't how the task is worded, so we should keep our current assumptions, i.e:

* **How many applications were submitted (`created_date` in $year)**
* **how many applications were accepted / rejected (`Provenance` records in $year)**

In which case, I think we need to look through the Provenance index for any accepted or rejected, then look up whether that's an Application or not. What you currently have is halfway between the two interpretations:

**count of applications submitted in $year that were also accepted / rejected in $year**

I think there's a case for clarifying which interpretation of the task to use, but for now we should commit to one of them. How about iterating to get all IDs of applications, then for each accepted / rejected `Provenance` record in $year check it's a new application? (ID appears in our list) That reduces our index to 2n scrolls, rather than n^2 queries and a scroll.

Steven-Eardley (changes requested):
Since we're having a second swing at this I'm being a bit more picky I'm afraid

portality/scripts/application_status_report.py
richard-jones: Here's my take on what the overall effect of the query should be:

1. Get each application (not UR) submitted prior to the start of 2019
2. Record the year that the application was submitted (the created_date)
3. Get the rejected or accepted provenance record for the application
4. Record the year and whether it was rejected or accepted

In terms of efficient implementation, I'm not immediately sure what the best is, but your options are:

1. Scroll the applications limited by the query in (1) above, then query the Provenance for that app id and the status accepted or rejected, which should give you only 1 result (1 scroll, n additional queries)
2. Scroll the applications by the query in (1), and record ALL the application ids and created_dates in memory.  Then scroll the provenance records for rejected or accepted applications and pick out the ids and the status from those that are already in memory (2 scrolls)

RK206: @Steven-Eardley 

> hmm, this might be an inefficient use of the index - we're querying Applications in a date range with size 0 (fine) then we're iterating through all them (a second query), then we're performing 2 Provenance queries per application.

Yes, agree I could have thought better implementation

> How many applications were submitted (created_date in $year)
how many applications were accepted / rejected (Provenance records in $year)

Yes, I understood the requirement this way

> In which case, I think we need to look through the Provenance index for any accepted or rejected, then look up whether that's an Application or not. What you currently have is halfway between the two interpretations:

I don't think we can find if the Provenance record is an application or UR. The type field in the record is recorded as 'suggestion' in both the cases. We can find the application record only through the application id from resource id. Incase of UR, resource_id is not application id. It is new unique id for each record incase of UR.

@richard-jones 
To me option (1) is the better option.

> Scroll the applications limited by the query in (1) above, then query the Provenance for that app id and the status accepted or rejected, which should give you only 1 result (1 scroll, n additional queries)

Here there can be a chance of more than one record. For example if the application got rejected and then accepted and vice versa. In this case I think we can pull latest record and get the status

RK206:
Updated the code as per suggestions. Thanks for your suggestions @Steven-Eardley @richard-jones 
Steve, just to let you know, -o argument takes directory path now not the file path.

Steven-Eardley (approved):
This looks fantastic

PR #2297, opened 2023-09-21
Hotfix/3402 facets accessibility rerelease

amdomanska:
Code for [3402](https://github.com/DOAJ/doajPM/issues/3402) rerelease


sssoz (approved):
Looks good — I honestly thought we already had an `.accordion` class before! Thanks for doing that now.

PR #2316, opened 2023-11-06
modify the script to split the results into 2-3 files

amdomanska:
Run with:

`DOAJENV=production python portality/scripts/230609_find_articles_with_invalid_issns.py -o ~/20231106_3570_invalid_issns`


Steven-Eardley (changes requested):
We should still use the `out` arg as a file prefix

Steven-Eardley:
Looks like it's just a dev change you forgot to reverse - your run instructions do expect the arg

PR #2334, opened 2023-12-12
Some ease of use improvements for the anon import script

Steven-Eardley:
QOL improvements to the anon import script (defaults to S3)


portality/scripts/anon_import.py
```diff
@@ -115,6 +120,10 @@ def do_import(config):
     parser = argparse.ArgumentParser()
 
     parser.add_argument("config", help="Config file for import run")
+    parser.add_argument('-l', '--localstore',
+                        help="Use StoreLocal rather than S3 anon data source",
+                        action='store_true',
+                        required=False)
```
richard-jones, line 126: Perhaps it would be a good idea to be explicit about which store to use, that could be very useful.  That means that we could leave our dev machines configured for StoreLocal, and just trigger use of S3 for this script, rather than having to flip the config

Steven-Eardley: that's exactly what this does - it defaults to using S3 as its anon import sourrce (no matter which is in app.cfg) unless `-l` is specified.

I'll change the arg so that you can supply either, with S3 as the default value. It'll behave the same, but with a bit more verbosity.

philipkcl (approved):

portality/scripts/anon_import.py
richard-jones: Ah, sorry, I misunderstood. In any case, this is a good addition

richard-jones (approved):

PR #2336, opened 2024-01-03
show errors in cms data to admins

Steven-Eardley:
# Display data loading errors on static pages server

Not sure if this is the right approach here - but the server does currently go down (looping exception on application start) when the data is currently unparsable, whcih makes debugging the `static_pages` branch more challenging. Alert looks like so (visible only to admins):

Screenshot from 2024-01-03 12-41-45


RK206 (approved):
Looks ok to me

richard-jones:
The actual feature looks fine, but this looks like broken data could leak these pages into the live UI were they to get through, can we have a config flag to turn them on and set that on the static pages server?

Steven-Eardley:
Updated to include config key

PR #2357, opened 2024-02-29
Feature/3739 email noti ur

philipkcl:
# Email notification for submitted UR needs to be added to the notifications system and is missing from yml file #3739

See #
* https://github.com/DOAJ/doajPM/issues/3739

## Categorisation

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [ ] affects the public site
- [ ] affects the monitoring

## Basic PR Checklist

Instructions for developers:

Instructions for reviewers:

### Code Style

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
### Testing

  - [ ] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [ ] N/A
  - [ ] Developer
  - [ ] Reviewer

- Have CSS/style changes been implemented?  If they are of a global scope (e.g. on base HTML elements) have the downstream impacts of the change in other areas of the system been considered?
  - [ ] N/A
  - [ ] Developer
  - [ ] Reviewer

### Documentation

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

  - [ ] N/A
  - [x] Developer
  - [ ] Reviewer
  
  - [x] N/A
  - [ ] Developer
  - [ ] Reviewer

### Release Readiness

  - [ ] N/A
  - [ ] Developer
  - [ ] Reviewer

  - [ ] N/A
  - [ ] Developer
  - [ ] Reviewer

  - 2024-02-28

## Testing

1. ...
2. ...

## Deployment

### Configuration changes

### Scripts

### Migrations

### Monitoring

### New Infrastructure

### Continuous Integration


portality/events/consumer_utils.py
```diff
@@ -0,0 +1,23 @@
...
+from portality.core import app
+
+
+def parse_application(application_raw: dict) -> models.Application:
+    try:
+        return models.Application(**application_raw)
+    except Exception as e:
+        raise exceptions.NoSuchObjectException(
+            "Unable to construct Application from supplied source - data structure validation error, {x}".format(x=e)
+        )
```
richard-jones, line 12: This should be in the `portality.bll.application.ApplicationService` object called something like `construct_application_from_raw` or similar

portality/events/consumer_utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def parse_email_issns(issns: list) -> str:
+    issns = issns or []
+    return ", ".join(issn for issn in issns)
```
richard-jones, line 17: If we really feel this should be a function in its own right, I would place this on the BibJSON object and call it something like "issns_as_text" or something.  That's `portality.models.v2.bibjson.JournalLikeBibJSON`

portality/events/consumer_utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def is_enable_publisher_email() -> bool:
+    # TODO: in the long run this needs to move out to the user's email preferences but for now it
+    # is here to replicate the behaviour in the code it replaces
+    return app.config.get("ENABLE_PUBLISHER_EMAIL", False)
```
richard-jones, line 23: I would place this on the user account model, and call it `is_email_enabled`, and do a check for both publisher role and this configuration option.  That will form a useful basis to build out this function later as the comment suggests.

It could also be a class method and take the account_id if needed, depending on the context

richard-jones: That's in `portality.models.account`

richard-jones (changes requested):
Some architectural changes to get rid of `util` library

portality/events/consumer_utils.py
philipkcl: Not recommend to make this method to public and general.
each components might have different ways to handle exception.

convert to `NoSuchObjectException` with specific message is fine for consumer module if they want to handle all exception with that message. But this convert will loss information of lower level exception and loss flexibility to handle other exceptions in different ways.

This method will stay in `consumer_utils` for now, since no other module would convert exception in this way.

portality/events/consumer_utils.py
philipkcl: we can move it to model layer for now.

But you will found that if some business requirements change, issns in email should show with other special format. Developer still need to create a `parse_email_issns` in `consumer_utils` because that is local logic of consumer module.

portality/events/consumer_utils.py
philipkcl: move to `portality.models.account` for now, but not recommend to put business logic in data layout, because complex data layout can have problems like high dependency, circle import, difficult to be extract and reuse. 

`AccountService` should be a good place for some where now general that `consumer_utils`

portality/events/consumer_utils.py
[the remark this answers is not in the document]
philipkcl: `consumer_utils` could be a bad name. This module is necessary.
Definition of `consumer_utils` is utils that help developer to build consumer, it is also a share library that contain private logic or common tools of consumer.
you will found out there have a lot of logic only for consumer, such as special string format or special exception handling.

Those kind of `utils` is good to keep cohesion in both higher (`consumer`) and lower (`application`) level. Because lower level (`application`) does not need to known higher level (`consumer`) exist or how to use there service

portality/models/account.py
```diff
@@ -246,3 +246,9 @@ def get_name_safe(cls, account_id) -> str:
             if author is not None and author.name:
                 return author.name
         return ''
+
+    @classmethod
+    def is_enable_publisher_email(cls) -> bool:
+        # TODO: in the long run this needs to move out to the user's email preferences but for now it
+        # is here to replicate the behaviour in the code it replaces
+        return app.config.get("ENABLE_PUBLISHER_EMAIL", False)
```
Steven-Eardley, line 254: I'm not sure that moving this config here currently makes sense, even with your note explaining why.

`ENABLE_PUBLISHER_EMAIL` is currently a **global** setting on the application, so it would make more sense to only register the consumer if `ENABLE_PUBLISHER_EMAIL` is True (i.e. move this to the `should_consume` in portality/events/consumers/update_request_publisher_submitted_notify.py)

However, I agree there _should_ be a user-set flag on their account to disable their own email notifications. In which case, it would go here and have a checkbox + model change to support it

portality/models/v2/bibjson.py
```diff
@@ -601,14 +601,17 @@ def waiver_url(self, url):
...
-    def issns(self):
+    def issns(self) -> list:
         issns = []
         if self.pissn:
             issns.append(self.pissn)
         if self.eissn:
             issns.append(self.eissn)
         return issns
 
+    def issns_as_text(self) -> str:
```
Steven-Eardley, line 612: Excellent, I like this too

Steven-Eardley (changes requested):
only the `ENABLE_PUBLISHER_EMAIL` configuration change is a blocker for release

portality/events/consumer_utils.py
Steven-Eardley: I disagree with this approach @richard-jones , see comment https://github.com/DOAJ/doaj/pull/2357/files/8e54c98cb20e82e9befefeba4b2cc96aae20f7eb#diff-da6267f143e247633ae29f8348a2fe67ffef3f56963b1d36bdbb863e5fab914a

portality/models/account.py
philipkcl: since User might have different email preferences in the future 
move function to `Account.is_enable_publisher_email` is **"pre-design"** for that feature.

even `Account.is_enable_publisher_email` using `ENABLE_PUBLISHER_EMAIL` is not make sense for now.
That would be still fine because `Account.is_enable_publisher_email` and `ENABLE_PUBLISHER_EMAIL` are **temporary** until User base email preferences implemented.

 I would suggest move `is_enable_publisher_email` to account_service too when implement that feature to avoid models layer too complex.

portality/events/consumer_utils.py
richard-jones: Ok, I'm happy to leave it for now, but I feel that probably there's a better reorganisation of code that we haven't yet thought of.

For example, the `NoSuchObjectException` is probably wrong in this case, anyway, as that's a BLL level exception meaning that the object isn't in the database.  This is a pre-existing oddity, not added by this PR.  But perhaps a better consumer-level exception would be suitable here instead, OR we add a more general application-level exception which says that an object could not be parsed, and which is more general than the SeamlessException that this wraps.

In any case, a discussion for another time!

richard-jones (approved):
I've approved this PR as is, but there is an open discussion regarding the publisher email enabled flag which we should resolve

portality/models/account.py
Steven-Eardley: I accept it for the timebeing :D

PR #2364, opened 2024-03-27
Make sure the bottom of notes section never is below the viewport edge

amdomanska:
https://github.com/DOAJ/doajPM/issues/3844

This fix makes sure that the bottom of the Notes section never is below the viewport's edge (when scrolled to the top of the form).


Steven-Eardley (approved):
Looks good to me - tested locally.

PR #2365, opened 2024-03-27
Remove unnecessary background

amdomanska:
Issue: https://github.com/DOAJ/doajPM/issues/3848


RK206 (approved):
account page looks better now

Steven-Eardley (commented):
An observation - this does remove the odd salmon background on the account page, but I'm pretty sure the marketing consent box did previously have a background of some sort - maybe grey?

And this does remove the pink on the bulk edit form too, but again those styles have more problems (e.g. the padding that's still present) so it's worth understanding why these styles got conflated and fix properly.

PR #2373, opened 2024-04-08
Allow admins to retrieve apps via API

amdomanska:
https://github.com/DOAJ/doajPM/issues/3864

### DIFF
--- a/portality/api/current/crud/applications.py
+++ b/portality/api/current/crud/applications.py
@@ -197,7 +197,7 @@ class ApplicationsCrudApi(CrudApi):
 
         # is the current account the owner of the application
         # if not we raise a 404 because that id does not exist for that user account.
-        if ap.owner != account.id:
+        if "admin" not in account.role and ap.owner != account.id:
             raise Api404Error()
 
         # if we get to here we're going to give the user back the application
--- a/portality/api/current/crud/applications.py
+++ b/portality/api/current/crud/applications.py
@@ -197,7 +197,7 @@ class ApplicationsCrudApi(CrudApi):
 
         # is the current account the owner of the application
         # if not we raise a 404 because that id does not exist for that user account.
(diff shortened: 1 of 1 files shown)


portality/api/current/crud/applications.py
```diff
@@ -197,7 +197,7 @@ def retrieve(cls, id, account):
 
         # is the current account the owner of the application
         # if not we raise a 404 because that id does not exist for that user account.
-        if ap.owner != account.id:
+        if "admin" not in account.role and ap.owner != account.id:
```
richard-jones, line 200: Use `account.is_super`

richard-jones (changes requested):
Need to do this with the account API rather than a magic string

portality/api/current/crud/applications.py
amdomanska: 👍🏼

portality/api/current/crud/applications.py
```diff
@@ -197,7 +197,7 @@ def retrieve(cls, id, account):
 
         # is the current account the owner of the application
         # if not we raise a 404 because that id does not exist for that user account.
-        if ap.owner != account.id:
+        if account.is_super not in account.role and ap.owner != account.id:
```
richard-jones, line 200: I don't think you tested this before committing it.

is_super is a function which determines if the account is the super user.

```
if not account.is_super and ap.owner != account.id
```

richard-jones (changes requested):
Test before you commit code changes :)

portality/api/current/crud/applications.py
amdomanska: > I don't think you tested this before committing it.

I'm afraid you're right. I wanted to push it quickly but got distracted. I'll fix it asap. I'm sorry for that!

Steven-Eardley (approved):

PR #2397, opened 2024-07-09
Received applications by country 2019-2023 #3918

amdomanska:
See: [#3918](https://github.com/DOAJ/doajPM/issues/3918)

The script returns a CSV file in a 3-column format:

`Country | Year | Count`

ordered by country, and then by year.

`python portality/scripts/3918_received_app_by_country_and_year/script.py -o output.csv`

### DIFF
--- a/portality/scripts/3918_received_app_by_country_and_year/script.py
+++ b/portality/scripts/3918_received_app_by_country_and_year/script.py
@@ -1,0 +1,156 @@
+from portality import models
+from portality.bll import exceptions
+import csv
+from datetime import datetime
+
+QUERY_BY_YEAR = {
+    "track_total_hits": True,
+    "query": {
+        "bool": {
+            "must": [
+                {"match_all": {}}
+            ],
+            "filter": [
+                {
+                    "range": {
+                        "created_date": {
+                            "gte": "2021-01-01T00:00:00Z",
+                            "lte": "2023-12-31T23:59:59Z"
+                        }
+                    }
+                }
+            ]
+        }
+    },
+    "sort": [{"created_date": {"order": "asc"}}],
+    "aggs": {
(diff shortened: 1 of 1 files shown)


Steven-Eardley (changes requested):
This is a good attempt, but I do have a few suggestions of how to improve it for accuracy, efficiency, and ease of use:

* The only difference of output report is the `country` and `created` columns, so I wondered why we have two reports at all, which led me to the following:
    + If it's just the counts of articles per year and country that we're after for reporting purposes, we can make just one query with both aggregations
    + Since the original issue has year span 2019-2023, I've expanded the range in the query
    + I've added the application type to the query so that we exclude URs as per the issue description too
* sorting or filtering by year and country are both easily done by spreadsheet software, so we should just iterate through all results and include those columns in the CSV and leave it at that. Your existing sort by date is a useful default giving us ascending years in the CSV.

So the query I recommend is:
```json
{
    "track_total_hits": true,
    "query": {
        "bool": {
            "must": [
                {"term": {"admin.application_type.exact": "new_application"}}
            ],
            "filter": [
                {
                    "range": {
                        "created_date": {
                            "gte": "2019-01-01T00:00:00Z",
                            "lte": "2023-12-31T23:59:59Z"
                        }
                    }
                }
            ]
        }
    },
    "sort": [{"created_date": {"order": "asc"}}],
    "aggs": {
        "applications_by_year": {
            "date_histogram": {
                "field": "created_date",
                "calendar_interval": "year",
                "format": "yyyy",
                "min_doc_count": 0
            }
        },
        "applications_by_country": {
            "terms": {
                "field": "bibjson.publisher.country.exact",
                "size": 1000
            }
        }
    }
}
```

Steven-Eardley (approved):

PR #2433, opened 2024-11-20
Feature/4016 anon import aliases

philipkcl:
* Issue: 
* https://github.com/DOAJ/doajPM/issues/4016

---

# anon import aliases

*briefly describe the PR here*

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [ ] affects the public site
- [ ] affects the monitoring

## Developer Checklist

* [x] Code meets acceptance criteria from issue
* [x] User Test Scripts (if required) are written and have been run through
* [x] Project's coding standards are met
* [x] There is a recent merge from `develop`

## Reviewer Checklist

## Testing

*List user test scripts that need to be run*

*List any non-unit test scripts that need to be run by reviewers*

## Deployment

### Configuration changes

### Scripts

### Migrations

### Monitoring

### New Infrastructure

### Continuous Integration


portality/dao.py
```diff
@@ -964,6 +966,25 @@ def refresh():
     return ES.indices.refresh()
 
 
+def find_indexes_by_prefix(index_prefix) -> list[str]:
+    data = ES.indices.get(f'{index_prefix}*')
```
Steven-Eardley, line 970: it might be good to enforce the separator here (`if not index_prefix.endswith('-') ...`)  because here is where a missing separator could match a whole lot more indexes than we indend. E.g. if you supply `doaj` as the index prefix you'll delete indexes doajstatic-article etc. That won't happen unless the comment in settings.py to include the separator is ignored.

Perhaps we should just ensure the setting always ends with `-` instead. Either way, not a blocker for release.

portality/scripts/anon_import.py
```diff
@@ -113,10 +164,23 @@ def do_import(config):
...
+    print("\n==Creating Aliases==")
+    for index_detail in index_details.values():
+        for retry in range(5):
+            if not es_connection.indices.exists(index_detail.alias_name):
+                break
+            print(f"Old alias exists, waiting for it to be removed, alias[{index_detail.alias_name}] retry[{retry}]")
+            sleep(5)
+
+        print("Creating alias: {:<30} -> {}".format(index_detail.instance_name, index_detail.alias_name))
+        es_connection.indices.put_alias(index=index_detail.instance_name, name=index_detail.alias_name)
```
Steven-Eardley, line 177: It would be helpful to perform the alias creation earlier in the script - i.e. immediately after initialisation rather than after data import, because the site requires the aliases to function, and it takes a number of minutes to insert all of the data.

If we create aliases immediately after the records, the DOAJ app will run successfully straight away with empty data until the type is reached in import.

Steven-Eardley (changes requested):
This is a great improvement, thanks @philipkcl  - just a couple of changes suggested before merge.

portality/dao.py
philipkcl: Not recommend it.

It will fail the first time import. The first time import meaning when the the index name still calling `doaj-account` instead of `doaj-account-20111212`, which mean it will help to migrate if the environment still using old index name `doaj-account`.

Moreover, prefix search by `index_type` and  the script will show which index will be removed, it should be safe enough.

Steven-Eardley (approved):

PR #2459, opened 2025-03-10
Hotfix/2025 02 24 4067 point in time scroll

richard-jones:
* Issue: https://github.com/DOAJ/doajPM/issues/4067

---

# PIT iteration and anon export

Uses search_after (and optionally PIT searching) to improve full index iteration performance, and introduces some performance enhancements to the anon_export process

This PR...
- [x] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [ ] affects the public site
- [ ] affects the monitoring

## Developer Checklist

* [x] Code meets acceptance criteria from issue
* [x] Unit tests are written and all pass
* [x] User Test Scripts (if required) are written and have been run through
* [x] Project's coding standards are met
* [x] Code documentation and related non-code documentation has all been updated
* [x] Migation has been created and tested
* [x] There is a recent merge from `develop`

## Reviewer Checklist

## Testing

This needs to be deployed to test, and then anon_export run.  First from the command line to confirm behaviour, and then also by schedule using the background jobs.  If both are successful, then the next step is to ensure that the data can be re-imported from the export.  To do that, export to the local machine, and then re-import from the local machine.

To export to the local machine, ensure the following setting:

```
STORE_IMPL = "portality.store.StoreLocal"
```

To import from the local machine run the command with the following arguments

```
python anon_import.py -s local [path to import config]
```

### Scripts

Once this has been deployed, the `anon_export.py` script should be run immediately to bring the anonymous data on S3 up to latest

### New Infrastructure

By default this code DOES NOT require any infrastructure changes.  It uses a `search_after` approach which will work on the current infrastructure, but does not absolutely guarantee the coherence of the output (probably good enough for testing).

This code change also allows for the possibility of using PIT search, which is only available in the `default` distributin of ES 7.10.x, that WILL NOT WORK on OSS 7.10.It also WILL NOT WORK on OpenSearch of any version, though PIT is supported in OS 2.x, it is not known if the ES client library will work with it, as the codebases have diverged.

Therefore, to use the PIT features we will need to upgrade our ES instance.

### DIFF
--- a/portality/dao.py
+++ b/portality/dao.py
@@ -579,6 +579,96 @@ class DomainObject(UserDict, object):
         :param limit: Limit the number of results returned (e.g. to take a slice)
         :param wrap: Whether to return the results in raw json or wrapped as an object
         :param keepalive: scroll timeout
+
+        TODO: this is the old method, we should evaluate and aim to make iterate_pit the default scroll method.
+        """
+        theq = {"query": {"match_all": {}}} if q is None else deepcopy(q)
+        theq["size"] = page_size
+        theq["from"] = 0
+        if "sort" not in theq:
+            # This gives the same performance enhancement as scan, use it by default. This is the order of indexing like sort by ID
+            theq["sort"] = ["_doc"]
+
+        # Initialise the scroll
+        try:
+            res = cls.send_query(theq, scroll=keepalive)
+        except Exception as e:
+            raise ScrollInitialiseException("Unable to initialise scroll - could be your mappings are broken", e)
+
+        # unpack scroll response
+        scroll_id = res.get('_scroll_id')
+        total_results = res.get('hits', {}).get('total', {}).get('value')
+
+        # Supply the first set of results
+        counter = 0
+        for r in cls.handle_es_raw_response(
+                res,
+                wrap=wrap,
+                extra_trace_info=
+                "\nScroll initialised:\n{q}\n"
+                "\n\nPage #{counter} of the ES response with size {page_size}."
+                        .format(q=json.dumps(theq, indent=2), counter=counter, page_size=page_size)):
+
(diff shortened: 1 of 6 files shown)


portality/dao.py
```diff
@@ -685,7 +886,20 @@ def dump(cls, q=None, page_size=1000, limit=None, out=None, out_template=None, o
             out = sys.stdout
 
         count = 0
-        for record in cls.scroll(q, page_size=page_size, limit=limit, wrap=False, keepalive=scroll_keepalive):
+
+        iterator = None
+        if iterate_method == "unstable":
```
Steven-Eardley, line 891: Since we've moved to python 3.10 this is a good place to use a `match` statement rather than iffy code. https://docs.python.org/3/reference/compound_stmts.html#match

portality/dao.py
```diff
@@ -658,17 +661,215 @@ def iterate(cls, q: dict = None, page_size: int = 1000, limit: int = None, wrap:
...
+
+            # get the next page and check that we haven't timed out
+            try:
+                res = cls.send_query(theq)
+                if len(res.get('hits', {}).get('hits', [])) == 0:
+                    break
+                search_after = first_resp.get('hits', {}).get('hits', [])[-1].get('sort', [])
+            except elasticsearch.exceptions.NotFoundError as e:
+                raise ScrollTimeoutException(
+                    "PIT timed out; {status} - {message}".format(status=e.status_code, message=e.info))
```
Steven-Eardley, line 837: Confusing error message, given that this one doesn't use PIT search

Steven-Eardley (approved):
Looks OK. minor comments

Steven-Eardley (approved):

PR #2502, opened 2025-08-13
Add chunking to sitemap generator, manually add new sitemap indexes t…

Steven-Eardley:
…o robots

* Issue: https://github.com/DOAJ/doajPM/issues/4152

---

# Generate 3 sitemap files

This will generate a number of index files according the setting `SITEMAP_INDEX_MAX_ENTRIES` and add them to the cache, rather than a single `sitemap.xml` file as the index. Things to note:

* The existing `sitemap` cache entry remains, and `get_latest_sitemap()` doesn't work correctly because it doesn't know about multiple indexes.
* I've manually added the new sitemap indexes to robots-production.txt.
* The naming scheme differs and might be confusing - where we previously had `sitemap.xml` (the index) and `sitemap0...n.xml` as the individual sitemaps, we now have `sitemap_index_0.xml` and `sitemap0.xml` etc. They're a bit clearer now only if you set `SITEMAP_INDEX_MAX_ENTRIES` low enough for multiple indexes to be created.
* We _could_ properly support the old `sitemap.xml` single route (and correctly cache as latest) if we use an index-of-indexes approach. TODO: check this actually fits the protocol - I've only ever seen it referred to as having multiple indexes.

Also, I had trouble testing this when I was using localstore - something in the sitemap / cache code seems to expect it to be S3 no matter which `STORE_IMPL` is set.

This PR...
- [ ] has scripts to run
- [ ] has migrations to run
- [ ] adds new infrastructure
- [ ] changes the CI pipeline
- [ ] affects the public site
- [ ] affects the monitoring

## Developer Checklist

## Reviewer Checklist

## Testing

* Generate the sitemap, check the background task log makes sense. You must have articles in the index. (dev anon import is fine)
* access via /sitemap.xml to see the index referring to sitemap0.xml, sitemap1.xml etc all in one file
* Set SITEMAP_INDEX_MAX_ENTRIES to 1
* re-run background job and check output
* Access the new sitemap indexes via /sitemap_index_0.xml 

## Deployment

Robots.txt has been updated so Google should try to read the new indexes, and not the old sitemap.

### Configuration changes

### Scripts

### Migrations

### Monitoring

### New Infrastructure

### Continuous Integration

### DIFF
--- a/deploy/robots-production.txt
+++ b/deploy/robots-production.txt
@@ -1,4 +1,6 @@
-Sitemap: https://doaj.org/sitemap.xml
+Sitemap: https://doaj.org/sitemap_index_0.xml
+Sitemap: https://doaj.org/sitemap_index_1.xml
+Sitemap: https://doaj.org/sitemap_index_2.xml
 
 User-agent: *
 Disallow: /search/
--- a/doajtest/mocks/models_Cache.py
+++ b/doajtest/mocks/models_Cache.py
@@ -41,6 +41,23 @@ class InMemoryCache(object):
         return cls.__memory__["sitemap"]
 
     @classmethod
+    def cache_sitemap_indexes(cls, urls):
+        """Cache multiple sitemap index URLs"""
+        for idx, url in enumerate(urls):
+            cls.__memory__[f"sitemap_index_{idx}"] = {
+                "filename": url
+            }
+
+    @classmethod
+    def get_sitemap_index(cls, n):
+        """Get a specific sitemap index URL"""
+        key = f"sitemap_index_{n}"
+        cached = cls.__memory__.get(key)
+        if cached is None:
+            return None
+        return cached.get("filename")
+
+    @classmethod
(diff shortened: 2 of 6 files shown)


richard-jones (changes requested):
I haven't pinpointed the exact line of code where the issue arises, but when I generate a set of sitemap index files, the list of sitemaps in each one is the same.  That is, they all start at `sitemap0.xml` and go to `sitemap89.xml`.

Furthermore, not all of these sitemaps are being stored to the cache.  I generated 315 sitemaps, which resulted in 4 sitemap indexes, and of those only 51 of the expected 319 records were in the cache after.

I think there's some numerical index re-use going on somewhere, which is causing the sitemaps to "collapse" into a smaller apparent number than there actually are.

I would need to investigate further to identify the actual issue.

Sitemaps themselves seem to be generated correctly, although I note (and this was not introduced by this PR) that the filenames are a little odd, of the form `_N_utf8.xml`, missing the `sitemap_doaj_DATE` prefix.  This may be a deliberate choice, as they are collected into a directory of that name.

PR #2591, opened 2026-05-06
quick and dirty status monitoring ui for background jobs

richard-jones:
* Issue: N/A

---

# Quick UI for Background Jobs status monitoring

Provides a webpage which monitors /status and gives easy to see highlights on background jobs health, for manual monitoring.

I have removed all the PR checklist stuff here, as this is a pure ops/dev change which does not impact the running application.


Steven-Eardley (approved):
Looks good. It's a new authenticated route but since it's under /admin it's already handled in the loadbalancer (confirmed).

PR #2616, opened 2026-07-23
Fix leaking/failing unit tests across the suite

Steven-Eardley:
Supersedes #2613, whose scope has broadened from tombstone-only fixes to clearing all remaining CI test-suite leaks.

## Summary

* Don't create a tombstone for the dummy article used to push mappings to a fresh index, so we don't leave more to clean up.
* Various unit test fixes: create journals when a tombstone is expected, clean up articles fully where repeatedly reused in tests, fix a type mismatch in the plausible test, more non-tombstoning fixes to crossref tests.
* Wrap `patch_config`-mutating test bodies in `try/finally` across 8 test files (sitemap, journal CSV, journal CSV prune, public data dump, shorturl, OAI-PMH, query filters, old data cleanup) so a failed assertion mid-test no longer leaves the app config patched for whatever test runs next in the same CI process.

## Testing

* [x] Unit tests are written and all pass


Steven-Eardley:
This also adds a utility script that pulls CI results to help AI-assisted debugging of CI a bit easier. It's a standalone script under `scratchpad/` that requires the `gh` tool to pull the results from CircleCI

doajtest/unit/test_tasks_ingestDOAJarticles.py
```diff
@@ -107,6 +107,25 @@ def tearDown(self):
             if os.path.exists(path):
                 os.remove(path)
 
+        # Most test methods here create articles/journals via shared
```
RK206, line 110: Just a suggestion. If this part of the code reusable at other places some other time, it can be moved to a common function.

RK206 (approved):
Looks good.

When I run all tests on my local, some tests are failing but they passed when tested individually. So that may not be an issue I believe.

doajtest/unit/test_tasks_ingestDOAJarticles.py
Steven-Eardley: I think there already is such a thing but it's used inconsistently across the test suite and wasn't my shortest route to working tests, but you're right that we should take it into account for a future refactor

